PR #6, opened 2015-03-16
MOBILE-823 login: Add site first implementation

dpaloucva:
Added a first implementation of the add site feature

Issue: https://tracker.moodle.org/browse/MOBILE-823


FMCorz:
Noting observations as I look through the patch:

- No need to add 'ctrl' to controller/*.js files
- You surely have written that before but we agreed to use 'BlahController' rather than 'BlahCtrl'
- I think we could use $mmConfig directly rather than using it in a 'resolve', and the $mmConfig.initConfig could be set in .run() of its module.
- Why do we need a specific 'retry' parameter to getUserToken? That method should not build a URL IMO.
- Not too sure what kind of key that is getItem('service'+siteurl);
- Why do we prioritise HTTP over HTTPS in determineService?

I'm conscious that most of these are coming from the original app, I'm just asking.

PR #49, opened 2015-05-21
MOBILE-1001 scss: Re-organised the SCSS files arrangment

FMCorz:
The extra styling of the app should happen in ./scss/app.scss

The core elements available are in ./www/core/scss along with
their default styles.

Any style that is specific should be saved under its component
or addon scss/ directory.


FMCorz:
Patch updated.

PR #346, opened 2016-01-13
MOBILE-1377 ui: Center avatar on side menu and profile screen

barbararamiro:


barbararamiro:
Note that MOBILE-1377 is based on MOBILE-1371

PR #376, opened 2016-02-08
Log message typo.

earcanal:


jleyva:
Thanks for the typo, unfortunately we cannot integrate it, we need a commit with the format:

MOBILE-XYZ component: Fix description

Feel free to create the issue in the tracker and submit again the fix!

Thanks again for spotting it

PR #387, opened 2016-02-21
Update package.json

jaavid:


jleyva:
Thanks for reporting this, I'm closing it because we already fixed it via a tracker issue. 

Thanks again, Juan

PR #410, opened 2016-03-03
Fix login via a browser window with a custom URL scheme

albertgasset:
Users can't to log in when using a custom URL scheme because the URL is checked against the string "moodlemobile://token=", instead of using the setting defined in www/config.json.


jleyva:
Thanks Albert, please can you review my comment in https://tracker.moodle.org/browse/MOBILE-1449?

PR #483, opened 2016-05-18
Mobile 1577

jleyva:


pedroist:
How can we allow attachments instead of preventing submission?

PR #595, opened 2016-08-09
MOBILE-1503 Use a toggle instead of long press to delete messages

msoni11:


jleyva:
Thanks for your contribution, we'll review it once we are back from vacation

msoni11:
Sure. There are still some limitations and style related issues.

Delete icon should come on right side for mine messages and left for their messages. It seems like ionic doesn't support delete button to come on right side.

jleyva:
Closing since this was integrated

PR #642, opened 2016-09-25
Mobile 1713 choice - Improve choice results in the app

HaakonBj:
Swapped the choice results in the app to using chart.js.
Unsure how to deal with the strings that are not being used anymore, some help on this would be appreciated.
Test:
1. Create a choice activity, allow the choices to be published.
2. Log in as students and vote on the choice activit. 
3. Log in as person who can see the results.

### DIFF
--- a/.bowerrc
+++ b/.bowerrc
@@ -1,3 +1,6 @@
 {
-  "directory": "www/lib"
+  "directory": "www/lib",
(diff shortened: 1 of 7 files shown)


HaakonBj:
Found some mistakes, with npm-resolver and users being displayed when choice is not published.

PR #699, opened 2016-10-24
German translation contributions.

lucaboesch:
Juan, unfortunately I did not see where in AMOS we could have translated. Please accept this pull request.
Best regards,
Luca


moodlemobile:
Thanks for your contribution but you can translate the German strings via https://lang.moodle.org/
You have to select the component: local_moodlemobileapp and Moodle version 2.6, you will see there the list of unstranslated strings

lucaboesch:
Oh, fair enough, I will deposit my contributions there!
You may spread this information further (didn't find it anywhere, I searched for it) and please do point out as well that you have to choose 2.6 as Moodle Version in AMOS ([https://lang.moodle.org/]) otherwise no result shows up.

jleyva:
Thanks for the suggestion, I created https://docs.moodle.org/dev/Translating_Moodle_Mobile

PR #713, opened 2016-11-06
Mobile 1636 courses, category: Browse courses by course category. 

brikendr:


moodlemobile:
Hi,

thanks for working on this but, unfortunately,  we were thinking in a different approach that would be validated for our UX team.

We were thinking in add a tabs navigation in the My courses page so the user could switch between listing and category browsing.

Category browsing will be a tree-based navigation, user will be able to expand and browse by categories, this will require Moodle 3.2 (some changes were made in Web Services to allow this)

Gachapen:
@moodlemobile Thank you for the feedback. We're interested in implementing the approach you're suggesting. Could we get in touch with your UX team to discuss the issue so that we can do it properly?

moodlemobile:
We should follow the discussion in https://tracker.moodle.org/browse/MOBILE-1636

jleyva:
Closing this, more information here: https://tracker.moodle.org/browse/MOBILE-1636

PR #725, opened 2016-11-09
MOBILE-1875 core: Run gulp tasks with ionic serve

martinmine:
This issue solves the issue where the gulp default and watch task are not executed on ionic-cli v2, which is installed by default when running `npm install -g ionic`. It also solves an issue where livereload reloads the page multiple times since both the files that are built by the gulp build task and the built files are watched by Ionic.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -356,6 +356,8 @@ var remoteAddonPaths = {
 
 gulp.task('default', ['build', 'sass', 'lang', 'config']);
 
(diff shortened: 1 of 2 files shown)


martinmine:
See issue [MOBILE-1875](https://tracker.moodle.org/browse/MOBILE-1875).

jleyva:
Thanks for working on this, this was integrated. 

More information here: https://tracker.moodle.org/browse/MOBILE-1875

PR #730, opened 2016-11-12
modified  it for customization for ARC

balamood2:


balamood2:
created the first customization test

PR #745, opened 2016-11-18
MOBILE-1479 Messages: Add copy functionality for messages

HaakonBj:
Solves [MOBILE-1479](https://tracker.moodle.org/browse/MOBILE-1479).
Test by sending a message and then pressing it for 500ms. A toast will appear with the message that the message was copied. Past wherever.


jleyva:
This was integrated: https://tracker.moodle.org/browse/MOBILE-1479
Thanks for working on this.

PR #769, opened 2016-11-30
MOBILE-1786 messages: Add unread message count in discussions

Gachapen:
Note that the count does not update in the discussion list after opening the discussion. The user has to pull down to refresh to get the updated count. At least this is a start for the feature.

Also note that it does one "mark as read" request per message. Maybe it would be better to use "mark all as read"? Like here: https://github.com/moodlehq/moodlemobile2/pull/729/files#diff-5ebdfd4a7501ca47e450e00dd02b7cf6R1104


jleyva:
Closing this as it has been already implemented. See https://tracker.moodle.org/browse/MOBILE-1786 for more information.

PR #911, opened 2017-02-17
Isea2017

san-2101:


san-2101:
hello sir ,
why my app icon isn't change i am do everything

san-2101:
hello sir,
i change everything but still moodle app icon is not change
how to solve my problem

PR #954, opened 2017-03-27
Pt-br translation for mod_assign.

santosleonardo:
Added Brazilian Portuguese translation for the Assignment activity.


jleyva:
Hello, thanks for working on this but translations are managed by https://lang.moodle.org/ you should manage the translation there as indicated in: https://docs.moodle.org/dev/Translating_Moodle_Mobile

PR #955, opened 2017-03-27
Some missing pt-br translations for competency addon.

santosleonardo:
Added a few missing pt-br translations for competency addon.


jleyva:
Hello, thanks for working on this but translations are managed by https://lang.moodle.org/ you should manage the translation there as indicated in: https://docs.moodle.org/dev/Translating_Moodle_Mobile

PR #1297, opened 2018-04-28
add translation of zh-cn

litong534:
I am using this program and i realize that the translation of zh-cn is unfinished,so i add them.


litong534:
This is the first time I commit the request, if I unfortunately make some mistake, tell me plz : )

PR #1761, opened 2019-02-11
Learnfuta

akinibrahim:


brions12:
Hello akinibrahim! 

I want to create an app for ios and android to join my website. My test website is hosted on moodlecloud.com

My problem is that I do not know what id I have to write. I searched the moodlecloud administration panel but I did not find my website's id.


How can I pass the parameters of my web to the app's moodle?

PR #1854, opened 2019-04-29
MOBILE-2925 install: Switch to using npx for package build

andrewnicols:


KriBetko:
Any news? I tried to test, so I have the same error with the "cordova-sqlite-storage" package in build.phonegap.com

PR #2316, opened 2020-03-13
Fix

jericsmith504:


Matt-Gleich:
How does this actually fix the app? This just looks like you are changing the data for your school.

PR #2631, opened 2020-11-28
Use docker-layer cache

LetsMelon:
Docker can cache each individual layer. 
If a layer changes, each subsequent layer is repeated. A file change counts as a change and therefore it is stupid to install the libraries after everything has been copied. Furthermore, they should be installed using ```npm ci```, otherwise ```package-lock.json``` will be ignored and errors may occur as a result.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,7 +17,9 @@ WORKDIR /app
 COPY package*.json ./
 RUN npm ci
 # Delete caches.
-RUN rm -rf /root/.npmCOPY . /app
+RUN rm -rf /root/.npm
+
+COPY . /app
 
 # Run gulp before starting.
 RUN npx gulp
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,10 +13,11 @@ EXPOSE 53703
 
 WORKDIR /app
 
-COPY . /app
-
 # Install npm libraries.
-RUN npm install && rm -rf /root/.npm
+COPY package*.json ./
+RUN npm ci
+# Delete caches.
+RUN rm -rf /root/.npmCOPY . /app
 
 # Run gulp before starting.
 RUN npx gulp


Dockerfile
```diff
@@ -13,10 +13,11 @@ EXPOSE 53703
...
 WORKDIR /app
 
-COPY . /app
-
 # Install npm libraries.
-RUN npm install && rm -rf /root/.npm
+COPY package*.json ./
+RUN npm ci
+# Delete caches.
+RUN rm -rf /root/.npmCOPY . /app
```
NoelDeMartin, line 20: I think you're missing a newline here, right? It should be:

```dockerfile
# Delete caches.
RUN rm -rf /root/.npm
COPY . /app
```

NoelDeMartin (changes requested):
I think that's it, once you fix that we can merge your changes :) Thanks!

Dockerfile
LetsMelon: Oh thanks, I'll fix this quickly.

NoelDeMartin (approved):

LetsMelon:
Thanks!

LetsMelon:
What's the status?

moodlemobile:
hi there, the original author of that file has provided this feedback:

1. You should be copying the package.lock file before running npm ci not just the Json file
2. I would keep the npm ci and rm commands on the same RUN to prevent the layer from increasing in size unnecessarily. That’s why we were doing that in the first place

NoelDeMartin:
Hey @LetsMelon thanks for your contribution!

We are working on a new version of the app in the [ionic5](https://github.com/moodlehq/moodleapp/tree/ionic5) branch, and since I've had to update the docker file I applied your suggestions there. I've added you as a co-author in a commit in this PR: #2725.

LetsMelon:
@NoelDeMartin Thank you!

PR #2729, opened 2021-04-29
Mobile 3744

crazyserver:


src/addons/calendar/components/calendar/addon-calendar-calendar.html
```diff
@@ -55,16 +55,23 @@ <h3>{{ periodName }}</h3>
                 <!-- In tablet, display list of events. -->
                 <div class="ion-hide-md-down addon-calendar-day-events">
                     <ng-container *ngFor="let event of day.filteredEvents | slice:0:4; let index = index">
-                        <p *ngIf="index < 3 || day.filteredEvents.length == 4" class="addon-calendar-event"
+                        <a *ngIf="index < 3 || day.filteredEvents.length == 4" class="addon-calendar-event"
```
NoelDeMartin, line 58: Is it ok to replace this with an anchor? I'm not sure if an anchor without an href attribute is valid markup, maybe we should use an unstyled button instead.

src/addons/calendar/pages/edit-event/edit-event.html
```diff
@@ -120,9 +120,10 @@ <h2 [core-mark-required]="true">
             </ng-container>
 
             <!-- Advanced options. -->
-            <ion-item-divider class="ion-text-wrap core-expandable" (click)="toggleAdvanced()">
-                <ion-icon *ngIf="!advanced" name="fas-caret-right" slot="start"></ion-icon>
-                <ion-icon *ngIf="advanced" name="fas-caret-down" slot="start"></ion-icon>
+            <ion-item-divider class="ion-text-wrap core-expandable" (click)="toggleAdvanced()"
```
NoelDeMartin, line 123: This isn't part of this issue, but I'll say it to keep in mind for later. This divider is not accessible because it cannot be focused and it doesn't have a role="button" (it should also respond to ENTER/SPACE keyboard, not just clicks). This happens in other places, so we should review that all click listeners are accessible.

src/core/features/editor/components/rich-text-editor/core-editor-rich-text-editor.html
```diff
@@ -9,105 +9,105 @@
...
         <ion-slide>
             <button [disabled]="!rteEnabled" [attr.aria-pressed]="toolbarStyles.p" (click)="buttonAction($event, 'p', 'block')"
                 (mousedown)="mouseDownAction($event)" [title]="'core.editor.p' | translate">
-                <ion-icon name="fas-paragraph"></ion-icon>
+                <ion-icon name="fas-paragraph" aria-hidden="true"></ion-icon>
             </button>
         </ion-slide>
         <ion-slide>
             <button [disabled]="!rteEnabled" [attr.aria-pressed]="toolbarStyles.h3" (click)="buttonAction($event, 'h3', 'block')"
                 (mousedown)="mouseDownAction($event)" [title]="'core.editor.h3' | translate">
```
NoelDeMartin, line 56: The title on these buttons should be converted to aria-label. I'm not sure if this is part of this issue, but since it's been done in other parts of the commit I'm mentioning it.

src/addons/messages/pages/group-conversations/group-conversations.html
```diff
@@ -27,14 +27,16 @@
...
                     <ion-label><h2>{{ 'addon.messages.contacts' | translate }}</h2></ion-label>
                     <ion-badge *ngIf="contactRequestsCount > 0" slot="end">{{contactRequestsCount}}</ion-badge>
                 </ion-item>
                 <!-- Favourite conversations. -->
-                <ion-item-divider class="ion-text-wrap core-expandable" (click)="toggle(favourites)" sticky="true">
-                    <ion-icon *ngIf="!favourites.expanded" name="fas-caret-right" slot="start"></ion-icon>
-                    <ion-icon *ngIf="favourites.expanded" name="fas-caret-down" slot="start"></ion-icon>
+                <ion-item-divider class="ion-text-wrap core-expandable" (click)="toggle(favourites)" sticky="true"
+                    [attr.aria-label]="(favourites.expanded ? 'core.collapse' : 'core.expand') | translate"
+                    [attr.aria-expanded]="favourites.expanded" role="heading">
```
NoelDeMartin, line 37: Is it ok to use role="heading" here? This has a click listener, so I think it should be role="button", be focusable and react to ENTER/SPACE keyboard.

src/addons/mod/chat/pages/sessions/sessions.html
```diff
@@ -31,7 +31,7 @@
             </ion-item>
 
             <ion-card *ngFor="let session of sessions.items" (click)="sessions.select(session)"
-                [class.core-selected-item]="sessions.isSelected(session)"
+                [attr.aria-current]="sessions.isSelected(session) ? 'page' :  'false'"
```
NoelDeMartin, line 34: Given that this pattern is so common when using a CorePageItemsListManager, I think we could create a new function that returns the `aria-current` value for an avoid the need to write a ternary every time.

src/core/features/grades/pages/courses/courses.html
```diff
@@ -22,7 +22,7 @@
                 <ion-item
                     *ngFor="let course of courses.items"
                     [title]="course.courseFullName"
-                    [class.core-selected-item]="courses.isSelected(course)"
+                    [attr.aria-current]="courses.isSelected(course) ? 'page' :  'false'"
```
NoelDeMartin, line 25: There is another CSS rule for selected items down this file that also needs to be updated.

NoelDeMartin (changes requested):
I've spotted some things to fix but overall it's good to go.

One comment I have regarding the approach to accessibility is that we should talk about implementing accessible components, instead of having to indicate all the accessibility attributes in pages. There is a lot of repetition going around, and it's very easy for a developer to forget some of these when they are working on a page.

src/addons/calendar/pages/edit-event/edit-event.html
crazyserver: I will include the usage of button property in ion-items and similar tags to make them tappable. I think ion-item-divider does not have button property, so probably I have to use aria roles.

src/core/features/editor/components/rich-text-editor/core-editor-rich-text-editor.html
crazyserver: Yep, I will take care of that, this is not an ion-button but a native button, I'll check if title does the trick.

src/addons/messages/pages/group-conversations/group-conversations.html
crazyserver: Probably it needs both roles. heading role is needed to apply aria-expanded.

src/addons/mod/chat/pages/sessions/sessions.html
crazyserver: Good catch, I'll have a look

src/core/features/editor/components/rich-text-editor/core-editor-rich-text-editor.html
crazyserver: Finally title is better than aria-label in this case, it will define an accessible name and also show a tooltip.
+info https://dev.opera.com/articles/ux-accessibility-aria-label/#accessible-name-calculation

src/core/classes/page-items-list-manager.ts
```diff
@@ -121,6 +121,16 @@ export abstract class CorePageItemsListManager<Item> {
         return this.selectedItem === item;
     }
 
+    /**
+     * Return the current aria role.
+     *
+     * @param item Item.
+     * @return Will return the current role of the item if selected, false otherwise.
+     */
+    getSelectedAriaCurrent(item: Item): string {
```
NoelDeMartin, line 130: I think this should be called `getItemAriaCurrent` instead, or some other variation that does not include the concept of "selected". This is because we'll call this method for all items, and the fact that the returned value is different for selected items is an implementation detail.

Also, I've been looking at [the spec](https://www.w3.org/TR/wai-aria-1.1/#aria-current) for `aria-current` and it doesn't make any reference to "roles", so the value of this attribute is not a role (so the function docs should not mention it either).

src/addons/mod/forum/pages/new-discussion/new-discussion.html
```diff
@@ -31,11 +31,10 @@
                     (contentChanged)="onMessageChange($event)">
                 </core-rich-text-editor>
             </ion-item>
-            <ion-item-divider class="ion-text-wrap core-expandable" (click)="toggleAdvanced()">
-                <ion-icon *ngIf="!advanced" name="fa-caret-right" slot="start">
-                </ion-icon>
-                <ion-icon *ngIf="advanced" name="fa-caret-down" slot="start">
-                </ion-icon>
+            <ion-item-divider class="ion-text-wrap core-expandable" (click)="toggleAdvanced()" [attr.aria-expanded]="advanced"
+                [attr.aria-label]="(advanced ? 'core.hideadvanced' : 'core.showadvanced') |translate" role="heading button">
```
NoelDeMartin, line 35: We're using two roles here: heading and button. It seems like it's not possible to use more than one role per element. According to [this spec](https://www.w3.org/TR/2014/REC-wai-aria-implementation-20140320/#mapping_role), it is a valid attribute value but multiple roles will only work as fallback, not all at once.

src/core/features/grades/pages/grade/grade.html
```diff
@@ -16,17 +16,18 @@
         <ion-list *ngIf="grade">
             <ion-item *ngIf="grade.itemname && grade.link" class="ion-text-wrap" detail="true" [href]="grade.link" core-link
             capture="true">
-                <ion-icon *ngIf="grade.icon" name="{{grade.icon}}" slot="start"></ion-icon>
-                <img *ngIf="grade.image" [src]="grade.image" slot="start" class="core-module-icon" />
+                <ion-icon *ngIf="grade.icon" name="{{grade.icon}}" slot="start" [attr.aria-label]="grade.iconAlt"></ion-icon>
+                <img *ngIf="grade.image" [src]="grade.image" slot="start" class="core-module-icon"
+                    [alt]="grade.iconAlt" role="presentation">
```
NoelDeMartin, line 21: Using a role="presentation" here will hide this image from screen readers, but it has an alt attribute. Was that your intention or is this just a typo? I'm asking to make sure because in all other instances with role="presentation" we're using alt="", unless I missed something.

src/addons/calendar/components/calendar/addon-calendar-calendar.html
```diff
@@ -55,16 +55,23 @@ <h3>{{ periodName }}</h3>
                 <!-- In tablet, display list of events. -->
                 <div class="ion-hide-md-down addon-calendar-day-events">
                     <ng-container *ngFor="let event of day.filteredEvents | slice:0:4; let index = index">
-                        <p *ngIf="index < 3 || day.filteredEvents.length == 4" class="addon-calendar-event"
-                        (click)="eventClicked(event, $event)" [class.addon-calendar-event-past]="event.ispast">
+                        <div role="button" *ngIf="index < 3 || day.filteredEvents.length == 4" class="addon-calendar-event"
+                            (click)="eventClicked(event, $event)" [class.addon-calendar-event-divast]="event.ispast">
```
NoelDeMartin, line 59: I think this is a typo, you didn't mean to change `addon-calendar-event-past` to `addon-calendar-event-divast`.

src/addons/mod/forum/components/post/post.html
```diff
@@ -120,12 +122,11 @@ <h3 *ngIf="post.author && post.author.fullname">{{post.author.fullname}}</h3>
             <ion-checkbox slot="end" [(ngModel)]="replyData.isprivatereply" name="isprivatereply"></ion-checkbox>
         </ion-item>
         <ng-container *ngIf="forum.id && forum.maxattachments > 0">
-            <ion-item-divider class="core-expandable ion-text-wrap" (click)="toggleAdvanced()">
+            <ion-item-divider class="core-expandable ion-text-wrap" (click)="toggleAdvanced()" [attr.aria-expanded]="advanced"
+                [attr.aria-label]="(advanced ? 'core.hideadvanced' : 'core.showadvanced') |translate" role="button">
```
NoelDeMartin, line 126: This may not be part of this issue but just for the record, even if this element has the "button" role it still not accessible because it doesn't receive focus, and so cannot react to keyboard events. We can look into this in [MOBILE-3753](https://tracker.moodle.org/browse/MOBILE-3753), but keep it in mind.

NoelDeMartin (changes requested):
I found a couple more things, but we can proceed by fixing typos and checking the rest in other issues.

src/core/features/grades/pages/grade/grade.html
crazyserver: Good catch thanks!

src/addons/calendar/components/calendar/addon-calendar-calendar.html
crazyserver: True thx!

src/addons/mod/forum/components/post/post.html
crazyserver: That's right! We'll have a look then

src/addons/mod/forum/pages/new-discussion/new-discussion.html
crazyserver: As a fallback will be ok, because anyway I would leave it as heading. We'll have to add keyboard focus on following issues

crazyserver:
Thanks Noel. I removed the presentation role on img tag on grade, also solved the typo on calendar and changed the aria current function and jsdoc.

src/addons/mod/forum/pages/new-discussion/new-discussion.html
NoelDeMartin: Ok, I'm going forward with the merge because buttons are not the focus of this issue. But I think this will be a problem, because if it has a click listener it should have a button role, not heading. If it's really serving as both I guess we should use two separate elements, I'm not sure. Let's check it out on the issue about buttons.

NoelDeMartin (approved):

PR #2745, opened 2021-05-05
Mobile 3713

dpalou:


src/core/components/iframe/core-iframe.html
```diff
@@ -5,7 +5,11 @@
         [attr.allowfullscreen]="allowFullscreen ? 'allowfullscreen' : null">
     </iframe>
 
+    <div *ngIf="displayHelp" class="ion-margin ion-text-center ion-text-wrap core-iframe-help">
+        <span (click)="openIframeHelpModal()" aria-haspopup="dialog" role="button">{{ 'core.iframehelp' | translate }}</span>
```
crazyserver, line 9: Role button is not enough to correctly handle a11y. 
Explanation found:
The custom control should implement onclick as well as onKeyDown events. Buttons can be activated through spacebar. Hence, if you are adding the role to a custom control, you need to handle these events yourself. Else, the semantic loses its meaning. A screen-reader user will try to activate the button using spacebar, but cannot.

 I would suggest to change it to a default/native button and style it properly or handle onKeyDown (or Up) to check space and enter keys when focused.

src/core/directives/format-text.ts
```diff
@@ -757,6 +761,33 @@ export class CoreFormatTextDirective implements OnChanges {
...
+        const helpDiv = document.createElement('div');
+
+        helpDiv.classList.add('ion-margin');
+        helpDiv.classList.add('ion-text-center');
+        helpDiv.classList.add('ion-text-wrap');
+        helpDiv.classList.add('core-iframe-help');
+
+        const span = document.createElement('span');
+        span.innerHTML = Translate.instant('core.iframehelp');
+        span.setAttribute('role', 'button');
```
crazyserver, line 779: Role button is not enough to correctly handle a11y. 
Explanation found:
The custom control should implement onclick as well as onKeyDown events. Buttons can be activated through spacebar. Hence, if you are adding the role to a custom control, you need to handle these events yourself. Else, the semantic loses its meaning. A screen-reader user will try to activate the button using spacebar, but cannot.

 I would suggest to change it to a default/native button and style it properly or handle onKeyDown (or Up) to check space and enter keys when focused.

src/core/lang.json
```diff
@@ -213,6 +215,7 @@
     "openfullimage": "Click here to display the full size image",
     "openinbrowser": "Open in browser",
     "openmodinbrowser": "Open {{$a}} in browser",
+    "opensettings": "Open settings",
```
crazyserver, line 218: Please add new strings to langindex.json

crazyserver (changes requested):
Code looks good but I found minor a11y changes.

src/core/components/iframe/core-iframe.html
crazyserver: Also, when using a custom element you should add tabindex=0 so is keyboard accessible.

And finally, you will also need to add :focus and :hover css. It could be added using span[role="button"]:focus so it solves some others.

dpalou:
Thanks Pau,

I modified both HTML codes and I added the strings to langindex. I also had to modify the rule in theme.base.scss. Can you please take a look again?

crazyserver (approved):
All ok!
https://tracker.moodle.org/browse/MOBILE-3713

PR #2746, opened 2021-05-06
MOBILE-3320: Tweaks

NoelDeMartin:


src/core/singletons/time.ts
```diff
@@ -0,0 +1,39 @@
...
+    /**
+     * Wrap a function so that it is called only once.
+     *
+     * @param fn Function.
+     * @return Wrapper that will call the underlying function only once.
+     */
+    static once<T extends unknown[]>(fn: (...args: T) => unknown): (...args: T) => void {
+        let called = false;
+
+        return (...args: unknown[]) => {
```
dpalou, line 29: I think args should have type T.

dpalou (changes requested):

src/core/singletons/time.ts
NoelDeMartin: Done!

PR #2747, opened 2021-05-06
Mobile 3745

crazyserver:


src/addons/mod/forum/components/index/index.html
```diff
@@ -67,15 +67,16 @@
...
-            <div *ngIf="!discussions.empty && sortingAvailable && selectedSortOrder" class="ion-text-wrap addon-forum-sorting-select">
-                <ion-button *ngIf="sortingAvailable" id="addon-mod-forum-sort-order-button"
-                    class="core-button-select button-no-uppercase"
-                    aria-haspopup="true" aria-controls="addon-mod-forum-sort-order-selector"
-                    [attr.aria-label]="('core.sort' | translate)"
-                    (click)="showSortOrderSelector()">
-                    <span class="core-button-select-text">{{ selectedSortOrder.label | translate }}</span>
-                    <div class="select-icon" slot="end"><div class="select-icon-inner"></div></div>
-                </ion-button>
+            <div *ngIf="!discussions.empty && sortingAvailable && selectedSortOrder"
```
dpalou, line 70: There's no need to split the line in here.

src/addons/mod/forum/components/index/index.html
```diff
@@ -67,15 +67,16 @@
...
+                class="ion-text-wrap">
+                <core-combobox *ngIf="sortingAvailable"
```
dpalou, line 72: sortingAvailable is already checked in the parent element, no need to do it again.

src/addons/mod/forum/components/index/index.html
```diff
@@ -67,15 +67,16 @@
...
+                    [modalOptions]="sortOrderSelectorModalOptions()"
```
dpalou, line 73: Have you tried if you can pass an object instead of a function? I think passing objects have a better performance than using a function. You'd need to store the "ModalOptions" in a property.

src/core/features/course/components/format/core-course-format.html
```diff
@@ -19,18 +19,15 @@
                 <div *ngIf="displaySectionSelector && sections && hasSeveralSections"
                     class="ion-text-wrap clearfix ion-justify-content-between core-button-selector-row"
                     [class.core-section-download]="downloadEnabled">
-                    <ion-button class="core-button-select button-no-uppercase" (click)="showSectionSelector()"
-                        aria-haspopup="listbox" id="core-course-section-button" expand="block">
-                        <ion-icon name="fas-folder" slot="start" aria-hidden="true"></ion-icon>
-                        <span class="sr-only" *ngIf="selectedSection">{{ 'core.course.sections' | translate }}:</span>
-                        <span class="core-button-select-text">
+                    <core-combobox [modalOptions]="sectionSelectorModalOptions()" interface="modal" listboxId="core-course-section-button"
```
dpalou, line 22: Have you tried if you can pass an object instead of a function? I think passing objects have a better performance than using a function. You'd need to store the "ModalOptions" in a property.

src/core/services/utils/dom.ts
```diff
@@ -1681,6 +1681,26 @@ export class CoreDomUtilsProvider {
         });
     }
 
+    /**
+     * Opens a Modal.
+     *
+     * @param opts Modal Options.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    async openModal<T = any>(
```
dpalou, line 1689: Can you use unknown instead of any?

src/core/features/sharedfiles/services/sharedfiles-helper.ts
```diff
@@ -150,16 +150,13 @@ export class CoreSharedFilesHelperProvider {
      * @return Promise resolved when a file is picked, rejected if file picker is closed without selecting a file.
      */
     async pickSharedFile(mimetypes?: string[]): Promise<CoreFileUploaderHandlerResult> {
-        const modal = await ModalController.create({
+        const modalData = await CoreDomUtils.openModal({
```
dpalou, line 153: Shouldn't you pass the FileEntry type here instead of casting it later?

src/core/services/utils/dom.ts
```diff
@@ -1684,23 +1684,45 @@ export class CoreDomUtilsProvider {
...
         }
     }
 
+    /**
+     * Opens a side Modal.
+     *
+     * @param modalOptions Modal Options.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    async openSideModal<T = any>(
```
dpalou, line 1710: Like above, can you use unknown instead of any?

src/core/services/utils/dom.ts
```diff
@@ -1723,6 +1722,27 @@ export class CoreDomUtilsProvider {
         return await this.openModal<T>(modalOptions);
     }
 
+    /**
+     * Opens a popover.
+     *
+     * @param popoverOptions Modal Options.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    async openPopover<T = any>(
```
dpalou, line 1731: Like above, can you use unknown instead of any?

src/core/features/mainmenu/pages/menu/menu.ts
```diff
@@ -203,6 +205,90 @@ export class CoreMainMenuPage implements OnInit, OnDestroy {
         }
     }
 
+    /**
+     * A11y key functionallity that prevents keyDown events.
+     *
+     * @param e Event.
+     */
+    keyDownAction(e: KeyboardEvent): void {
+        if (e.key == ' ' ||
```
dpalou, line 214: Maybe we should have a function to compare this, it's the second time we have this code. That way it's more reusable and easy to change.

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -31,12 +32,12 @@ import { CoreNavigator } from '@services/navigator';
     templateUrl: 'core-user-avatar.html',
     styleUrls: ['user-avatar.scss'],
 })
-export class CoreUserAvatarComponent implements OnInit, OnChanges, OnDestroy {
+export class CoreUserAvatarComponent extends CoreAriaRoleButton implements OnInit, OnChanges, OnDestroy {
```
dpalou, line 35: Like Noel always says, consider composition over inheritance. If in the future this class (or another) needs to inherit from AriaRoleButton and another class at the same time we won't be able to do it. IMO it would be better to do it like with ItemsManager, create a button instance inside this class and use it.

dpalou (changes requested):

src/core/services/utils/dom.ts
crazyserver: Thanks I preffered to use void instead.

src/core/services/utils/dom.ts
crazyserver: Thanks! I finally used a wide used type void instead.

src/core/features/mainmenu/pages/menu/menu.ts
crazyserver: I did the same as in button role to have a wide proper solution.

src/core/components/user-avatar/user-avatar.ts
crazyserver: Solved

crazyserver:
Thanks Dani, can you review again?

I added a new commit to change some colors and add some vars

src/core/features/mainmenu/pages/menu/menu.ts
```diff
@@ -93,6 +98,8 @@ export class CoreMainMenuPage implements OnInit, OnDestroy {
                 }
             });
         }
+
+        this.tabAction = new CoreMainmenuRoleTab(this);
```
dpalou, line 102: This second tabAction isn't needed, right?

src/core/features/mainmenu/pages/menu/menu.ts
```diff
@@ -246,3 +254,42 @@ export class CoreMainMenuPage implements OnInit, OnDestroy {
     }
 
 }
+
+/**
+ * Helper class to manage rol tab.
+ */
+class CoreMainmenuRoleTab extends CoreAriaRoleTab<CoreMainMenuPage> {
```
dpalou, line 261: Main menu classes are usually called "MainMenu", not "Mainmenu".

dpalou (changes requested):

crazyserver:
Thanks Dani! Now it should be ok!

PR #2757, opened 2021-05-11
Mobile 3753

crazyserver:


src/addons/messages/pages/group-conversations/group-conversations.html
```diff
@@ -25,21 +25,32 @@
...
+                <ion-item
+                    button
+                    class="ion-text-wrap core-expandable divider"
+                    (click)="toggle(favourites)" sticky="true"
                     [attr.aria-label]="(favourites.expanded ? 'core.collapse' : 'core.expand') | translate"
-                    [attr.aria-expanded]="favourites.expanded" role="heading button">
+                    [attr.aria-expanded]="favourites.expanded"
+                    role="heading"
+                    detail="false"
+                >
```
NoelDeMartin, line 45: I've been looking into the `aria-expanded` attribute and I don't think this is the correct way to use it. As I understand it, it should make reference to the thing that is expanded/collapsed, which in this case is the list below, not this button/header. This can be combined with the `aria-controls` attribute to refer the list by an id, but even in that case the aria attributes should appear on the button, not on the header, so I think we should take another look at how to solve this.

For reference, I've found an aria design pattern that solves this use-case, they call it "Accordion": https://www.w3.org/TR/wai-aria-practices/#accordion

src/addons/calendar/components/calendar/addon-calendar-calendar.html
```diff
@@ -29,58 +29,88 @@ <h3>{{ periodName }}</h3>
...
+                    *ngFor="let day of week.days"
+                    class="addon-calendar-day ion-text-center"
+                    [ngClass]='{
+                        "hasevents": day.hasevents,
+                        "today": isCurrentMonth && day.istoday,
+                        "weekend": day.isweekend,
+                        "duration_finish": day.haslastdayofevent
+                    }'
+                    [class.addon-calendar-event-past-day]="isPastMonth || day.ispast"
+                    role="button cell"
```
NoelDeMartin, line 59: I don't think it's possible to use two roles for the same element. The specs allow the `role` attribute to have multiple values, but these are treated as fallbacks, not both at the same time. Keeping that in mind, I think we should keep the "cell" role on the `ion-col` element and create the button inside.

src/addons/calendar/components/calendar/calendar.ts
```diff
@@ -525,3 +535,31 @@ export class AddonCalendarCalendarComponent implements OnInit, DoCheck, OnDestro
...
+class AddonCalendarDayButton extends CoreAriaRoleButton<AddonCalendarCalendarComponent> {
+
+    /**
+     * @inheritdoc
+     */
+    click(event: Event, day: number): void {
+        this.componentInstance.dayClicked(day);
+    }
+
+}
```
NoelDeMartin, line 551: I think this approach can become cumbersome if we use this pattern a lot in the app, would it be possible to use a directive or some other approach that simplifies it? For example, something like this would be nice and it conveys the same information:

```
<div [core-aria-button.click]="dayClicked(day)">...</div>
```

I'm not sure about the syntax, but I think that type of approach should work.

src/addons/messages/pages/group-conversations/group-conversations.html
```diff
@@ -25,21 +25,32 @@
 
         <core-loading [hideUntil]="loaded" [message]="loadingMessage">
             <ion-list>
-                <ion-item class="ion-text-wrap addon-message-discussion" (click)="gotoContacts()"
-                    [attr.aria-label]="'addon.messages.contacts' | translate" detail="true" button>
+                <ion-item class="ion-text-wrap addon-message-discussion" (click)="gotoContacts()" detail="true" button>
                     <ion-icon name="fas-address-book" slot="start" aria-hidden="true"></ion-icon>
                     <ion-label><h2>{{ 'addon.messages.contacts' | translate }}</h2></ion-label>
-                    <ion-badge *ngIf="contactRequestsCount > 0" slot="end">{{contactRequestsCount}}</ion-badge>
+                    <ion-badge slot="end" aria-hidden="true">{{contactRequestsCount}}</ion-badge>
```
NoelDeMartin, line 31: I think you didn't mean to remove the `*ngIf="contactRequestsCount > 0"` part from here, right? This will show a badge even with 0 requests for non-screen reader users, and that wasn't be original behaviour.

NoelDeMartin (changes requested):

src/addons/messages/pages/group-conversations/group-conversations.html
crazyserver: I think using aria-controls will be enough for now. Let's see what happens

src/addons/calendar/components/calendar/addon-calendar-calendar.html
crazyserver: I think combining roles will be enough. I annotated the question to ask an expert.

src/addons/calendar/components/calendar/calendar.ts
crazyserver: This is a great idea. Done!!

There's a similar problem with CoreAriaRoleTab class but I think it's not that easy to move to a directive. However, feel free to improve it if you like!

NoelDeMartin (approved):
There are still some potential issues that we've discussed about using aria roles and attributes, but I'll go forward merging this issue given that they don't break any existing functionality. We can review them on a separate issue when we go over a11y again.

PR #2762, opened 2021-05-13
MOBILE-3750: Improve accessibility of horizontal scrolls

NoelDeMartin:


crazyserver (changes requested):
Hi Noel!

Changes are ok, but I found that RTL is not supported:
You'll need to flip the buttons order. You can easily do that using flex instead of inline block. I think it will be automatically reversed.

NoelDeMartin:
Thanks @crazyserver for the review, good catch I didn't think about RTL at all. Other than that, I've also done a rebase and fixed a bug where multiple instances of a component with an horizontal scroll wouldn't work properly. Please take another look.

src/addons/block/recentlyaccessedcourses/components/recentlyaccessedcourses/recentlyaccessedcourses.ts
```diff
@@ -62,6 +63,10 @@ export class AddonBlockRecentlyAccessedCoursesComponent extends CoreBlockBaseCom
      * Component being initialized.
      */
     async ngOnInit(): Promise<void> {
+        // Generate unique id for scroll element.
+        const scrollId = CoreUtils.getUniqueId('AddonBlockStarredCoursesComponent-Scroll');
```
crazyserver: You are using AddonBlockStarredCoursesComponent on AddonBlockRecentlyAccessedCoursesComponent

crazyserver (approved):

PR #2774, opened 2021-05-18
Mobile 3693

crazyserver:


dpalou:
Hi Pau,

the changes look good. Have you checked if having audios/videos inside radio buttons or similar work? It was a common use case in our users, we should make sure it works.

crazyserver:
NIce catch Dani! Should be fixed now. I also added some fixes to css I found during the testing.

dpalou (approved):

PR #2775, opened 2021-05-19
Mobile 3691

dpalou:


NoelDeMartin (approved):
At some point we should stop using this pattern of depending on plugins with `git+https://`, and as we've been discussing we should explore the possibility to have custom native code in this same repository (be it implementing a plugin here, following the same strategy we are following for `MainActivity.java`, or something else).

But I'll go ahead with the merge because this solves the issue for now, and this is a pattern we are already using with other plugins.

PR #2782, opened 2021-05-20
MOBILE-3751: Headings a11y

NoelDeMartin:


crazyserver (changes requested):
Hi Noel!

Code looks good but I think it would be better to directly use h1 instead of ion-title or core-title. We can add css ion-header h1 { } or similar.

Also, check styles from ios and android (https://github.com/ionic-team/ionic-framework/tree/master/core/src/components/title)

NoelDeMartin:
@crazyserver Thanks for the review, I followed your suggestion and I also noticed that some headers should actually be h2 (the ones inside modals).

crazyserver (approved):
Great changes!

PR #2788, opened 2021-05-21
Mobile 3320

crazyserver:


dpalou (approved):

dpalou:
Hi Pau,

the changes look good to me, but it seems some E2E tests are failing. Maybe they only look for text in aria-label and not in title?

crazyserver:
Thanks Dani! Can you have a look again?

src/core/features/comments/pages/viewer/viewer.page.ts
```diff
@@ -262,48 +275,65 @@ export class CoreCommentsViewerPage implements OnInit, OnDestroy {
     }
 
     /**
-     * Add a new comment to the list.
+     * Send the comment or store it offline.
      *
-     * @param e Event.
+     * @param text Comment to add
+      Event.
```
dpalou, line 281: This line should be removed.

src/core/features/comments/pages/viewer/viewer.page.ts
```diff
@@ -160,9 +164,15 @@ export class CoreCommentsViewerPage implements OnInit, OnDestroy {
                     commentsResponse.comments.length >= CoreCommentsProvider.pageSize;
             }
 
-            comments = await Promise.all(comments.map((comment) => this.loadCommentProfile(comment)));
+            this.comments = comments.concat(this.comments);
```
dpalou, line 167: This change means that we will display comments without user data, and then the user data will be loaded. Wouldn't it be better to wait for all data to be loaded before displaying them, as it was done before?

dpalou (changes requested):

crazyserver:
Thanks Dani! Solved! I also changed the way it scrolls to bottom to avoid scrolling when the page is partially loaded.

crazyserver:
Also added a new commit to prevent new line from being shown before submitting from send message form

dpalou (approved):

PR #2794, opened 2021-05-27
Mobile 3761

dpalou:


src/core/constants.ts
```diff
@@ -173,6 +173,7 @@ export interface EnvironmentConfig {
     displayqroncredentialscreen?: boolean;
     displayqronsitescreen?: boolean;
     forceOpenLinksIn: 'app' | 'browser';
+    iosopenfilepicker?: boolean;
```
NoelDeMartin, line 176: We were using all lowercase characters for historical purposes, but now that we already have camel cased properties, could we call this one `iOSOpenFilePicker` instead?

Other than that, what do you think of using a string instead of a boolean here? We could call it something like `defaultFilePickerAction` and it could have the values "open" or "open-with" (iOS only). I think the code would be more readable because you would see something like `openFile('open');' and `openFile('open-with');`.

In any case, these are just suggestions so feel free to apply the ones you like.

src/core/components/local-file/core-local-file.html
```diff
@@ -18,18 +18,25 @@ <h3 *ngIf="!editMode">{{fileName}}</h3>
...
-                <ion-icon name="fas-pen" slot="icon-only" aria-hidden="true"></ion-icon>
+            <ion-button fill="clear" *ngIf="isIOS" (click)="openFile($event, true)"
+                [attr.aria-label]="openButtonLabel | translate">
+                <ion-icon slot="icon-only" [name]="openButtonIcon" aria-hidden="true"></ion-icon>
             </ion-button>
 
-            <ion-button *ngIf="editMode" fill="clear" [attr.aria-label]="'core.save' | translate" color="success" type="submit">
-                <ion-icon name="fas-check" slot="icon-only" aria-hidden="true"></ion-icon>
-            </ion-button>
+            <ng-container *ngIf="manage">
```
NoelDeMartin, line 26: This ng-container is unnecessary, given that this is already inside of a div with the same `ngIf` condition.

NoelDeMartin (changes requested):
I found one improvement and a small fix, other than that I think it's pretty much ready for merging.

src/core/components/local-file/core-local-file.html
dpalou: Nice catch, I forgot to remove the ngIf from the parent div.

src/core/constants.ts
dpalou: I added a new commit with these changes.

crazyserver (approved):

PR #2795, opened 2021-05-27
MOBILE-3320: Support alerts on top of loading

NoelDeMartin:


src/core/classes/ion-loading.ts
```diff
@@ -13,43 +13,148 @@
...
+     * Dismiss the loading element.
+     *
+     * @param data Dismiss data.
+     * @param role Dismiss role.
+     */
+    async dismiss(data?: unknown, role?: string): Promise<void> {
+        if (this.paused) {
+            this.paused = false;
+
+            return;
```
dpalou, line 46: Shouldn't you call the dismiss listeners even if the modal is paused or hasn't been displayed yet? Otherwise some listeners might never be called.

BTW I found a bit hard to understand the code at first, maybe it would be nice to include some comments like "No need to dismiss the modal if it's paused since it's already hidden" and things like that.

src/core/services/utils/dom.ts
```diff
@@ -1480,19 +1488,19 @@ export class CoreDomUtilsProvider {
...
-        await alert.present();
-
         const isDevice = CoreApp.isAndroid() || CoreApp.isIOS();
         if (!isDevice) {
             // Treat all anchors so they don't override the app.
             const alertMessageEl: HTMLElement | null = alert.querySelector('.alert-message');
             alertMessageEl && this.treatAnchors(alertMessageEl);
         }
+
+        await alert.onDidDismiss();
```
dpalou, line 1503: Do we really want to wait for the modal to be dismissed? The function name and JSDoc doesn't say anything about it. Currently we never wait for this function to be resolved so this change shouldn't break anything, just saying.

dpalou (changes requested):
Hi Noel, I added a couple of comments in the code but I also want to comment the general approach of supporting alerts on top of loading. Couldn't you just put a higher z-index to the alerts so they're displayed on top of the loading modals?

src/core/services/utils/dom.ts
dpalou: Ok, I've just seen the result will be used now, so ignore this comment.

src/core/classes/ion-loading.ts
NoelDeMartin: Thanks for spotting that, you're right we should call the listeners here as well, I fixed it now.

I also added a new comment explaining the purpose of the pause/resume.

NoelDeMartin:
Thanks for the review @dpalou, I suppose it would be possible to use the z-index approach but I decided to do it this way because stacking modals on top of one another makes the overlay ever darker. With this approach, every time a modal is displayed the previous modal will be hidden visually but remain active conceptually, which I think makes better UX.

dpalou (approved):

PR #2799, opened 2021-05-28
Mobile 3320

crazyserver:


src/addons/qtype/ddimageortext/component/addon-qtype-ddimageortext.html
```diff
@@ -15,10 +15,13 @@
                 [contextLevel]="contextLevel" [contextInstanceId]="contextInstanceId" [courseId]="courseId"
                 (afterRender)="textRendered()">
             </core-format-text>
-
+        </ion-label>
+    </ion-item>
+    <ion-item class="ion-text-wrap" [hidden]="!ddQuestion.loaded">
```
dpalou, line 20: Have you verified if this change works? I remember doing something similar and then I couldn't drag items from one item to the other, I had to put all the drags and drops in the same ion-item.

The same thing goes for all the other DD question types.

dpalou (changes requested):

src/addons/qtype/ddimageortext/component/addon-qtype-ddimageortext.html
crazyserver: Hey! Yes, it's verified, as you can see this change is not done on ddwtos because there it broke the question.

dpalou (approved):

PR #2803, opened 2021-06-01
Mobile 3320

crazyserver:


src/core/directives/format-text.ts
```diff
@@ -80,6 +80,7 @@ export class CoreFormatTextDirective implements OnChanges {
     @Input() maxHeight?: number;
 
     @Output() afterRender: EventEmitter<void>; // Called when the data is rendered.
+    @Output() action: EventEmitter<void> = new EventEmitter(); // Called when clicked.
```
dpalou, line 83: The name "action" doesn't convince me. What happens if in the future we have another way to interact with format-text? We'd have two "actions". Maybe it would be better to name it "onClick" like we do in core-local-file.

dpalou (changes requested):

crazyserver:
Thanks Dani! Changed!

dpalou (approved):

PR #2804, opened 2021-06-01
MOBILE-3320 core: Change some icons to fontawesome

crazyserver:


src/core/features/sitehome/pages/index/index.html
```diff
@@ -53,7 +53,7 @@
                         </ng-container>
                     </ng-container>
                 </ion-list>
-            <core-empty-box *ngIf="!hasContent" icon="qr-scanner" [message]="'core.course.nocontentavailable' | translate">
+            <core-empty-box *ngIf="!hasContent" icon="fas-qrcode" [message]="'core.course.nocontentavailable' | translate">
```
dpalou, line 56: IIRC the qr-scanner icon wasn't a qrcode, it was a square with some missing parts, like the one used by scanner apps to indicate the area where to put the QR code. So IMO we should find another icon for these cases instead of qrcode.

dpalou (changes requested):

src/core/features/sitehome/pages/index/index.html
crazyserver: True! I've replaced that with an open box (on 2 places).

dpalou (approved):

PR #2813, opened 2021-06-07
Mobile 3320

crazyserver:


src/core/features/settings/services/settings-helper.ts
```diff
@@ -54,6 +54,15 @@ export const enum CoreZoomLevel {
     HIGH = 'high',
 }
 
+/**
+ * Listener function to toggle dark mode.
+ *
+ * @param e Event object.
+ */
+const toggleDarkModeListener = (e: MediaQueryListEvent): void => {
```
dpalou, line 62: Why do we need to move this function outside of the class?

src/core/features/styles/services/styles.ts
```diff
@@ -207,123 +310,30 @@ export class AddonRemoteThemesProvider {
...
-            if (state == CoreConstants.NOT_DOWNLOADED) {
-                // File not downloaded, URL has changed or first time. Delete downloaded CSS files.
-                await CoreFilepool.removeFilesByComponent(siteId, AddonRemoteThemesProvider.COMPONENT, 1);
-            }
-        } catch {
-            // An error occurred while getting state (shouldn't happen). Don't delete downloaded file.
-        }
-
-        return CoreFilepool.downloadUrl(siteId, url, false, AddonRemoteThemesProvider.COMPONENT, 1);
+        CoreApp.setStatusBarColor();
```
dpalou, line 320: I think this line will be called too much. disableStyleElementsBySelector is usually called in an iterator, this means it will probably be called several times in a row. Since it's a protected method I think it makes more sense to move this line to the functions that call this one, and it's already done like this. This function is only used in the clear method, and that method already calls CoreApp.setStatusBarColor.

src/core/features/styles/services/styles.ts
```diff
@@ -158,26 +251,40 @@ export class AddonRemoteThemesProvider {
...
+    /**
+     * Disabled an element based on site and source name.
+     *
+     * @param siteId Site Id.
+     * @param sourceName Source or handler name.
+     * @param disable Whether to disable or enable the element.
      */
-    protected createStyleElement(id: string, disabled: boolean): HTMLStyleElement {
-        const styleEl = document.createElement('style');
+    disableStyleElementByName(siteId: string, sourceName: string, disable: boolean): void {
```
dpalou, line 280: Since disableStyleElementsBySelector is protected, maybe disableStyleElementByName and disableStyleElement should be protected too. Otherwise it could eventually happen that some external code disables some style but doesn't call CoreApp.setStatusBarColor. It depends on how you think this delegate should be used.

If you think we need a public method to disable styles, then I'd suggest creating a new public method that disables the style and calls CoreApp.setStatusBarColor. However, this service internally should keep using the protected methods that don't call CoreApp.setStatusBarColor, that way we prevent calling it too much.

dpalou (changes requested):

src/core/features/settings/services/settings-helper.ts
crazyserver: Done!

src/core/features/styles/services/styles.ts
crazyserver: Done; thanks!

src/addons/remotethemes/services/remotethemes-handler.ts
```diff
@@ -0,0 +1,144 @@
...
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    async isEnabled(siteId: string, config?: CoreSitePublicConfigResponse): Promise<boolean> {
+        return true;
+    }
+
+    /**
+     * @inheritDoc
+     */
+    async getStyle(siteId: string, config?: CoreSitePublicConfigResponse): Promise<string> {
+        if (siteId == CoreStylesService.TMP_SITE_ID && config) {
```
dpalou, line 58: IMO if siteId is TMP_SITE_ID but no config is received then you should stop the execution. Otherwise it will call getSite for a site that will never exist.

dpalou (changes requested):

src/addons/remotethemes/services/remotethemes-handler.ts
crazyserver: good catch, solved!

dpalou (approved):

PR #2814, opened 2021-06-08
MOBILE-3320: Forum improvements

NoelDeMartin:


src/addons/mod/forum/components/index/index.ts
```diff
@@ -726,6 +727,10 @@ class AddonModForumDiscussionsManager extends CorePageItemsListManager<Discussio
         this.discussionsPathPrefix = discussionsPathPrefix;
     }
 
+    get loaded(): boolean {
+        return super.loaded && (!CoreApp.isOnline() || this.onlineLoaded || this.fetchFailed);
```
dpalou, line 731: I think the CoreApp.isOnline() condition shouldn't be here. When the user is offline we can still load online discussions (if they're cached), so the app would display the contents before the cached data is displayed.

dpalou (changes requested):

src/addons/mod/forum/components/index/index.ts
NoelDeMartin: Thanks for spotting that, while looking at this I noticed there was another bug; when the online discussions array was empty `onlineLoaded` wouldn't ever be set to `true`, so that's probably what tripped me up doing some tests before. I also fixed it so it should be fine now.

dpalou (approved):

PR #2837, opened 2021-06-17
Mobile 3320

dpalou:


crazyserver (changes requested):
Code looks good but there are missing strings on langindex, please add them before merging.

PR #2845, opened 2021-06-21
MOBILE-3320 files: Fix clearTmpFiles deleting files that shouldn't be

dpalou:


src/core/features/fileuploader/services/fileuploader.ts
```diff
@@ -191,14 +191,14 @@ export class CoreFileUploaderProvider {
...
     /**
      * Clear temporary attachments to be uploaded.
-     * Attachments already saved in an offline store will NOT be deleted.
+     * Attachments already saved in an offline store will NOT be deleted, only files in tmp folder will be deleted.
      *
      * @param files List of files.
      */
-    clearTmpFiles(files: (CoreWSFile | FileEntry)[]): void {
-        // Delete the local files.
+    async clearTmpFiles(files: (CoreWSFile | FileEntry)[]): Promise<void> {
```
NoelDeMartin, line 198: Why did you convert this to async?

src/core/features/fileuploader/services/fileuploader.ts
```diff
@@ -191,14 +191,14 @@ export class CoreFileUploaderProvider {
...
+        // Delete the temporary files.
         files.forEach((file) => {
-            if ('remove' in file) {
+            if ('remove' in file && CoreFile.removeBasePath(file.toURL()).indexOf(CoreFileProvider.TMPFOLDER) === 0) {
```
NoelDeMartin, line 201: You could use `String.startsWith()` method instead.

NoelDeMartin (approved):

src/core/features/fileuploader/services/fileuploader.ts
dpalou: First I did some changes in the function that required it to be async, but they weren't needed in the end and I forgot to remove it. I just removed it.

src/core/features/fileuploader/services/fileuploader.ts
dpalou: Done!

PR #2848, opened 2021-06-22
MOBILE-3320 calendar: Fix timezone in ion-datetime

NoelDeMartin:
This is a temporary fix, it should be considered further in MOBILE-3784


dpalou:
Hi Noel, we also have ion-datetime in the calendar event page (for reminders) and in the datetime user profile field. Is this change not needed there?

NoelDeMartin:
@dpalou Yes I noticed there were also in other places, but those dates are not initialized using those functions (they are initialized by ionic). We do use the functions to parse them, but those work as expected because if they end with Z, the offset will be 0 already. And if they don't, they should be parsed properly by momentjs (I've recently been testing reminders during QA and it worked as expected).

dpalou (approved):

PR #2850, opened 2021-06-23
MOBILE-3320: Fix regressions

NoelDeMartin:
The last commit in particular fixes database templates, that were broken in [this commit](https://github.com/moodlehq/moodleapp/commit/49f6cc56e55a503c696102946f73a4b3b96be0f5#diff-23f88613aff553b7ff16c3c2f1541eca6be80af122be539ce36247a0b6fec225R661), and I think it was a mistake done by using global replacements.

The fix may undo the a11y improvements that were introduced in that commit, but the proper solution would involve making a regex that captures the whole link (including contents and closing tag) and replacing it for a button. I think that's too risky at this stage, so it's better to just revert it to what we had before and revisit this issue later on, when we review a11y for the database activity.

### DIFF
--- a/src/addons/mod/forum/pages/discussion/discussion.html
+++ b/src/addons/mod/forum/pages/discussion/discussion.html
@@ -23,15 +23,15 @@
             (action)="doRefresh(null, $event, true)">
         </core-context-menu-item>
         <core-context-menu-item [hidden]="sort == 'flat-oldest'"
-            [priority]="500" [content]="'addon.mod_forum.modeflatoldestfirst' | translate" iconAction="arrow-round-down"
+            [priority]="500" [content]="'addon.mod_forum.modeflatoldestfirst' | translate" iconAction="fa-arrow-down"
             (action)="changeSort('flat-oldest')">
         </core-context-menu-item>
         <core-context-menu-item [hidden]="sort == 'flat-newest'"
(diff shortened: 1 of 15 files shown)


src/addons/mod/forum/pages/discussion/discussion.html
```diff
@@ -23,15 +23,15 @@ <h1 *ngIf="startingPost">
             (action)="doRefresh(null, $event, true)">
         </core-context-menu-item>
         <core-context-menu-item [hidden]="sort == 'flat-oldest'"
-            [priority]="500" [content]="'addon.mod_forum.modeflatoldestfirst' | translate" iconAction="arrow-round-down"
+            [priority]="500" [content]="'addon.mod_forum.modeflatoldestfirst' | translate" iconAction="fa-arrow-down"
```
dpalou, line 26: We should use "fas" or "far" icons instead of "fa". I think that right now we default to "fas" if you use "fa".

dpalou (changes requested):

src/addons/mod/forum/pages/discussion/discussion.html
NoelDeMartin: Thanks for the review, I added a new commit updating all the `fa-` to `fas-` across the app to keep it consistent.

dpalou (approved):

PR #2857, opened 2021-06-30
MOBILE-3320: UI tweaks

NoelDeMartin:


src/core/directives/format-text.ts
```diff
@@ -236,6 +237,7 @@ export class CoreFormatTextDirective implements OnChanges {
             });
 
             img.parentNode?.appendChild(button);
+            img.onload = () => button.classList.remove('hidden');
```
dpalou, line 240: I'm worried that in some cases this onload isn't triggered for images that load fine (e.g. if the image is already loaded). Have you checked if this can happen?

In this same file we check when external images are loaded, and in that case we use the external content directive to check it (see externalImage.onLoad). But in this case we don't have the directive so we cannot use the same approach.

dpalou (changes requested):

src/core/directives/format-text.ts
NoelDeMartin: Thanks for the review, I looked into this and I think `onload` should be called always, even if the image is cached or it comes from a base64 string, etc. What I did notice looking at the code is that this could be called with the image already in the document. With the current implementation I think that's not the case, but that's something outside of this method's responsibility so I updated it to show the button if the image is already loaded.

dpalou (approved):

PR #2907, opened 2021-08-03
MOBILE-3799 h5p: Let teachers view users attempts in activity

dpalou:


src/addons/mod/h5pactivity/h5pactivity-lazy.module.ts
```diff
@@ -36,6 +36,11 @@ const routes: Routes = [
         loadChildren: () => import('./pages/attempt-results/attempt-results.module')
             .then( m => m.AddonModH5PActivityAttemptResultsPageModule),
     },
+    {
```
NoelDeMartin, line 39: This is route is declared lazy, but given that this page is only used in this route we could declare the component directly and that way we don't have to declare a page module. We don't have any guidelines when to declare routes lazy or not, so it's ok to leave it like this (I see the other pages in this file have the same situation). But maybe we should reconsider this and be less aggressive with lazy loading, for having less code and better readability. In situations like this one, the performance improvement may not be worth the verboseness.

src/addons/mod/h5pactivity/pages/users-attempts/users-attempts.html
```diff
@@ -0,0 +1,70 @@
...
+    </ion-toolbar>
+</ion-header>
+<ion-content>
+    <ion-refresher slot="fixed" [disabled]="!loaded" (ionRefresh)="doRefresh($event.target)">
+        <ion-refresher-content pullingText="{{ 'core.pulltorefresh' | translate }}"></ion-refresher-content>
+    </ion-refresher>
+    <core-loading [hideUntil]="loaded">
+        <ion-list *ngIf="users.length">
+            <!-- "Header" of the table -->
+            <ion-item class="addon-mod_h5pactivity-table-header hide-detail" detail="true">
```
NoelDeMartin, line 20: Why are we using `detail="true"` with `hide-detail`? Wouldn't it be better just to remove both?

src/addons/mod/h5pactivity/pages/users-attempts/users-attempts.scss
```diff
@@ -0,0 +1,5 @@
+:host {
+    .addon-mod_h5pactivity-table-header {
+        font-weight: bold;
```
NoelDeMartin, line 3: Given that this file is only necessary to change the font weight, I think it'd be better to use a utility class instead. If it were missing we could create it, but turns out there is already a `core-bold` class for this purpose. (I guess at some point we should standardize these utility classes because they are a bit over the place).

NoelDeMartin (changes requested):
Overall I think it's ok, I just found a couple of suggestions.

Something I noticed is that we're duplicating a lot of similar code, and we could extract it into components/classes instead. For example, all the PTR, infinite loading, `canLoadMore`, prefetch, etc. appears in a lot of places with the same structure. At some point it would be easier to just implement something similar to the `CorePageItemsListManager` and have some reusable components. Other than making our lives as developers easier, it'd also be easier to be consistent in the UI.

I'm not saying to do that now, but it's food for thought.

src/addons/mod/h5pactivity/h5pactivity-lazy.module.ts
dpalou: We use this lazy loading approach in most places of the app, I'd suggest leaving it like this for now and we can re-evaluate it in the performance issues. Since this page is only used for teachers maybe it makes sense to keep it in a different module, I don't know how much it improves performance.

src/addons/mod/h5pactivity/pages/users-attempts/users-attempts.html
dpalou: It's to make the item "save" the space used by the detail icon, otherwise the header and the content columns aren't aligned because the content rows have the icon but the header doesn't.

src/addons/mod/h5pactivity/pages/users-attempts/users-attempts.scss
dpalou: Right, I had more CSS in that file but then I removed it because I changed some things and it ended up like this. I removed the file and I'm using the font-bold class now.

src/addons/mod/h5pactivity/h5pactivity-lazy.module.ts
NoelDeMartin: Ok, I think the performance is negligible at this level. But if we already have it like this in most places it's ok to leave it like that, it's just more verbose.

NoelDeMartin (approved):

PR #2923, opened 2021-08-11
MOBILE-3833: Testing improvements, DX improvements & clean up

NoelDeMartin:


src/core/initializers/inject-ios-scripts.ts
```diff
@@ -0,0 +1,26 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { CoreApp } from '@services/app';
+import { CoreIframeUtils } from '@services/utils/iframe';
+import { Platform } from '@singletons';
+
+export default async function(): Promise<void> {
+    if (!CoreApp.isIOS() || 'WKUserScript' in window) {
```
dpalou, line 20: This if is wrong, the second condition should be negated: if window does NOT have WKUserScript, stop.

dpalou (changes requested):

src/core/initializers/inject-ios-scripts.ts
NoelDeMartin: Nice catch! I messed up inverting the conditional, I just fixed it.

dpalou (approved):

PR #2932, opened 2021-09-01
Mobile 3793

dpalou:


src/core/services/utils/dom.ts
```diff
@@ -1728,13 +1728,13 @@ export class CoreDomUtilsProvider {
     /**
      * Opens a popover.
      *
-     * @param popoverOptions Modal Options.
+     * @param options Options.
      */
     async openPopover<T = void>(
```
NoelDeMartin, line 1733: This was possible before, but I noticed that this could be collapsed on a single line (it probably ended up like this after refactoring).

src/core/services/utils/dom.ts
```diff
@@ -1728,13 +1728,13 @@ export class CoreDomUtilsProvider {
...
-        popoverOptions: PopoverOptions,
+        options: OpenPopoverOptions,
     ): Promise<T | undefined> {
 
-        const popover = await PopoverController.create(popoverOptions);
+        const popover = await PopoverController.create(options);
```
NoelDeMartin, line 1737: I don't think this will be a problem, but we're sending extra options into this function (for example, `waitForDismiss`). Maybe a safer way to do it would be:

```js
const { waitForDismiss, ...popoverOptions } = options;
const popover = await PopoverController.create(popoverOptions);
```

That way we can also use `waitForDismiss` directly.

src/core/services/utils/dom.ts
```diff
@@ -1743,16 +1743,16 @@ export class CoreDomUtilsProvider {
...
+                    fixIOSPopoverPosition(popover, options.event);
                     break;
                 case 'md':
-                    fixMDPopoverPosition(popover, popoverOptions.event);
+                    fixMDPopoverPosition(popover, options.event);
                     break;
             }
         }
 
         // If onDidDismiss is nedded we can add a new param to the function to wait one function or the other.
```
NoelDeMartin, line 1754: This comment should be removed, given that it's talking about the option that has been added in this PR so it's no longer relevant.

src/core/services/utils/dom.ts
```diff
@@ -2046,3 +2046,10 @@ export const CoreDomUtils = makeSingleton(CoreDomUtilsProvider);
 
 type AnchorOrMediaElement =
     HTMLAnchorElement | HTMLImageElement | HTMLAudioElement | HTMLVideoElement | HTMLSourceElement | HTMLTrackElement;
+
+/**
+ * Options for the openPopover function.
+ */
+export type OpenPopoverOptions = PopoverOptions & {
+    waitForDismiss?: boolean;
```
NoelDeMartin, line 2054: This name is misleading, because `waitForDismiss: false` would suggest that opening the popover will not wait for dismiss, but it will. The difference between true/false is that the dismiss animation will be completed or not, but this method will always return after the user dimisses the popover.

Some names that come to mind: `waitForDismissCompleted`, `waitForDismissAnimation`, etc.

NoelDeMartin (commented):
Overall it looks good, I just found a couple of suggestions that are not even related with the posts functionality.

src/core/services/utils/dom.ts
dpalou: I changed to a single line :)

src/core/services/utils/dom.ts
dpalou: You're right, I just changed it to the way you suggested.

src/core/services/utils/dom.ts
dpalou: Done.

src/core/services/utils/dom.ts
dpalou: I like waitForDismissCompleted, I changed it to use that name.

NoelDeMartin (approved):

PR #2933, opened 2021-09-01
Mobile 3846

dpalou:


src/core/classes/errors/errorwithtitle.ts
```diff
@@ -0,0 +1,31 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { CoreError } from './error';
+
+/**
+ * Error with title different than just "Error".
+ * If title is not displayed or empty, no title will be displayed.
+ */
+export class CoreErrorWithTitle extends CoreError {
```
NoelDeMartin, line 21: Looking at this class, it seems redundant. What's the difference between `title` and `message`? I think what this is meant to do is that `message` is internal and `title` is something that can be shown to users, but `title` is not set anywhere in the app (that's probably a problem introduced during a refactor).

Looking at how this is actually used, I think a better approach would be to call it something like `CoreLocalizedError`, and take a translation key as an argument. The constructor would use that key to get a localized string, and use that as the `message` calling the super constructor. Like this, we'll know that the message for any instance of this error class can be displayed to users, and there's not need to have two different messages (one for internal and one for users).

src/core/services/update-manager.ts
```diff
@@ -48,6 +63,10 @@ export class CoreUpdateManagerProvider {
```
NoelDeMartin, line 63: I think we can use this commit to refactor this method a bit. Instead of calling it `load`, we could call it `initialize` and call it from within `src/core/initializers/initialize-services.ts`, that way we can remove the `load-update-manager` initializer.

src/core/services/update-manager.ts
```diff
@@ -31,9 +35,20 @@ const VERSION_APPLIED = 'version_applied';
...
         this.logger = CoreLogger.getInstance('CoreUpdateManagerProvider');
+        this.doneDeferred = CoreUtils.promiseDefer();
+    }
+
+    /**
+     * Returns a promise resolved when the load function is done.
+     *
+     * @return Promise resolved when the load function is done.
+     */
+    done(): Promise<void> {
```
NoelDeMartin, line 50: In order to make this more consistent with other services, we could call this `donePromise` and make it a getter instead of a method (I think we are only doing this in `ApplicationInitStatus` for now though, but it's good to set a precedent).

src/core/services/utils/dom.ts
```diff
@@ -1376,6 +1377,8 @@ export class CoreDomUtilsProvider {
 
         if (this.isNetworkError(message, error)) {
             alertOptions.cssClass = 'core-alert-network-error';
+        } else if (error instanceof CoreErrorWithTitle) {
+            alertOptions.header = error.title || undefined;
```
NoelDeMartin, line 1381: Is this correct? or should we use `Translate.instant('core.error')` if the title is missing? (or message is missing, if you like the refactor I mentioned about the error class).

src/core/services/sites.ts
```diff
@@ -823,6 +825,11 @@ export class CoreSitesProvider {
 
         const site = await this.getSite(siteId);
 
+        const siteUrlAllowed = await CoreLoginHelper.isSiteUrlAllowed(site.getURL(), false);
+        if (!siteUrlAllowed) {
+            throw new CoreErrorWithTitle(Translate.instant('core.login.sitenotallowed'));
```
NoelDeMartin, line 830: Looking at the error message and where this situation can happen, I'm not sure if it will be understandable for the user. It just says "This site is no longer available", but it may not be clear for the user which site that means. I think this code can be reached when they click a site on the sites page, and it would be clear then. But it could also happen if they click on a deep link.

Also, if the site is no longer allowed, why would we keep it in the storage? Maybe before throwing the error we should delete the site to avoid repeating the problem.

In any case, I'm not entirely sure when this will be called, so we can leave it as is. I just wanted to point it out in case it can be a problem.

NoelDeMartin (commented):
Overall I think it works, I just have a couple of suggestions.

src/core/services/update-manager.ts
dpalou: Right, I changed it.

src/core/services/update-manager.ts
dpalou: Done.

src/core/services/utils/dom.ts
dpalou: In my use case I want to hide the title, so I did it like this (no title specified = no title displayed). I could define the title property as `string | false`, but in Ionic's alert options setting it to undefined hides it, so IMO it makes sense to do the same.

src/core/services/sites.ts
dpalou: I agree it can be a bit confusing for deep links, though I guess it won't be a common use case (BMAs that change the URL isn't a common use case to begin with ^^'). We can talk this with Juan & Edu to try to find a better message if you prefer.

We considered with Juan the option to delete the site, but he preferred not to do it.

src/core/classes/errors/errorwithtitle.ts
NoelDeMartin: Ok, as we discussed internally I didn't understand how it worked at first. So it was only an issue with the nomenclature, updating the documentation block should suffice.

NoelDeMartin (approved):

PR #2937, opened 2021-09-07
Mobile 3849

dpalou:


NoelDeMartin (approved):
I think the changes look good. As we have been discussing internally though, the current approach is a bit clunky and maybe it could be improved. I opened a new issue to look into that: [MOBILE-3856](https://tracker.moodle.org/browse/MOBILE-3856).

PR #2938, opened 2021-09-07
Mobile 3855

dpalou:


src/core/features/course/services/course-helper.ts
```diff
@@ -1033,7 +1033,7 @@ export class CoreCourseHelperProvider {
         }
 
         // There's no prefetch handler for the module, just download the files.
-        files = files || module.contents;
+        files = files || module.contents!;
```
NoelDeMartin, line 1036: Why are we sure that contents are loaded here? What if prefetchHandler was missing?

NoelDeMartin (commented):
I've looked through all the changes and I think most of them are fine, I just added a comment to a method I have some doubts about.

Other than that, I think we're abusing non null assertions. It was fine during the upgrade from Ionic 3 to Ionic 5, but maybe now it's an opportunity to start cleaning up the code. I would even push for using the [no-non-null-assertion](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-non-null-assertion.md) eslint rule. It will probably give a lot of errors now, but maybe we could start using it as a warning.

For this particular issue, one way to improve on this would be to have a real class for modules instead of working with the objects returned by webservices. That way, we could define methods like the following which would be useful to avoid null assertions (and more):

```
class CoreCourseWSModule {

    requireContents(): CoreCourseModuleContentFile[] {
        return this.contents ?? fail;
    }
    
    async getContents(): Promise<CoreCourseModuleContentFile[]> {
        await CoreCourseHelper.downloadModule(this, this.course);
        
        return this.requireContents();
    }

}

function fail(message: string): any {
    throw new Error(message);
}
```

src/core/features/course/services/course-helper.ts
dpalou: You're right, I set it to an empty array by default.

dpalou:
Thanks Noel,

as we discussed internally, I've created a new function to get contents (and fail if not present). Can you please review it again?

NoelDeMartin (approved):

PR #2959, opened 2021-09-28
Mobile 2748

crazyserver:


src/core/features/course/classes/module-base-handler.ts
```diff
@@ -34,24 +34,25 @@ export class CoreModuleHandlerBase implements Partial<CoreCourseModuleHandler> {
...
+    ): Promise<CoreCourseModuleHandlerData> {
         return {
-            icon: CoreCourse.getModuleIconSrc(module.modname, 'modicon' in module ? module.modicon : undefined),
+            icon: await CoreCourse.getModuleIconSrc(module.modname, 'modicon' in module ? module.modicon : undefined),
             title: module.name,
             class: 'addon-mod_' + module.modname + '-handler',
             showDownloadButton: true,
-            action: (event: Event, module: CoreCourseModule, courseId: number, options?: CoreNavigationOptions): void => {
+            mainPageName: this.pageName,
+            action(event: Event, module: CoreCourseModule, courseId: number, options?: CoreNavigationOptions): void {
```
dpalou, line 49: Why did you change this back? When I reviewed the issue about the super class we agreed it was better to use arrow function in "action", that way you can use this.pageName in it and mainPageName isn't needed.

src/addons/mod/resource/services/handlers/module.ts
```diff
@@ -164,28 +161,22 @@ export class AddonModResourceModuleHandlerService extends CoreModuleHandlerBase
...
             const mimetype = module.contentsinfo.mimetypes[0];
             if (mimetype) {
-                resourceData.icon = CoreMimetypeUtils.getMimetypeIcon(mimetype);
+                mimetypeIcon = CoreMimetypeUtils.getMimetypeIcon(mimetype);
             }
-            resourceData.extra = CoreTextUtils.cleanTags(module.afterlink);
+            extra.push(CoreTextUtils.cleanTags(module.afterlink));
 
-        } else if (files && files.length) {
+        } else if (module.contents[0]) {
```
dpalou, line 175: I think it's better to make sure that module.contents exists, like it was done before, since it can be undefined. They should have been loaded in hideOpenButton, but that could change in the future  so IMO it's better to do: module.contents && module.contents[0]

src/addons/mod/url/services/handlers/module.ts
```diff
@@ -118,10 +118,11 @@ export class AddonModUrlModuleHandlerService extends CoreModuleHandlerBase imple
 
             handlerData.buttons[0].hidden = hideButton;
 
-            if (module.contents && module.contents[0]) {
-                // Calculate the icon to use.
-                handlerData.icon = await CoreCourse.getModuleIconSrc(module.modname, module.modicon) ||
-                    AddonModUrl.guessIcon(module.contents[0].fileurl);
+            if (module.contents[0]) {
```
dpalou, line 121: Like above, I think it's better to keep the previous check to prevent future errors.

src/core/features/course/services/course.ts
```diff
@@ -1484,12 +1489,12 @@ export type CoreCourseWSModule = {
     noviewlink?: boolean; // Whether the module has no view page.
     completion?: number; // Type of completion tracking: 0 means none, 1 manual, 2 automatic.
     completiondata?: CoreCourseModuleWSCompletionData; // Module completion data.
-    contents?: CoreCourseModuleContentFile[];
+    contents: CoreCourseModuleContentFile[];
```
dpalou, line 1492: This change is wrong. contents can be undefined, e.g. if the module has an access restriction.

src/core/features/course/services/course.ts
```diff
@@ -402,29 +402,30 @@ export class CoreCourseProvider {
         ): Promise<CoreCourseWSSection[]> => {
             const params: CoreCourseGetContentsParams = {
                 courseid: courseId!,
-                options: [],
             };
+            params.options = [];
```
dpalou, line 406: Is this change needed? Is TS unable to detect that options exist if they're in the initial object?

src/addons/mod/resource/services/handlers/module.ts
```diff
@@ -94,7 +93,7 @@ export class AddonModResourceModuleHandlerService extends CoreModuleHandlerBase
         }];
 
         this.getResourceData(module, courseId, handlerData).then((data) => {
-            handlerData.icon = data.icon;
+            handlerData.icon = data.icon; // @TODO
```
dpalou, line 96: What is this TODO about?

Also, I think the function getResourceData doesn't return an icon right now, and I don't see it changed in your patch.

src/addons/mod/url/services/handlers/module.ts
```diff
@@ -111,17 +111,18 @@ export class AddonModUrlModuleHandlerService extends CoreModuleHandlerBase imple
...
             handlerData.buttons[0].hidden = hideButton;
 
-            if (module.contents && module.contents[0]) {
-                // Calculate the icon to use.
-                handlerData.icon = AddonModUrl.guessIcon(module.contents[0].fileurl) ||
-                    CoreCourse.getModuleIconSrc(module.modname, 'modicon' in module ? module.modicon : undefined);
+            if (module.contents[0]) {
+                const icon = AddonModUrl.guessIcon(module.contents[0].fileurl);
+
+                // Calculate the icon to use. @TODO
```
dpalou, line 124: Should this TODO be here? Aren't you calculating the icon already?

src/core/features/course/services/course.ts
```diff
@@ -632,7 +631,11 @@ export class CoreCourseProvider {
      * @param modicon The mod icon string to use in case we are not using a core activity.
      * @return The IMG src.
      */
-    getModuleIconSrc(moduleName: string, modicon?: string): string {
+    async getModuleIconSrc(moduleName: string, modicon?: string, mimetypeIcon = ''): Promise<string> {
+        if (mimetypeIcon) {
+            return mimetypeIcon;
+        }
+
```
dpalou, line 638: I see that the modicon is only returned if the module isn't supported in the app. Does this mean that for core icons we'll always use the local icons? I thought we wanted to use the remote icons.

dpalou (changes requested):

dpalou:
One thing I forgot to mention is that in some places you're passing componentId, so the icons will be linked to that instance. This means that removing the stored files for that instance will delete the icon. This is right for unique icons (e.g. LTI with a custom icon), but generic icons shared by all instances shouldn't be deleted if an instance data is deleted.

src/core/features/course/classes/module-base-handler.ts
crazyserver: I think it's a rebase error, reverted now.

src/addons/mod/resource/services/handlers/module.ts
crazyserver: Checking CoreCourseWSModule contents is always set. That's why I changed it.

crazyserver: Ok, reverting it

src/core/features/course/services/course.ts
crazyserver: Ah ok, I see the problem now.

src/core/features/course/services/course.ts
crazyserver: It says Object possibly undefined.

src/addons/mod/url/services/handlers/module.ts
crazyserver: no, deleted.

src/addons/mod/resource/services/handlers/module.ts
crazyserver: TODO deleted, getResourceData is returning icon, probably you misread some code.

crazyserver:
Thanks Dani! Ready to review again!

dpalou (approved):

PR #2964, opened 2021-09-30
MOBILE-3833 forum: Fix discussion link navigation

crazyserver:


src/addons/mod/forum/services/handlers/discussion-link.ts
```diff
@@ -55,10 +55,11 @@ export class AddonModForumDiscussionLinkHandlerService extends CoreContentLinksH
         return [{
             action: (siteId): void => {
                 const discussionId = parseInt(params.d, 10);
+                const cmId = data?.cmid && Number(data.cmid);
+                courseId = courseId || (params.courseid && Number(params.courseid)) || (params.cid && Number(params.cid));
+
                 const pageParams: Params = {
-                    forumId: data?.instance && parseInt(data.instance, 10),
-                    cmId: data?.cmid && parseInt(data.cmid, 10),
```
dpalou, line 60: I think you should keep the cmId and courseId params when using the discussion path, that way if one of them is set (but not both) the page will get the param that is set.

dpalou (changes requested):

src/addons/mod/forum/services/handlers/discussion-link.ts
crazyserver: Thanks Dani ! I thought the same but then I just thought it was difficult to have it one or the other. Fixed

dpalou (approved):

PR #2978, opened 2021-10-19
Mobile 3887

dpalou:


src/core/classes/site.ts
```diff
@@ -86,6 +86,7 @@ export class CoreSite {
         '3.9': 2020061500,
         '3.10': 2020110900,
         '3.11': 2021051700,
+        '4.0': 2021100300, // @todo replace with right value when released. Using a tmp value to be able to test new things.
```
NoelDeMartin, line 89: I'm not sure if there are any more TODOs for 4.0 release, but I think we should follow a format that is easily searchable when the time comes. For example `@todo [4.0]`, or if a TODO depends on an issue being fixed `@todo [MDL-123456]`. I mention this because it's already happened to me that a release is made, or an issue is fixed, and I didn't remove this type of code because I didn't remember I had to search for it. This wouldn't happen if we review all TODOs before every release, but in practice we're not doing that. If TODOs are written like this, we at least can review all the TODOs that have something to do with the release. With issues it's more tricky, because we won't search for TODOs every time an issue from core is fixed, but since that's not the case here we can ignore it for now.

src/addons/notifications/services/notifications.ts
```diff
@@ -272,28 +225,55 @@ export class AddonNotificationsProvider {
      * @param siteId Site ID. If not defined, use current site.
      * @return Promise resolved with the message notifications count.
      */
-    async getUnreadNotificationsCount(userId?: number, siteId?: string): Promise<number> {
+    async getUnreadNotificationsCount(userId?: number, siteId?: string): Promise<number | string> {
```
NoelDeMartin, line 228: Looking at how this method is used, I think it'd be better to return an object instead. This value is basically encoding two types of information, and that's why we have to resort to this `number | string` shenanigans which is cumbersome to use later on (you have to parse a string to get the int value, etc.).

What this method is trying to convey is the count of unread notifications and whether there are more or not. For example, if a user has 1000 unread notifications, this will return '20+', and that is not the "notifications count", it's just an approximation. Instead, we could return an object like `{ count: 20, hasMore: true }`. That way, we won't have to parse any strings and use the boolean in a readable way. Also, the fact that we want to display more than 20 notifications as "20+" (instead of, for example, "you have more than 20 unread notifications") is a concern of the UI layer of the application, given that this is the logic layer it shouldn't have any opinions as to how to display it.

src/addons/notifications/services/notifications.ts
```diff
@@ -272,28 +225,55 @@ export class AddonNotificationsProvider {
...
+            }
+        }
 
+        // Fallback call
         try {
-            return await site.read<number>('message_popup_get_unread_popup_notification_count', params, preSets);
+            const unread = await this.getUnreadNotifications({ limit: AddonNotificationsProvider.LIST_LIMIT, siteId });
+
+            if (unread.length === AddonNotificationsProvider.LIST_LIMIT) {
+                // Maybe there are more notifications, include a '+';
```
NoelDeMartin, line 256: Instead of "maybe" (this will also show the + if we only have 20 unread notifications), why not ask for the limit+1 and that wat we'll be sure if there are more or not?

src/addons/notifications/services/notifications.ts
```diff
@@ -464,6 +444,7 @@ export type AddonNotificationsGetMessagesMessage = {
     component?: string; // @since 3.7. The component that generated the notification.
     eventtype?: string; // @since 3.7. The type of notification.
     customdata?: string; // @since 3.7. Custom data to be passed to the message processor.
+    iconurl?: string; // @since 4.0. Icon URL, only for notifications.
```
NoelDeMartin, line 447: Can we use this issue to rename `AddonNotificationsGetMessagesMessage` to `AddonNotificationsMessage` or `AddonNotificationsNotificationMessage`? We end up using this type in different places, and the fact that it was returned form a "getter" is irrelevant to the rest of the app.

The same idea for `AddonNotificationsGetMessagesMessageFormatted` and other related types (if any).

src/addons/notifications/services/notifications-helper.ts
```diff
@@ -59,53 +60,63 @@ export class AddonNotificationsHelperProvider {
...
      * @param options Other options.
      * @return Promise resolved with notifications and if can load more.
      */
     async getNotifications(
-        notifications: AddonNotificationsAnyNotification[],
-        options?: AddonNotificationsHelperGetNotificationsOptions,
-    ): Promise<{notifications: AddonNotificationsAnyNotification[]; canLoadMore: boolean}> {
+        notifications: AddonNotificationsGetMessagesMessageFormatted[],
+        options?: AddonNotificationsGetNotificationsOptions,
+    ): Promise<{notifications: AddonNotificationsGetMessagesMessageFormatted[]; canLoadMore: boolean}> {
```
NoelDeMartin, line 72: I think it is confusing that we have two functions: `AddonNotifications.getNotifications` and `AddonNotificationsHelper.getNotifications`.

Looking at the code, I noticed that the one that is actually used is the one in the helper, and the one from the notifications service is only used internally by the former. In order to simplify the public API and the code overall, what do you think of having only one `getNotifications` method? That way, everything can be in a single file (I don't care if the helper or the notifications service, since I think both fullfill the same role). It should be a matter of just refactoring code, the functionality can be the same and I think it'll be even simpler because methods such as `AddonNotifications.getReadNotifications` can be removed.

If some of these methods are still kept for backwards compatibility, I'd be explicit about deprecating them and just calling whatever they must underneath.

src/addons/notifications/services/notifications-helper.ts
```diff
@@ -59,53 +60,63 @@ export class AddonNotificationsHelperProvider {
...
+                        throw error;
+                    }
                 }
             }
         }
 
         return {
             notifications: newNotifications,
-            canLoadMore: notifications.length >= options.limit,
+            canLoadMore: newNotifications.length >= options.limit,
```
NoelDeMartin, line 119: Is this correct? if the limit was 5 and we get 5 new notifications, that doesn't mean that there is more, maybe there were only 5 left. So I guess it should be `>` instead of `>=`, but I understand it will never be a greater number because we just used the limit. So, could we request limit + 1 instead and remove the last item if necessary? I know it sounds like wasting network resources, but consider that saying "canLoadMore" when it isn't true can potentially trigger a whole new request for nothing. Maybe there's a better way though?

src/addons/notifications/services/notifications.ts
```diff
@@ -157,28 +155,30 @@ export class AddonNotificationsProvider {
...
-        options?: AddonNotificationsGetNotificationsOptions,
+        read: AddonNotificationsGetReadType,
+        options: AddonNotificationsGetNotificationsOptions = {},
     ): Promise<AddonNotificationsGetMessagesMessageFormatted[]> {
-        options = options || {};
+        options.offset = options.offset || 0;
         options.limit = options.limit || AddonNotificationsProvider.LIST_LIMIT;
 
-        this.logger.debug(`Get ${(read ? 'read' : 'unread')} notifications from ${offset}. Limit: ${options.limit}`);
+        const typeText = read === AddonNotificationsGetReadType.READ ?
```
NoelDeMartin, line 168: Enum types in Typescript can also be defined as strings, so you can define them directly as strings and avoid this casting.

src/addons/notifications/services/notifications.ts
```diff
@@ -157,28 +155,30 @@ export class AddonNotificationsProvider {
...
+        this.logger.debug(`Get ${typeText} notifications from ${options.offset}. Limit: ${options.limit}`);
 
         const site = await CoreSites.getSite(options.siteId);
+
         const data: AddonNotificationsGetMessagesWSParams = {
             useridto: site.getUserId(),
             useridfrom: 0,
             type: 'notifications',
-            read: !!read,
+            read: read,
```
NoelDeMartin, line 179: `read` expects a `boolean | number` here, but you are passing a `AddonNotificationsGetReadType` here. Typescript doesn't complain because at this moment `AddonNotificationsGetReadType` is always a number, but that doesn't mean that this number has the same meaning as the number this parameter expects (and will break if we ever change the type of the enum values, for example if you accept my previous suggestion).

NoelDeMartin (changes requested):
I didn't find any bug or issue introduced in this PR, but I spotted a bunch of potential improvements.

src/addons/notifications/services/notifications.ts
dpalou: Right, I changed it to return an object.

src/addons/notifications/services/notifications-helper.ts
dpalou: Right, I moved the function from the helper to the notifications service. The one in notifications is now protected and has a different name.

src/addons/notifications/services/notifications.ts
dpalou: As we discussed in chat, this enum matches Moodle constants. I changed the WSParams type to specify it accepts an enum.

NoelDeMartin (approved):

PR #2982, opened 2021-10-28
MOBILE-3899 mainmenu: Hide main menu on navigation level > 1

crazyserver:


src/core/singletons/colors.ts
```diff
@@ -64,7 +64,12 @@ export class CoreColors {
         document.body.appendChild(d);
 
         // Color in RGB .
-        const rgba = getComputedStyle(d).color.match(/\d+/g)!.map((a) => parseInt(a, 10));
+        const matches = getComputedStyle(d).color.match(/\d+/g) || [];
+        if (!matches) {
```
dpalou, line 68: This condition will never happen because you're initializing matches to an empty array if the regex isn't found.

dpalou (changes requested):

src/core/singletons/colors.ts
crazyserver: Oh yeah... sorry, amended now. Thanks!

dpalou (approved):

PR #2983, opened 2021-10-29
Mobile 3807

crazyserver:


src/core/features/login/components/sites/sites.html
```diff
@@ -0,0 +1,104 @@
...
+                            alt="{{ 'core.pictureof' | translate:{$a: accountsList.currentSite.fullName} }}"
+                            onError="this.src='assets/img/user-avatar.png'">
+                    </ion-avatar>
+                    <ion-label>
+                        <p class="item-heading">{{accountsList.currentSite.fullName}}</p>
+                    </ion-label>
+                    <ion-icon color="success" name="fas-check"></ion-icon>
+                </ion-item>
+
+                <ion-item button (click)="login($event, site.id)" *ngFor="let site of accountsList.sameSite" detail="true">
```
dpalou, line 47: The HTML for sites in sameSite and otherSites is the same, you can use an ng-template to avoid duplicating all the HTML.

src/core/features/login/services/login-helper.ts
```diff
@@ -407,8 +409,21 @@ export class CoreLoginHelperProvider {
...
+        let params: Params = { openAddSite: true , showKeyboard };
+
+        if (CoreSites.isLoggedIn()) {
+
+            if (CoreSitePlugins.hasSitePluginsLoaded) {
+                // The site has site plugins so the app will be restarted. Store the data and logout.
+                CoreApp.storeRedirect(CoreConstants.NO_SITE_ID, path, { params });
+            }
+
+            await CoreSites.logout();
```
dpalou, line 423: I think it would be better to stop the execution when there are site plugins instead of navigating to login/sites after logout. Otherwise it can happen that the navigation is performed before the app is refreshed, and navigating to login can consume the redirect that you have just stored. Usually when we do logout and there are site plugins we don't do anything else, we just wait for the app to be refreshed.

src/core/features/login/services/login-helper.ts
```diff
@@ -407,8 +409,21 @@ export class CoreLoginHelperProvider {
      * @param showKeyboard Whether to show keyboard in the new page. Only if no fixed URL set.
      * @return Promise resolved when done.
      */
-    async goToAddSite(setRoot?: boolean, showKeyboard?: boolean): Promise<void> {
-        const [path, params] = this.getAddSiteRouteInfo(showKeyboard);
+    async goToAddSite(setRoot = false, showKeyboard = false): Promise<void> {
+        let path = '/login/sites';
+        let params: Params = { openAddSite: true , showKeyboard };
```
dpalou, line 414: The "openAddSite" can probably be implemented using the "nextNavigation" introduced recently in the navigator. But since it's something that already exists you can leave it like it is if you prefer.

src/core/features/login/components/sites/sites.html
```diff
@@ -0,0 +1,104 @@
...
+                        <ion-icon name="fas-trash" slot="icon-only" aria-hidden="true"></ion-icon>
+                    </ion-button>
+                </ion-item>
+            </ng-container>
+
+            <ng-container *ngFor="let sites of accountsList.otherSites">
+                <ion-item-divider sticky="true">
+                    <ion-label>
+                        <h2>
+                            <core-format-text [text]="sites[0].siteName" clean="true" [siteId]="sites[0].id"></core-format-text>
```
dpalou, line 71: You're assuming that each list of sites will have a first element, and looking at the implementation of getAccountsList and deleteAccountFromList it will. But this could change in the future, IMO it would be better to verify that sites[0] exists so it doesn't break so easily.

The same happens in sites.html.

dpalou (changes requested):

dpalou:
Along with the comments above, I think the commit name "Add some delegates to user menu" is wrong, you're adding handlers, not delegates.

src/theme/globals.variables.scss
```diff
@@ -127,7 +127,15 @@ $core-login-loading-color-dark: $text-color-dark !default;
 $core-login-hide-forgot-password: false !default;
 $core-login-hide-need-help: false !default;
 
-// Configuration options for more page.
+// Configuration options for more page. (deprecated)
```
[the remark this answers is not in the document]
crazyserver: ok!! thx!

src/addons/privatefiles/services/handlers/user.ts
```diff
@@ -14,44 +14,57 @@
...
 
     /**
-     * Returns the data needed to render the handler.
-     *
-     * @return Data needed to render the handler.
+     * @inheritdoc
      */
-    getDisplayData(): CoreMainMenuHandlerData {
+    async isEnabledForUser(user: CoreUserProfile): Promise<boolean> {
+        // All course grades only available for the current user.
```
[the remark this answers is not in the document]
crazyserver: Nice catch, thanks!

src/core/features/login/components/sites/sites.html
crazyserver: Done!

src/core/features/login/services/login-helper.ts
crazyserver: I will leave it as it is. Thanks!!

src/core/features/mainmenu/components/user-menu/user-menu.html
```diff
@@ -0,0 +1,63 @@
...
+            </ion-item>
+
+            <ion-item button *ngFor="let handler of handlers" class="ion-text-wrap" (click)="handlerClicked($event, handler)"
+                [ngClass]="['core-user-menu-handler', handler.class || '']" [hidden]="handler.hidden"
+                [attr.aria-label]="handler.title | translate" detail="true">
+                <ion-icon *ngIf="handler.icon" [name]="handler.icon" slot="start" aria-hidden="true"></ion-icon>
+                <ion-label>
+                    <p class="item-heading">{{ handler.title | translate }}</p>
+                </ion-label>
+            </ion-item>
```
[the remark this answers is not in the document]
crazyserver: Solved in lastest commit

dpalou (approved):

PR #2987, opened 2021-11-04
MOBILE-3896 test: Allow using existing instance to mock class

dpalou:


NoelDeMartin (approved):
I'm merging this now, but make sure to update the related documentation. I think the [Unit testing for the Moodle App](https://docs.moodle.org/dev/Unit_testing_for_the_Moodle_App) page is the only one affected.

PR #2991, opened 2021-11-09
MOBILE-3905: Swipe navigation

NoelDeMartin:


src/core/classes/items-management/list-items-manager.ts
```diff
@@ -0,0 +1,218 @@
...
+     */
+    async select(item: Item): Promise<void> {
+        await this.navigateToItem(item, { reset: this.resetNavigation() });
+    }
+
+    /**
+     * Log activity when the page starts.
+     */
+    protected async logActivity(): Promise<void> {
+        //
```
dpalou, line 140: It looks weird to have an empty comment in here.

src/core/classes/items-management/items-manager.ts
```diff
@@ -0,0 +1,213 @@
...
+ * Helper to manage a collection of items in a page.
+ */
+export abstract class CoreItemsManager<
+    _Source extends CoreItemsManagerSource = CoreItemsManagerSource,
+    Item extends CoreItemsManagerSourceItem<_Source> = CoreItemsManagerSourceItem<_Source>,
+    Source extends CoreItemsManagerSource<Item> = CoreItemsManagerSource<Item>
+> {
+
+    protected source?: { instance: Source; unsubscribe: () => void };
+    protected itemsList: Item[] | null = null;
```
dpalou, line 36: Do we need to replicate the list of items? It's already inside source, I think it's better to always use the same list.

src/core/classes/page-items-list-manager.ts
```diff
@@ -22,6 +22,8 @@ import { CoreUtils } from '@services/utils/utils';
 
 /**
  * Helper class to manage the state and routing of a list of items in a page, for example on pages using a split view.
+ *
+ * @deprecated use CoreListItemsManager instead.
```
dpalou, line 26: Can you specify the app version? @deprecated since app 4.0

src/core/classes/items-management/items-manager.ts
```diff
@@ -0,0 +1,213 @@
...
+     * Set source.
+     *
+     * @param newSource New source.
+     */
+    protected setSource(newSource: Source | null): void {
+        if (this.source) {
+            this.source.unsubscribe();
+
+            if (CoreItemsManagerSourcesTracker.isTracked(this.source.instance)) {
+                CoreItemsManagerSourcesTracker.removeReference(this.source.instance);
```
dpalou, line 117: As you told me in chat, it can cause problems that this function decreases the reference counter but it doesn't increase it when setting a new source.

src/core/classes/items-management/items-manager-source.ts
```diff
@@ -0,0 +1,159 @@
...
+
+    /**
+     * Load items for the next page, if any.
+     */
+    async loadNextPage(): Promise<void> {
+        if (!this.hasMoreItems) {
+            return;
+        }
+
+        const { items, hasMoreItems } = await this.loadPageItems(this.getPagesLoaded() + 1);
```
dpalou, line 126: Maybe we should remove this "+ 1". I think in general we count pages starting from 0, so the first page is the number 0, not the number 1. Adding 1 will probably make most sources have to decrease it by 1 like you do in the participants one.

src/core/classes/items-management/items-manager-source.ts
```diff
@@ -0,0 +1,159 @@
...
+
+        this.setItems((this.items ?? [])?.concat(items), hasMoreItems);
```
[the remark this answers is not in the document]
NoelDeMartin: Done

src/core/classes/items-management/items-manager-source.ts
```diff
@@ -0,0 +1,159 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Updates listener.
+ */
+export interface CoreItemsListSourceListener<Item> {
+    onItemsUpdated(items: Item[]): void;
```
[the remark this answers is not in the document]
NoelDeMartin: Done

src/core/classes/items-management/items-manager-sources-tracker.ts
```diff
@@ -0,0 +1,117 @@
...
+     *
+     * @param constructor Source constructor.
+     * @param constructorArguments Arguments to create a new instance, used to find out if an instance already exists.
+     * @returns Source.
+     */
+    static getOrCreateSource<T extends CoreItemsManagerSource>(
+        constructor: SourceConstructor<T>,
+        constructorArguments: ConstructorParameters<SourceConstructor<T>>,
+    ): T  {
+        const id = constructorArguments.filter(argument => !!argument).map(argument => String(argument)).join('-');
```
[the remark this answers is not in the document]
NoelDeMartin: The problem I was facing was that the second argument of the participants source had a default value, and the ids would be different depending on that value being passed or not (for example, you could have `42-null` and `42-undefined` as different ids).

So, looking further into this my solution wasn't correct either, because what would happen if we have a default argument with a non-empty value?

I refactored this taking a different approach, let me know what you think about it.

src/core/classes/items-management/items-manager-sources-tracker.ts
```diff
@@ -0,0 +1,117 @@
...
+/**
+ * Tracks CoreItemsManagerSource instances to reuse between pages.
+ */
+export class CoreItemsManagerSourcesTracker {
+
+    private static instances: WeakMap<SourceConstructor, Instances> = new WeakMap();
+    private static instanceIds: WeakMap<CoreItemsManagerSource, string> = new WeakMap();
+
+    /**
+     * Get an instance of the given source or retrieve one if it's already in use.
```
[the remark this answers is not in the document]
NoelDeMartin: Done

src/core/classes/items-management/list-items-manager.ts
NoelDeMartin: We already had this in the previous manager, but I just added a comment to clarify that this is intended to be overridden if necessary. We do this instead of declaring it as abstract because it's optional, some pages may not want to log the activity.

src/core/classes/items-management/items-manager.ts
NoelDeMartin: Done

src/core/classes/page-items-list-manager.ts
NoelDeMartin: I think it's ok to leave it like this because this is just a marker whilst we continue implementing swipe gestures across the app. This API is internal, and hopefully this entire file will be deleted before the 4.0 release.

src/core/classes/items-management/items-manager.ts
NoelDeMartin: I've refactored the way this works, let me know what you think.

src/core/classes/items-management/items-manager-source.ts
NoelDeMartin: I had the impression that pages always start from 1, but it seems like we are already using the concept of a page starting from 0 so I just applied your suggestion.

src/core/classes/items-management/items-manager-sources-tracker.ts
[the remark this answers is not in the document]
dpalou: Thanks Noel, I like the new approach :)

src/core/classes/items-management/items-manager.ts
dpalou: It looks good to me!

dpalou (approved):

PR #2994, opened 2021-11-15
Mobile 3914

crazyserver:


src/core/features/course/components/format/format.ts
```diff
@@ -191,6 +190,7 @@ export class CoreCourseFormatComponent implements OnInit, OnChanges, OnDestroy {
 
             this.displaySectionSelector = CoreCourseFormatDelegate.displaySectionSelector(this.course);
             this.displayBlocks = CoreCourseFormatDelegate.displayBlocks(this.course);
+            this.hasBlocks = (await CoreBlockHelper.getCourseBlocks(this.course.id)).length > 0;
```
dpalou, line 193: The getCourseBlocks call can fail, causing an uncaught exception, and then the following code won't be executed and no error message will be displayed. IMO we should either:

1- Display an error message if the WS call fails and don't display the button.
2- Don't display error but don't display the button.

src/core/features/course/components/format/format.ts
```diff
@@ -498,8 +498,13 @@ export class CoreCourseFormatComponent implements OnInit, OnChanges, OnDestroy {
             await component.callComponentFunction('doRefresh', [refresher, done, afterCompletionChange]);
         }) || [];
 
-        if (this.courseBlocksComponent) {
-            promises.push(this.courseBlocksComponent.doRefresh());
+        if (this.course) {
+            const courseId = this.course.id;
+            promises.push(CoreCourse.invalidateCourseBlocks(courseId).then(async () => {
+                this.hasBlocks = (await CoreBlockHelper.getCourseBlocks(courseId)).length > 0;
```
dpalou, line 504: As above, this WS call can fail.

src/core/features/sitehome/pages/index/index.ts
```diff
@@ -140,17 +140,17 @@ export class CoreSiteHomeIndexPage implements OnInit, OnDestroy {
...
                 this.currentSite.getInfo()?.sitename,
-            ).catch(() => {
-                // Ignore errors.
-            });
+            ));
         } catch (error) {
             CoreDomUtils.showErrorModalDefault(error, 'core.course.couldnotloadsectioncontent', true);
         }
+
+        this.hasBlocks = (await CoreBlockHelper.getCourseBlocks(this.siteHomeId)).length > 0;
```
dpalou, line 153: Since this code is outside of the try/catch it can cause an uncaught exception too if it fails.

src/core/features/courses/services/dashboard.ts
```diff
@@ -63,6 +63,41 @@ export class CoreCoursesDashboardProvider {
...
+        blocks.forEach((block) => {
+            if (block.region == 'content' || block.region == 'main') {
+                mainBlocks.push(block);
+            } else {
+                sideBlocks.push(block);
+            }
+        });
+
+        if (mainBlocks.length == 0) {
+            blocks.forEach((block) => {
```
dpalou, line 88: Shouldn't you empty the sideBlocks array before this forEach? Otherwise you can have duplicated blocks in sideBlocks.

dpalou (changes requested):

src/core/features/course/components/format/format.ts
crazyserver: I chose the second!

src/core/features/courses/services/dashboard.ts
crazyserver: Oh !thats correct, I thought about that when implementing and then forgot about that.

dpalou (approved):

dpalou:
Thanks Pau, I just noticed something else. If you click a block with navigation (e.g. Private Files) the navigation is performed behind the modal.

crazyserver:
Solved!

dpalou (approved):

PR #3004, opened 2021-11-24
Mobile 3810

crazyserver:


crazyserver:
Checks passed, I'm removing Workaround commit (will fail checks)

src/core/services/utils/dom.ts
```diff
@@ -1690,26 +1691,36 @@ export class CoreDomUtilsProvider {
     async openModal<T = unknown>(
         options: OpenModalOptions,
     ): Promise<T | undefined> {
-
         const { waitForDismissCompleted, closeOnNavigate, ...modalOptions } = options;
         const listenCloseEvents = closeOnNavigate ?? true; // Default to true.
 
-        const modal = await ModalController.create(modalOptions);
+        const modalId = <string> Md5.hashAsciiStr(options.component?.toString() || '');
```
NoelDeMartin, line 1697: Is the component name enough to identify a unique modal? Alert ids are created using title and message, which is pretty much the entire content of an alert, but modals can be more complex. What about the same modal with different properties? I'm not sure if that is currently happening in the app, but it can potentially happen.

src/addons/mod/resource/services/handlers/module.ts
```diff
@@ -226,6 +226,32 @@ export class AddonModResourceModuleHandlerService extends CoreModuleHandlerBase
         };
     }
 
+    /**
+     * @inheritdoc
+     */
+    async getIconSrc(module?: CoreCourseWSModule): Promise<string | undefined> {
```
NoelDeMartin, line 232: Looking at the code, I didn't find any instance where this method is called passing a `module` argument, it's always undefined. Is this argument just for plugins to use?

src/core/features/course/components/module-info/module-info.ts
```diff
@@ -12,21 +12,55 @@
...
 
     @Input() module!: CoreCourseModule; // The module to render.
-    @Input() showManualCompletion = false; // Whether to show manual completion.
+    @Input() courseId!: number; // The courseId the module belongs to.
+
+    @Input() component!: string; // Component for format text directive.
+    @Input() componentId!: string | number; // Component ID to use in conjunction with the component.
+
+    @Input() showDescription = true; // Whether to show description.
+    @Input() description?: string; // The description to display.
```
NoelDeMartin, line 45: Instead of having these two properties, could we just have description defined as `string | false`? If we want to show the description, we pass the string, if we don't we just use `false`.

For example, for forums we have the following:

```html
[showDescription]="forum && forum.type != 'single'" [description]="description"
```

and it could become:

```html
[description]="forum && forum.type != 'single' && description"
```

src/addons/mod/wiki/pages/index/index.ts
```diff
@@ -47,22 +47,15 @@ export class AddonModWikiIndexPage extends CoreCourseModuleMainActivityPage<Addo
...
+     * @inheritdoc
      */
     updateData(data: { name: string } | string): void {
         if (typeof data == 'string') {
             // We received the title to display.
-            this.title = data;
-        } else {
-            // We received a wiki instance.
-            this.title = this.pageTitle || data.name || this.title;
+            this.pageTitle = data;
```
NoelDeMartin, line 58: Given that we're not setting title anymore, shouldn't this call `super.updateData(data)`? It was actually sort of wrong before, because we omitted calling super because of implementation details (we knew it was just setting `title` and it wasn't necessary).

src/core/directives/format-text.ts
```diff
@@ -73,6 +71,8 @@ export class CoreFormatTextDirective implements OnChanges {
     @Input() openLinksInApp?: boolean; // Whether links should be opened in InAppBrowser.
     @Input() hideIfEmpty = false; // If true, the tag will contain nothing if text is empty.
 
+    @Input() fullOnClick?: boolean | string; // @deprecated on 4.0. Won't do nothing.
+    @Input() fullTitle?: string; // @deprecated on 4.0. Won't do nothing..
```
NoelDeMartin, line 75: `Won't do nothing` should say `Won't do anything` (and avoid double `.`). I've seen this string elsewhere in the app.

src/core/directives/format-text.ts
```diff
@@ -267,37 +269,61 @@ export class CoreFormatTextDirective implements OnChanges {
...
     /**
-     * Display the "Show more" in the element.
+     * Toggles if expand button is enabled or not.
+     *
+     * @param enable Wether enable or disable.
      */
-    protected displayShowMore(): void {
-        const expandInFullview = CoreUtils.isTrueOrOne(this.fullOnClick) || false;
-        const showMoreButton = document.createElement('ion-button');
+    protected toggleExpandButtonEnabled(enable: boolean): void {
```
NoelDeMartin, line 280: Should this be called `setExpandButtonEnabled` instead? It's common to call toggle methods that change the current state, but if the state argument is not optional, you're just setting or updating the state, not toggling.

src/core/directives/format-text.ts
```diff
@@ -321,38 +347,16 @@ export class CoreFormatTextDirective implements OnChanges {
             return;
         }
 
-        const expandInFullview = CoreUtils.isTrueOrOne(this.fullOnClick) || false;
-
-        if (!expandInFullview && !this.showMoreDisplayed) {
+        if (!this.toggleExpandEnabled /* || this.expanded*/) {
```
NoelDeMartin, line 350: Why is this code commented? Can we remove it instead?

Also, this method is called `elementClicked` but that isn't too descriptive, specially since the clicked element is not the root of the directive, but the button to toggle expand status.

src/core/directives/format-text.ts
```diff
@@ -321,38 +347,16 @@ export class CoreFormatTextDirective implements OnChanges {
...
-                    component: this.component,
-                    componentId: this.componentId,
-                    filter: filter,
-                    contextLevel: this.contextLevel,
-                    instanceId: this.contextInstanceId,
-                    courseId: this.courseId,
-                },
-            );
-        }
+        // Change class.
```
NoelDeMartin, line 358: The code below does more than just changing a class, so this comment is misleading. I think `toggleExpand` is already descriptive enough, so we could just remove the comment altogether.

src/core/features/course/components/module-info/core-course-module-info.html
```diff
@@ -1,18 +1,45 @@
...
+                [contextInstanceId]="module.id" [courseId]="courseId">
+            </core-format-text>
+        </h1>
+        <ng-content select="[title]"></ng-content>
+    </ion-label>
+</ion-item>
+<ion-item class="ion-text-wrap" *ngIf="showDescription && description" lines="none">
+    <ion-label>
+        <core-format-text [text]="description" [component]="component" [componentId]="componentId" contextLevel="module"
+            [contextInstanceId]="module.id" [courseId]="courseId" maxHeight="120">
```
NoelDeMartin, line 16: This should be `[maxHeight]=120` instead (using brackets). I've seen this issue elsewhere in the app.

src/core/directives/collapsible-header.ts
```diff
@@ -0,0 +1,258 @@
...
+    /**
+     * Gets the loading content id to wait for the loading to finish.
+     *
+     * @TODO: If no core-loading is present, load directly.
+     *
+     * @return Promise resolved with Loading Id, if any.
+     */
+    protected async getLoadingId(): Promise<string | undefined> {
+        if (!this.content) {
+            this.content = this.header.parentElement?.querySelector('ion-content:not(.disable-scroll-y)');
```
NoelDeMartin, line 71: `this.content` is initialized as a side-effect of calling `getLoadingId`, but this attribute is used elsewhere in this class. It works properly now because `getLoadingId()` is called before any of the other methods, but this could change if we refactor the code at some point (specially since this method already has a TODO mentioning that it will change).

I assume we cannot initialize it in the constructor, but I suppose doing it inside `ngOnInit` or something should work?

src/core/directives/collapsible-header.ts
```diff
@@ -0,0 +1,258 @@
...
+        this.loadingObserver = CoreEvents.on(CoreEvents.CORE_LOADING_CHANGED, async (data) => {
+            const loadingId = await this.getLoadingId();
+            if (loadingId && data.loaded && data.uniqueId == loadingId) {
+                // Remove event when loading is done.
+                this.loadingObserver.off();
+
+                // Wait to render.
+                setTimeout(() => {
+                    this.setupRealTitle();
+                }, 100);
```
NoelDeMartin, line 57: Instead of setting an arbitrary timeout of 100ms, if we're only waiting for render using `await CoreUtils.nextTick()` would be more appropriate.

src/core/directives/collapsible-header.ts
```diff
@@ -0,0 +1,258 @@
...
+     * @param contentH1 Heading 1 of title, if found.
+     * @param scrollDetail Event details.
+     */
+    protected onScroll(
+        title: HTMLElement,
+        contentH1: HTMLElement,
+        contentSubheading: HTMLElement | null,
+        scrollDetail: ScrollDetail,
+    ): void {
+        let progress = scrollDetail.scrollTop / this.titleTopDifference;
```
NoelDeMartin, line 211: You can declare this as `const progress = CoreMath.clamp(scrollDetail.scrollTop / this.titleTopDifference, 0, 1)` instead and simplify the rest of the code.

src/core/classes/page-transition.ts
```diff
@@ -0,0 +1,364 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { createAnimation } from '@ionic/angular';
+import { Animation, NavOptions } from '@ionic/core';
+
+/**
+ * Adaptation from Ionic 5 iOs transition.
```
NoelDeMartin, line 19: I assume this file is mostly copy-pasted from Ionic source code, right? Keeping that in mind, I didn't review this code, but I think that it would be helpful to include a link here to the original file (a link to github in the specific release version you copied it from should suffice).

src/core/services/utils/dom.ts
crazyserver: Usually only one modal should be open at the same time and I assume they will be different, this could be improved if something like the situation you mentioned happen but it's very unlikely to happen right now and we will detect it rapidly.

crazyserver: Also component is not the name, it's the whole class.

src/core/features/course/components/module-info/module-info.ts
crazyserver: Good catch, done!

src/core/directives/format-text.ts
crazyserver: Fixed

src/core/directives/format-text.ts
crazyserver: Done

src/addons/mod/resource/services/handlers/module.ts
crazyserver: This is called on the module-delegate.ts func getModuleIconSrc and it passes the module.

src/addons/mod/wiki/pages/index/index.ts
crazyserver: Oh yeah, calling the super in the else and also , recovered the dataReceived previously removed call.

src/core/directives/format-text.ts
crazyserver: Fixed

src/core/features/course/components/module-info/core-course-module-info.html
crazyserver: true, fixed

src/core/classes/tabs.ts
```diff
@@ -43,8 +43,8 @@ export class CoreTabsBaseComponent<T extends CoreTabBase> implements OnInit, Aft
 
     // Minimum tab's width.
     protected static readonly MIN_TAB_WIDTH = 107;
-    // Max height that allows tab hiding.
-    protected static readonly MAX_HEIGHT_TO_HIDE_TABS = 768;
+    // Max height that allows tab hiding. WARNING: Hide tabs on scroll disabled. If confirmed, remove the associated code.
```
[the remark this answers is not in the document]
crazyserver: done

src/core/directives/collapsible-header.ts
crazyserver: Fixed

src/core/directives/collapsible-header.ts
crazyserver: Fixed

src/core/classes/page-transition.ts
crazyserver: fixed

src/core/directives/collapsible-header.ts
crazyserver: I added a comment since I don't see any better solution now. In many cases content is not added to the HTML on ngInit and we should wait a bit.

src/core/directives/format-text.ts
crazyserver: Oh yes, removed. The click is received by the element, if you click on any part of the element that is not a link or button or anything listening to the click, the text expand will be toggled.

NoelDeMartin: Ok, I didn't look at the code properly. Then it's fine.

src/addons/mod/resource/services/handlers/module.ts
NoelDeMartin: I see, I didn't find it because it's calling it without proper type references. We already have many of these in handlers, so I guess it's ok for now. But this is something we should improve at some point.

PR #3012, opened 2021-12-01
MOBILE-3926: Add swipe navigation to pages with split-view

NoelDeMartin:


src/addons/badges/pages/issued-badge/issued-badge.page.ts
```diff
@@ -103,10 +110,46 @@ export class AddonBadgesIssuedBadgePage implements OnInit {
         ]));
 
         await CoreUtils.ignoreErrors(Promise.all([
+            this.badges?.getSource().reload(),
```
dpalou, line 113: I think this reload isn't needed.

dpalou (changes requested):

src/addons/badges/pages/issued-badge/issued-badge.page.ts
NoelDeMartin: Thanks, I've fixed it.

dpalou (approved):

PR #3019, opened 2021-12-09
MOBILE-3939: Add swipe navigation to feedback attempts

NoelDeMartin:


src/addons/mod/feedback/pages/attempts/attempts.ts
```diff
@@ -75,79 +104,47 @@ export class AddonModFeedbackAttemptsPage implements AfterViewInit {
...
+            this.fetchFailed = true;
 
-            if (!refresh) {
-                // Some call failed on first fetch, go back.
-                CoreNavigator.back();
-            }
+            CoreDomUtils.showErrorModalDefault(error, 'core.course.errorgetmodule', true);
         }
+
+        await attempts.start(this.splitView);
```
dpalou, line 120: Should you call `start` when the fetch fails? One of the things that `start` does is call `logActivity`, and we usually don't call the log WS if the fetch has failed.

dpalou (changes requested):

src/addons/mod/feedback/pages/attempts/attempts.ts
NoelDeMartin: As we discussed internally, this is the approach we are using in most parts of the code already. I prefer to keep it consistent and we can tackle this issue everywhere later on.

dpalou (approved):

PR #3021, opened 2021-12-13
Mobile 3927

dpalou:


src/core/classes/items-management/slides-dynamic-items-manager-source.ts
```diff
@@ -0,0 +1,217 @@
...
+        return [];
+    }
+
+    /**
+     * Load a certain item and preload next and previous ones.
+     *
+     * @param item Basic data about the item to load.
+     * @return Promise resolved when done.
+     */
+    async loadItem(item: Partial<Item>): Promise<void> {
```
NoelDeMartin, line 55: Looking at the dynamic items manager, I think the concept of dynamic items and "unloaded items" has been conflated. What I mean with this is that the dynamic nature of this list makes reference to not all items being available in a list, given that it's theoretically an infinite list. However, the fact that items need to be loaded by themselves is not related with this. For example, if we had a calendar without any events or any data tied to each month, items would just be year/month, and we wouldn't need to load them individually. I think this is also part of the reason for the partial items I mentioned earlier. Is it too late to decouple these two concepts? We could also make reusable code for items that need to be loaded individually, but that code doesn't need to be in this class. Items can also have their own classes and inheritance, they don't always need to be plain JS objects. So this method could be replaced whenever it's called in the code with `item.load()` instead.

src/addons/calendar/components/calendar/calendar.ts
```diff
@@ -144,170 +152,248 @@ export class AddonCalendarCalendarComponent implements OnInit, DoCheck, OnDestro
...
+    weekDays: AddonCalendarWeekDaysTranslationKeys[];
+    weeks: AddonCalendarWeek[];
+};
+
+/**
+ * Helper to manage swiping within months.
+ */
+class AddonCalendarMonthSlidesItemsManagerSource extends CoreSwipeSlidesDynamicItemsManagerSource<PreloadedMonth> {
+
+    monthLoaded = false;
```
NoelDeMartin, line 354: Would it be possible to remove this and calculate it from the selected item instead?

src/addons/calendar/components/calendar/calendar.ts
```diff
@@ -318,151 +404,177 @@ export class AddonCalendarCalendarComponent implements OnInit, DoCheck, OnDestro
...
+                        result = await AddonCalendarHelper.getOfflineMonthWeeks(month.year, month.monthNumber);
+                    } else {
+                        throw error;
+                    }
+                }
+            }
+
+            const weekDays = AddonCalendar.getWeekDays(result.daynames[0].dayno);
+            const weeks = result.weeks as AddonCalendarWeek[];
+            const isCurrentMonth = CoreTimeUtils.isSameMonth(month, CoreTimeUtils.getCurrentMonth());
```
NoelDeMartin, line 515: Instead of recalculating this every time it's necessary, would it be easier to encode this information into each item? For example, we could have `month.isCurrent`, `month.isPrevious`, etc. as booleans.

src/addons/calendar/pages/day/day.page.ts
```diff
@@ -562,156 +396,447 @@ export class AddonCalendarDayPage implements OnInit, OnDestroy {
...
+    // Offline events classified in month & day.
+    offlineEvents: Record<string, Record<number, AddonCalendarEventToDisplay[]>> = {};
+    offlineEditedEventsIds: number[] = []; // IDs of events edited in offline.
+    categories: { [id: number]: CoreCategoryData } = {};
+    deletedEvents: number[] = []; // Events deleted in offline.
+    timeFormat?: string;
+    canCreate = false;
+
+    protected dayPage: AddonCalendarDayPage;
+    protected categoriesRetrieved = false;
```
NoelDeMartin, line 515: Instead of tracking this separately, could we just define `categories` as null and check whether categories have been retrieved or not from the status of that property?

src/addons/calendar/components/calendar/calendar.ts
```diff
@@ -61,23 +70,12 @@ export class AddonCalendarCalendarComponent implements OnInit, DoCheck, OnDestro
     @Output() onDayClicked = new EventEmitter<{day: number; month: number; year: number}>();
 
     periodName?: string;
-    weekDays: AddonCalendarWeekDaysTranslationKeys[] = [];
-    weeks: AddonCalendarWeek[] = [];
+    manager?: CoreSwipeSlidesDynamicItemsManager<PreloadedMonth, AddonCalendarMonthSlidesItemsManagerSource>;
     loaded = false;
```
NoelDeMartin, line 74: I think this property could be removed and inferred from the status of the manager.

src/addons/calendar/components/calendar/calendar.ts
```diff
@@ -92,9 +90,15 @@ export class AddonCalendarCalendarComponent implements OnInit, DoCheck, OnDestro
...
-                    week.days.forEach((day) => {
-                        AddonCalendar.scheduleEventsNotifications(day.eventsFormated || []);
+                this.manager?.getSource().getItems()?.forEach((month) => {
+                    if (!month.loaded) {
+                        return;
+                    }
+
+                    month.weeks.forEach((week) => {
+                        week.days.forEach((day) => {
+                            AddonCalendar.scheduleEventsNotifications(day.eventsFormated || []);
```
NoelDeMartin, line 100: I think this isn't related to this commit, but I want to make sure of something. How come this code is registered here? It seems like scheduling of events is not something that should be tied to a page being focused or not. If the default notification changes from another screen or any other service, we would want this to be called as well, right? So why not set this listener up in a service instead and decouple it from this page being present?

src/addons/calendar/pages/day/day.page.ts
```diff
@@ -562,156 +396,447 @@ export class AddonCalendarDayPage implements OnInit, OnDestroy {
...
+                // Re-calculate "hasOffline".
+                day.hasOffline = day.events.length != day.onlineEvents.length ||
+                    day.events.some((event) => event.deleted || event.offline);
+            }
+
+            return true;
+        });
+
+        // Add it or remove it from the list of deleted events.
+        if (deleted) {
```
NoelDeMartin, line 830: Could `deletedItems` be a Set instead of an array? It seems like that would simplify this code and possibly other parts.

src/core/classes/items-management/slides-dynamic-items-manager-source.ts
```diff
@@ -0,0 +1,217 @@
...
+        }
+
+        this.setInitialized();
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected async loadItems(): Promise<Item[]> {
+        // Not used in dynamic slides.
```
NoelDeMartin, line 45: Instead of doing this, we could make the `load` method abstract instead. In this class, it would be implemented as is (notice that it isn't calling super), and in other managers where `loadItems` is actually necessary we could implement `load` and keep this one as abstract.

src/core/classes/items-management/slides-dynamic-items-manager-source.ts
```diff
@@ -0,0 +1,217 @@
...
+     * @return Promise resolved when done.
+     */
+    async loadItem(item: Partial<Item>): Promise<void> {
+        const previousItem = this.getPreviousItem(item);
+        const nextItem = this.getNextItem(item);
+
+        await Promise.all([
+            this.loadItemInList(item, false),
+            previousItem ? this.loadItemInList(previousItem, true) : undefined,
+            nextItem ? this.loadItemInList(nextItem, true) : undefined,
```
NoelDeMartin, line 62: These can also be written as `item && this.loadItemInList(item, true)` to make them more terse.

src/core/classes/items-management/slides-items-manager-source.ts
```diff
@@ -0,0 +1,121 @@
...
+    protected abstract loadItems(): Promise<Item[]>;
+
+    /**
+     * Get a certain item.
+     *
+     * @param item Partial data about the item to search.
+     * @return Item, null if not found.
+     */
+    getItem(item: Partial<Item>): Item | null {
+        const index = this.getItemPosition(item);
```
NoelDeMartin, line 82: In relation with some of my previous comments, this method is confusing. Should it be called `getItemById` and expect an id instead?

Just looking at the documentation and method signature, it gives the false impression that you can pass any partial fields of the item. Whereas in reality, it only works with the fields necessary to get the id.

src/core/classes/items-management/slides-items-manager-source.ts
```diff
@@ -0,0 +1,121 @@
...
+        return this.getItemPosition(this.initialItem);
+    }
+
+    /**
+     * Get the ID of an item.
+     *
+     * @param item Data about the item.
+     * @return Item ID.
+     */
+    abstract getItemId(item: Partial<Item>): string | number;
```
NoelDeMartin, line 119: This declaration is misleading, it declares that any partial item is accepted, but then implementations override the declaration with incompatible signatures (it may be related with some Typescript strict settings we are not using). For example, `{}` is a partial month, but the item in the calendar manager is declared as `getItemId(item: YearAndMonth)` which isn't valid.

src/core/components/swipe-slides/swipe-slides.html
```diff
@@ -0,0 +1,6 @@
+<ion-slides *ngIf="manager && manager.getSource().isLoaded()" (ionSlideWillChange)="slideWillChange()"
+    (ionSlideDidChange)="slideDidChange()" [options]="options">
+    <ion-slide *ngFor="let item of manager.getSource().getItems()">
```
NoelDeMartin, line 3: `manager.getSource().getItems()` could be encapsulated in an `items` getter, like it's already happening in the list manager (same for the `manager.getSource().isLoaded()` above).

src/core/services/utils/dom.ts
```diff
@@ -769,21 +769,29 @@ export class CoreDomUtilsProvider {
      *
      * @param scrollEl The element that must be scrolled.
      * @param element DOM element to check.
+     * @param point The point of the element to check.
      * @return Whether the element is outside of the viewport.
      */
-    isElementOutsideOfScreen(scrollEl: HTMLElement, element: HTMLElement): boolean {
+    isElementOutsideOfScreen(scrollEl: HTMLElement, element: HTMLElement, point: 'top' | 'mid' | 'bottom' = 'mid'): boolean {
```
NoelDeMartin, line 775: I think this `point` argument should be declared as an enum instead. That way, the code below can also be implemented using a switch statement without a default case which is more type-safe, performant, and easier to understand.

src/core/classes/items-management/slides-dynamic-items-manager-source.ts
```diff
@@ -0,0 +1,217 @@
...
+    markAllItemsDirty(): void {
+        this.getItems()?.forEach(item => {
+            item.dirty = true;
+        });
+    }
+
+    /**
+     * Mark all items as not loaded.
+     */
+    markAllItemsUnloaded(): void {
```
NoelDeMartin, line 181: Aside from the fact that items don't have to be dirty/loaded as I mentioned in a previous comment, I'm not sure if we need to track `loaded` with a property. Dirtyness is already the concept that something exists, but the data must be updated. Loaded, as I understand it, should be a state derived from the existance of other state, so technically it shouldn't be set unless it's necessary to optimize performance. In this case, I think dirtyness already does the same job, or am I missing something?

src/core/classes/items-management/items-manager-source.ts
```diff
@@ -137,84 +101,21 @@ export abstract class CoreItemsManagerSource<Item = unknown> {
...
-     * @param items Items.
-     * @param hasMoreItems Whether there are more pages to be loaded.
+     * Notify that items have been updated.
      */
-    protected setItems(items: Item[], hasMoreItems: boolean): void {
-        this.items = items;
-        this.hasMoreItems = hasMoreItems;
-
-        this.listeners.forEach(listener => listener.onItemsUpdated?.call(listener, items, hasMoreItems));
+    protected notifyItemsUpdated(): void {
```
NoelDeMartin, line 117: I realize this has been extracted to a method because it's called when items are modified in dynamic managers. However, I don't think this should be the case, the items array should only be modified calling the setter (so in dynamic managers, they would create a new array instead of modifying the existing one).

Looking at dynamic managers, I think one of the motivations to do the current approach is that conceptually the array of items is modified, and not creating a new one. But in reality, this array of items is just a collection of items, because the positions don't have a meaning (as exposed by the fact that these change every time we navigate backwards). So I was thinking, maybe a linked list would be more appropriate for those managers. But then it may not be straight-forward to reconcile with this array, and the concept of position/index that is used in other parts of the parent classes. So I guess just calling `setItems` with a new array should suffice, but the idea of a linked list seems more appropriate.

src/core/classes/items-management/slides-dynamic-items-manager-source.ts
```diff
@@ -0,0 +1,217 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { CoreSwipeSlidesItemsManagerSource } from './slides-items-manager-source';
+
+/**
+ * Items collection source data for "swipe slides".
+ */
+export abstract class CoreSwipeSlidesDynamicItemsManagerSource<Item extends CoreSwipeSlidesDynamicItem>
```
NoelDeMartin, line 20: This file is missing the `swipe-` prefix in the filename, as do all the others starting with `slides-*`.

src/core/classes/items-management/swipe-items-manager.ts
```diff
@@ -16,17 +16,17 @@ import { ActivatedRoute, ActivatedRouteSnapshot, UrlSegment } from '@angular/rou
...
 
-import { CoreItemsManager } from './items-manager';
-import { CoreItemsManagerSource } from './items-manager-source';
+import { CoreRoutedItemsManager } from './routed-items-manager';
+import { CoreRoutedItemsManagerSource } from './routed-items-manager-source';
 
 /**
  * Helper class to manage the state and routing of a swipeable page.
  */
 export class CoreSwipeItemsManager<
```
NoelDeMartin, line 25: This should be renamed to `CoreSwipeNavigationItemsManager`, tying it up with the `CoreSwipeNavigationComponent`. And the file should be called `swipe-navigation-items-manager`.

NoelDeMartin (changes requested):
Well, I wrote a lot of comments but most of them are about code style and architecture, I actually don't think I found any real issue (other than some related with CSS). Let me know what you think about them.

src/core/components/swipe-slides/swipe-slides.scss
```diff
@@ -0,0 +1,15 @@
+:host {
```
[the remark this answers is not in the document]
dpalou: I rebased the code and it no longer happens, it seems it wasn't related with my code.

src/addons/calendar/components/calendar/addon-calendar-calendar.html
```diff
@@ -27,74 +31,81 @@ <h2 id="addon-calendar-monthname">{{ periodName }}</h2>
...
-            <ion-row role="row">
-                <ion-col class="ion-text-center addon-calendar-weekday" *ngFor="let day of weekDays" role="columnheader">
-                    <span class="sr-only">{{ day.fullname | translate }}</span>
-                    <span class="ion-hide-md-up" aria-hidden="true">{{ day.shortname | translate }}</span>
-                    <span class="ion-hide-md-down" aria-hidden="true">{{ day.fullname | translate }}</span>
-                </ion-col>
-            </ion-row>
-        </div>
-        <div role="rowgroup">
+    <core-swipe-slides [manager]="manager" (onWillChange)="slideChanged($event)">
```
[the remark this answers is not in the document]
dpalou: Now the managers notify the change, but I'm using a listener. That way pages only need to register a listener instead of creating a subclass of the manager.

src/core/components/swipe-slides/swipe-slides.ts
```diff
@@ -0,0 +1,263 @@
...
+    }
+
+    /**
+     * Slide to a certain item.
+     *
+     * @param item Item.
+     * @param speed Animation speed.
+     * @param runCallbacks Whether to run callbacks.
+     */
+    slideToItem(item: Partial<Item>, speed?: number, runCallbacks?: boolean): void {
```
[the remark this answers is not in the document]
dpalou: Now all the methods using items use just Item instead of Partial<Item>. In subclasses, the items basic properties are required, and all the loaded properties are optional.

src/addons/mod/book/components/index/index.ts
```diff
@@ -245,3 +237,96 @@ export class AddonModBookIndexComponent extends CoreCourseModuleMainResourceComp
     }
 
 }
+
+type LoadedChapter = {
+    id: number;
+    content: string;
+    tags?: CoreTagItem[];
```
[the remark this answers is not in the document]
dpalou: Actually, to fix the issue with the partial items I made that Items only require the properties to identify them (in this case, id), so now content will be optional too.

src/addons/mod/book/components/index/index.ts
```diff
@@ -245,3 +237,96 @@ export class AddonModBookIndexComponent extends CoreCourseModuleMainResourceComp
...
+                return {
+                    id: chapter.id,
+                    content,
+                    tags: this.TAGS_ENABLED ? this.contentsMap[chapter.id].tags : [],
+                };
+            }));
+
+            return newChapters;
+        } catch (exception) {
+            const error = exception ?? new CoreError(Translate.instant('addon.mod_book.errorchapter'));
```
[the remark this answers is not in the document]
dpalou: I don't know when it can happen, or if it can even happen right now. But usually when this kind of checks are there it's because we found the issue in the past, looking at the code my guess is that it was related to the download file part.

I changed the code, now it'll just check if the error has a message, and put a message if it didn't.

src/core/components/swipe-slides/swipe-slides.ts
```diff
@@ -0,0 +1,263 @@
...
+    templateUrl: 'swipe-slides.html',
+    styleUrls: ['swipe-slides.scss'],
+})
+export class CoreSwipeSlidesComponent<Item = unknown> implements OnChanges, OnDestroy {
+
+    @Input() manager?: CoreSwipeSlidesItemsManager<Item>;
+    @Input() options: CoreSwipeSlidesOptions = {};
+    @Output() onInit = new EventEmitter<CoreSwipeCurrentItemData<Item>>();
+    @Output() onWillChange = new EventEmitter<CoreSwipeCurrentItemData<Item>>();
+    @Output() onDidChange = new EventEmitter<CoreSwipeCurrentItemData<Item>>();
```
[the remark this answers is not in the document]
dpalou: The onInit output wasn't used, it was used at first but then I refactored the code so I removed it.

src/core/components/swipe-slides/swipe-slides.ts
```diff
@@ -0,0 +1,263 @@
...
+    slidePrev(speed?: number, runCallbacks?: boolean): void {
+        this.slides?.slidePrev(speed, runCallbacks);
+    }
+
+    /**
+     * Get current item.
+     *
+     * @return Current item. Undefined if no current item yet.
+     */
+    getCurrentItem(): Item | null {
```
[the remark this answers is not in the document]
dpalou: I changed it to always use "selected" because in some contexts the word current can be confusing. E.g. if I do:

const month = manager.getCurrentItem();

it could be confusing whether the current month is the month I'm viewing or the actual month we're at.

src/core/classes/items-management/slides-items-manager-source.ts
```diff
@@ -0,0 +1,121 @@
...
+
+        return index ?? -1;
+    }
+
+    /**
+     * Get initial item position.
+     *
+     * @return Initial item position.
+     */
+    getInitialPosition(): number {
```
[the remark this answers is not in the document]
dpalou: Using the word index now.

src/core/classes/items-management/slides-items-manager-source.ts
```diff
@@ -0,0 +1,121 @@
...
+
+import { CoreItemsManagerSource } from './items-manager-source';
+
+/**
+ * Items collection source data for "swipe slides".
+ */
+export abstract class CoreSwipeSlidesItemsManagerSource<Item = unknown> extends CoreItemsManagerSource<Item> {
+
+    protected initialItem?: Partial<Item>;
+    protected initialized = false;
```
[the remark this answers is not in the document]
dpalou: I changed it to use the word "loaded", and I moved it to the root class. As we talked in chat, I preferred to use a loaded boolean property instead of refactoring the dynamic slides source to add the items "in block".

src/addons/calendar/components/calendar/calendar.ts
dpalou: I could remove the "monthLoaded" one, but this one has a special case where I need it: when the user goes to the current month/day. In this case the user might be going to an item that isn't in the list yet, so I cannot select/get the item. So I use `this.loaded = false` to display a spinner, load the item and then we'll slide to the loaded item.

src/addons/calendar/components/calendar/calendar.ts
dpalou: I opened MOBILE-3951 to change this.

src/core/classes/items-management/slides-dynamic-items-manager-source.ts
dpalou: If we make the `load` method abstract in CoreSwipeSlidesItemsManagerSource, then all the classes extending this one will have to implement the load method (e.g. the book manager), and we don't want that. So I left this as it was, we can improve it later if needed.

src/core/classes/items-management/slides-dynamic-items-manager-source.ts
dpalou: I cannot derive the `loaded` from other properties because `CoreSwipeSlidesDynamicItem` doesn't know the properties that are loaded later. So I'd have to move the `markAllItemsUnloaded` method to the pages that need it, making it less reusable.

IMO the advantage of using loaded & dirty is that you can behave differently depending on the case. E.g. in the case of the day page, `loaded=false` means I display a spinner, but I don't when `dirty=true` (because I only use `dirty=true` when the user performs a PTR, and we already have the PTR spinner in that case). I could do it checking if one of the optional properties has been loaded, but then in another similar page I'd have to reimplement it with a different property. I think this way is easier to reuse.

src/core/classes/items-management/items-manager-source.ts
dpalou: As we talked in chat, I prefer to keep it like this instead of creating a new array every time I want to add an item to the list.

src/core/components/swipe-slides/swipe-slides.ts
```diff
@@ -0,0 +1,263 @@
...
+     * Initialize some properties based on the manager.
+     */
+    protected async initialize(manager: CoreSwipeSlidesItemsManager<Item>): Promise<void> {
+        this.unsubscribe = manager.getSource().addListener({
+            onItemsUpdated: () => this.onItemsUpdated(),
+        });
+
+        // Don't call default callbacks on init, emit our own events instead.
+        // This is because default callbacks aren't triggered for position 0, and to prevent auto scroll on init.
+        this.options.runCallbacksOnInit = false;
```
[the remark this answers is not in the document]
dpalou: I don't like the idea of having two different option objects, I think it can be confusing. IMO having an overridden option is not a big deal, we can document it as a limitation of our component. And passing extra properties shouldn't be a problem either, it's something that TypeScript accepts.

dpalou:
Thanks Noel, I fixed most of the things you commented. If I didn't comment on any of the things you mentioned it means I just fixed it, no need to put "Done" everywhere :)

src/core/classes/items-management/items-manager.ts
```diff
@@ -86,94 +89,64 @@ export abstract class CoreItemsManager<Item = unknown, Source extends CoreItemsM
...
-        await CoreNavigator.navigate(pathPrefix + itemPath, { params, ...options });
+        this.listeners.splice(index, 1);
     }
 
     /**
      * Called when source items have been updated.
      *
      * @param items New items.
      */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
NoelDeMartin, line 143: This line can be removed, `items` is not unused anymore.

src/addons/calendar/components/calendar/calendar.ts
```diff
@@ -144,171 +152,257 @@ export class AddonCalendarCalendarComponent implements OnInit, DoCheck, OnDestro
...
+                });
             }
         }
     }
 
+    get timeFormat(): string {
+        return this.manager?.getSource().timeFormat || 'core.strftimetime';
+    }
+
+    get isCurrentMonth(): string {
```
NoelDeMartin, line 174: It seems like this has been added by mistake and can be removed.

NoelDeMartin (changes requested):
I found a couple of small things, after fixing those this'll be ready for merging.

NoelDeMartin (approved):

PR #3026, opened 2021-12-15
Mobile 3950

crazyserver:


src/core/features/viewer/components/image/image.ts
```diff
@@ -24,22 +25,79 @@ import { ModalController, Translate } from '@singletons';
     templateUrl: 'image.html',
     styleUrls: ['image.scss'],
 })
-export class CoreViewerImageComponent implements OnInit {
+export class CoreViewerImageComponent implements OnInit, AfterViewInit {
+
+    @ViewChild(IonSlides) protected slides!: IonSlides;
```
dpalou, line 30: I think it's better to declare this as optional, because we don't know how much it'll take for Angular to initialize this variable.

src/core/features/viewer/components/image/image.ts
```diff
@@ -24,22 +25,79 @@ import { ModalController, Translate } from '@singletons';
...
+            maxRatio: 8,
+            minRatio: 0.5,
+        },
+    };
+
+    protected zoomRatio = 1;
+
+    slidesSwiper: any; // eslint-disable-line @typescript-eslint/no-explicit-any
+
+    constructor(protected element: ElementRef) {
```
dpalou, line 51: I think it's better to declare this as ElementRef<HTMLElement>, otherwise nativeElement has type "any". It's something I discovered "recently" and now I always try to specify the type of the element.

src/core/features/viewer/components/image/image.ts
```diff
@@ -24,22 +25,79 @@ import { ModalController, Translate } from '@singletons';
...
+
+        if (!this.slidesSwiper || !imageElement) {
+            return;
+        }
+
+        zoomIn
+            ? this.zoomRatio *= 2
+            : this.zoomRatio /= 2;
+
+        this.zoomRatio = CoreMath.clamp(this.zoomRatio, 1, 8);
```
dpalou, line 91: You might want to use slidesOpts.zoom.maxRatio and minRatio to make sure the values match. Right now they don't, here the min is 1 but in the options is 0.5.

dpalou (changes requested):

src/core/features/viewer/components/image/image.ts
dpalou: It seems Github removed what I put between "<" and ">". I meant: ElementRef  < HTMLElement >

crazyserver:
Thanks Dani! I resolved all you found. 
During the clamp I'm using 1 as minimum instead of minRatio because the min-maxRatio is only used during pinch gesture.

dpalou (approved):

PR #3029, opened 2021-12-20
Mobile 3932

dpalou:


src/addons/mod/resource/components/index/index.scss
```diff
@@ -0,0 +1,11 @@
+@import "~theme/globals";
+
+:host {
+    .addon-mod_resource-outdated {
+        @include padding(5px, 0px, 0px, 0px);
```
crazyserver, line 5: Where this 5px come from? I prefer using an power of 4: 4, 8, or 16px. I think Material design uses that.

crazyserver (changes requested):

crazyserver (approved):

PR #3037, opened 2021-12-22
MOBILE-3895 mathjax: Add overflow to block equations

dpalou:


src/theme/components/format-text.scss
```diff
@@ -211,6 +211,12 @@ core-format-text {
     .core-format-text-content .filter_mathjaxloader_equation .MathJax_Preview a {
         pointer-events: none;
     }
+
+    .MathJax_SVG_Display {
+        overflow: auto;
```
crazyserver, line 216: I think It would be better to use overflow-x to avoid any possible problem with Y axis.

crazyserver (changes requested):

src/theme/components/format-text.scss
dpalou: Right, changed!

crazyserver (approved):

PR #3040, opened 2022-01-08
there is any document or video tutorials for edit this app

heonical:
there is any document or video tutorials for edit this app


dpalou:
Please see https://docs.moodle.org/dev/Moodle_App_Development_Guide.

PR #3042, opened 2022-01-10
Mobile 3954

crazyserver:


src/addons/storagemanager/pages/course-storage/course-storage.ts
```diff
@@ -52,16 +92,156 @@ export class AddonStorageManagerCourseStoragePage implements OnInit {
...
+        this.downloadCourseEnabled = !!this.course && !CoreCourses.isDownloadCourseDisabledInSite();
+        this.downloadEnabled = !CoreSites.getRequiredCurrentSite().isOfflineDisabled();
+
+        const sections = await CoreCourse.getSections(this.courseId, false, true);
+        this.sections = CoreCourseHelper.addHandlerDataForModules(sections, this.courseId).sections
+            .map((section) => ({ ...section, totalSize: 0 }));
+
+        await this.loadSizes();
+        await this.initCoursePrefetch();
+        await this.initModulePrefetch();
```
dpalou, line 112: I think these 3 calls can be done in parallel using Promise.all. The last one doesn't event use await right now, it's like a sync function but you can keep it as async if you prefer.

dpalou (changes requested):
Thanks Pau,

besides the other comments, I think you should also add a new entry in upgrade.txt to specify that the downloadEnabled input has been removed for certain components.

src/addons/storagemanager/pages/course-storage/course-storage.ts
crazyserver: Nice catch. I had to transform CoreCourseHelper.addHandlerDataForModules to async, since it was incorrectly sync and it was causing some trouble showing the buttons on modules.

src/addons/storagemanager/pages/course-storage/course-storage.html
```diff
@@ -4,62 +4,99 @@
...
                 <ion-card-header>
                     <ion-item class="ion-no-padding">
                         <ion-label>
                             <p class="item-heading ion-text-wrap">{{ section.name }}</p>
+                            <ion-badge color="light" *ngIf="section.totalSize > 0">
+                                {{ section.totalSize | coreBytesToSize }}
+                            </ion-badge>
+                            <!-- Download progress. -->
+                            <p>
+                                <ion-progress-bar *ngIf="downloadEnabled && section.isDownloading"
```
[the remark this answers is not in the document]
crazyserver: Done. I added a new inteterminate status to our progress bar when number is negative.

dpalou (approved):

PR #3056, opened 2022-01-18
MOBILE-3962 blocks: Add course list block

crazyserver:


src/addons/block/courselist/courselist.module.ts
```diff
@@ -0,0 +1,36 @@
...
+import { APP_INITIALIZER, NgModule } from '@angular/core';
+import { IonicModule } from '@ionic/angular';
+import { TranslateModule } from '@ngx-translate/core';
+import { CoreBlockDelegate } from '@features/block/services/block-delegate';
+import { AddonBlockCourseListHandler } from './services/block-handler';
+
+@NgModule({
+    imports: [
+        IonicModule,
+        TranslateModule.forChild(),
```
NoelDeMartin, line 24: I realize most blocks are already declared like this, but I wonder if it'd be better to import `CoreSharedModule` instead? It imports more modules than necessary, but those are modules that are already loaded in the app (given that the "shared module" is probably always loaded on startup). So I wonder if that'd cause any issues or we can import the shared module to avoid duplicating these two lines all the time.

I don't know, this is just a comment to keep in mind though, I'll go ahead with the merge.

NoelDeMartin (approved):

src/addons/block/courselist/courselist.module.ts
crazyserver: We can check that on future releases!

PR #3057, opened 2022-01-18
Mobile 3811

crazyserver:


src/core/directives/collapsible-header.ts
```diff
@@ -58,7 +58,7 @@ export class CoreCollapsibleHeaderDirective implements OnDestroy {
                 this.loadingObserver.off();
 
                 // Wait to render.
-                await CoreUtils.nextTick();
+                await CoreUtils.nextTicks(10);
```
NoelDeMartin, line 61: This `10` seems a bit arbitrary, is there a way to wait for something specific to happen instead?

src/core/features/course/services/course.ts
```diff
@@ -125,9 +125,8 @@ export class CoreCourseProvider {
     static readonly COMPONENT = 'CoreCourse';
 
     readonly CORE_MODULES = [
-        'assign', 'assignment', 'book', 'chat', 'choice', 'data', 'database', 'date', 'external-tool',
-        'feedback', 'file', 'folder', 'forum', 'glossary', 'ims', 'imscp', 'label', 'lesson', 'lti', 'page', 'quiz',
-        'resource', 'scorm', 'survey', 'url', 'wiki', 'workshop', 'h5pactivity', 'bigbluebuttonbn',
+        'assign', 'bigbluebuttonbn', 'book', 'chat', 'choice', 'data', 'feedback', 'folder', 'forum', 'glossary', 'h5pactivity',
+        'imscp', 'label', 'lesson', 'lti', 'page', 'quiz', 'resource', 'scorm', 'survey', 'url', 'wiki', 'workshop',
```
NoelDeMartin, line 129: I'm not sure why this changed, but looking at the LMS it seems like some of the values that existed before (like `mod_database`) never existed. How come?

NoelDeMartin (approved):
I think this is ready for merge, but I'd like to solve a couple of questions before moving forward.

src/core/features/course/services/course.ts
crazyserver: I think this comes from very old version, first versions of the app (in 2015) https://github.com/moodlehq/moodleapp/commit/93c9f9e5dd4102cddc48d4743aec7450f169bc95

That's why I revised the list.

src/core/directives/collapsible-header.ts
crazyserver: Hm nice catch, it was just a test. Reverted!

NoelDeMartin (approved):

PR #3070, opened 2022-01-25
MOBILE-3971: Improve config db performance

NoelDeMartin:


src/core/classes/database-table.ts
```diff
@@ -0,0 +1,162 @@
...
+
+    constructor(db: SQLiteDB) {
+        this.db = db;
+        this.data = {};
+    }
+
+    /**
+     * Find a record matching the given conditions.
+     *
+     * @param conditions Mathinc conditions or record id.
```
dpalou, line 48: Typo, it says "Mathinc".

src/core/classes/database-table.ts
```diff
@@ -0,0 +1,162 @@
...
+     * @returns Database record.
+     */
+    find(conditions: Partial<DBRecord> | string): DBRecord | null  {
```
dpalou, line 51: Maybe it would be better to have 2 functions, something like _findByPrimaryKey_ and _find_. The first would accept a string the second would accept an object. And the same for the delete function. I think the code will be easier to understand like this.

dpalou (changes requested):

src/core/classes/database-table.ts
NoelDeMartin: Good catch, I don't know how that happened xD

src/core/classes/database-table.ts
NoelDeMartin: I have updated this as we've discussed in private.

dpalou (approved):

PR #3076, opened 2022-01-28
Mobile 3833

dpalou:


crazyserver (commented):
Thanks Dani! There's hard work on here!

I would like to see a singleton ton the deep link manager but  I will leave the decision to you since your code is correct. Singleton would make function clearer (since constructor it does not say nothing by his own).

Just answer me and I will approve changes.

dpalou:
Hi Pau,

I think making it a singleton wouldn't benefit us much because we'd still need 2 calls: one to read the parameters and another to treat them. In some pages (home, group-conversations, discussions) those 2 actions aren't consecutive.

Also, making it static has a small problem, although it's a little edge:

1- The app enters a site with a deep link. The default tab is "home", so the home page starts and the app reads and stores the deep link parameters in the manager.
2- The user changes the tab while "home" is still loading. This new tab doesn't have the deep link parameters, so we have 2 options here:

Option A: We delete the stored parameters.
3A- The new tab doesn't open the deep link because there are no parameters. And the "home" tab either because the parameters were deleted. So the deep link isn't opened.

Option B: We don't delete the stored parameters
3B- The first tab that calls the function to treat the deep link will open the page, and the other won't. So we have a race condition here, if both tabs call the "treat" function asynchronously then we won't be able to know which one will open the link.

For this reasons I think it's better to have an instance per page.

crazyserver:
Thanks Dani!! So let's merge!

crazyserver (approved):

PR #3087, opened 2022-02-01
MOBILE-3915 course: Align icons on course index

crazyserver:


src/core/features/course/components/course-index/course-index.ts
```diff
@@ -70,28 +71,31 @@ export class CoreCourseCourseIndexComponent implements OnInit {
             return;
         }
 
-        // Collapse all sections first.
-        this.sections.forEach((section) => section.expanded = false);
+        // Clone sections to add information.
+        this.sectionsToRender = Object.assign([], this.sections);
```
dpalou, line 75: This instruction doesn't clone each item, only the array. If you add a new item to "sectionsToRender" it won't be added to "sections", but if you modify a section inside "sectionsToRender" it will also be modified inside "sections". So I understand this is wrong.

dpalou (changes requested):

dpalou (dismissed):

src/core/features/course/components/course-index/course-index.ts
crazyserver: Hi Dani! Check if you like the current solution :-)

dpalou (approved):

PR #3101, opened 2022-02-08
Mobile 3808

crazyserver:


src/core/components/download-refresh/core-download-refresh.html
```diff
@@ -1,14 +1,14 @@
...
     <ion-button *ngIf="status == statusNotDownloaded" fill="clear" (click)="download($event, false)" color="dark" @coreShowHideAnimation
-        [attr.aria-label]="(statusTranslatable || 'core.download') | translate" [size]="size">
+        [attr.aria-label]="(statusTranslatable || 'core.download') | translate">
         <ion-icon slot="icon-only" name="cloud-download" aria-hidden="true"></ion-icon>
     </ion-button>
 
     <!-- Refresh button. -->
     <ion-button *ngIf="status == statusOutdated || (status == statusDownloaded && !canTrustDownload)" fill="clear"
         (click)="download($event, true)" color="dark" @coreShowHideAnimation
-        [attr.aria-label]="(statusTranslatable || 'core.refresh') | translate" [size]="size">
```
dpalou, line 11: You're removing this from the template, but "size" is an input of the component. We should probably mark it as deprecated.

src/core/services/utils/utils.ts
```diff
@@ -1558,6 +1557,31 @@ export class CoreUtilsProvider {
...
+     * @param fn Function to throttle.
+     * @param duration Time that must pass until the function is called.
+     * @return Throttled function.
+     */
+    throttle<T extends unknown[]>(fn: (...args: T) => unknown, duration: number): (...args: T) => void {
+        let shouldWait = false;
+
+        const throttled = (...args: T): void => {
+            if (!shouldWait) {
+                fn.apply(this, args);
```
dpalou, line 1572: I think you should pass _null_ instead of _this_, otherwise you will execute the function with the context "CoreUtils". If the function requires a context we expect to receive a function with the context already bound, so this change shouldn't make much difference, but I think it's better not to use a wrong context.

dpalou (changes requested):

src/core/components/download-refresh/core-download-refresh.html
crazyserver: True and done!

src/core/services/utils/utils.ts
crazyserver: Thanks!

dpalou (approved):

PR #3115, opened 2022-02-14
Mobile 3970

crazyserver:


src/core/features/course/pages/preview/preview.html
```diff
@@ -8,19 +8,22 @@ <h1>
                 {{'core.course.coursesummary' | translate}}
             </h1>
         </ion-title>
+        <ion-buttons slot="end" *ngIf="avoidOpenCourse">
```
dpalou, line 11: IMO it looks a bit weird that a variable called "avoidOpenCourse" decides if it's a modal or not. Since this is no longer a parameter I would rename the variable to something that makes more sense, like "isModal".

src/core/features/course/course-lazy.module.ts
```diff
@@ -32,9 +32,9 @@ const routes: Routes = [
         loadChildren: () => import('./pages/list-mod-type/list-mod-type.module').then(m => m.CoreCourseListModTypePageModule),
     },
     {
-        path: ':courseId/preview',
+        path: ':courseId/summary',
```
dpalou, line 35: Can you document this route change in upgrade.txt?

dpalou (changes requested):

src/core/features/course/course-lazy.module.ts
crazyserver: Done!

src/core/features/course/pages/preview/preview.html
crazyserver: ok! Done!

dpalou (approved):

PR #3118, opened 2022-02-15
MOBILE-3982 core: Fix collapsible header flicker

NoelDeMartin:


src/core/directives/collapsible-header.ts
```diff
@@ -260,55 +260,64 @@ export class CoreCollapsibleHeaderDirective implements OnDestroy {
...
-     *
-     * @param scrollTop Scroll top measure.
+     * Update collapse progress according to the current scroll position.
      */
-    protected onScroll(
-        scrollTop: number,
-    ): void {
-        if (!this.title || !this.contentH1) {
+    protected async updateCollapseProgress(): Promise<void> {
+        const scrollElement = await this.content?.getScrollElement();
```
crazyserver, line 277: I think this is something we should save for later use, since it won't change and calculating elements on dom can compromise performance.

crazyserver (changes requested):

src/core/directives/collapsible-header.ts
NoelDeMartin: I don't think this affects performance too much because the content already has a reference to the scroll, it doesn't need to search it every time. And I preferred to do it this way because it's awkward to write "computed properties" in Angular, so I wanted to avoid it.

In any case, I followed your suggestion and I'm storing it in a variable that depends on the content.

crazyserver: oOOk.. ;-)

crazyserver (approved):

PR #3141, opened 2022-02-24
MOBILE-3982 core: Refactor collapsible-header

NoelDeMartin:


src/theme/components/collapsible-header.scss
```diff
@@ -0,0 +1,74 @@
...
+        h1 {
+            opacity: 0;
+        }
+
+    }
+
+    &.has-floating-title {
+
+        .collapsible-header--expanded {
+            height: calc(var(--collapsible-header-expanded-height) * (1 - var(--collapsible-header-progress)));
```
crazyserver, line 36: I would not modify the height of any element to avoid flickering. This is also causing that the title is placed over some content text.

src/theme/components/collapsible-header.scss
```diff
@@ -0,0 +1,74 @@
...
+            opacity: calc(1 - var(--collapsible-header-progress));
+            transform: translateY(calc(var(--collapsible-header-title-top-delta) * var(--collapsible-header-progress)));
```
crazyserver, line 38: I think transform: none is always winning this rule so, it's not needed.

src/theme/components/collapsible-header.scss
```diff
@@ -0,0 +1,74 @@
...
+            flex-shrink: 0;
```
crazyserver, line 39: Why do we need that?

src/theme/components/collapsible-header.scss
```diff
@@ -0,0 +1,74 @@
...
+            overflow: visible;
```
crazyserver, line 40: If we don't set the height this is also not needed

crazyserver (approved):

src/core/directives/collapsible-header.ts
```diff
@@ -12,336 +12,335 @@
...
-        } else {
-            // Checking position over parent because transition may not be finished.
-            const contentH1Position = this.contentH1.getBoundingClientRect().left - this.content.getBoundingClientRect().left;
-            const headerH1Position = headerH1.getBoundingClientRect().left - this.header.getBoundingClientRect().left;
-
-            this.h1StartDifference = Math.round(contentH1Position - (headerH1Position + parseFloat(headerH1Styles.paddingLeft)));
-        }
+        const pageBoundingBox = this.page.getBoundingClientRect();
+        const collapsedHeaderTitle = this.collapsedHeader.querySelector('h1') as HTMLElement;
+        const collapsedTitleStyles = getComputedStyle(collapsedHeaderTitle);
```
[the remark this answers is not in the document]
NoelDeMartin: I think they are still needed because we're using them when the transition is halfway done, so that the change is not so jarring at the end.

src/core/directives/collapsible-header.ts
```diff
@@ -12,336 +12,335 @@
...
+        const collapsedHeaderTitleBoundingBox = collapsedHeaderTitle.getBoundingClientRect();
+        const floatingTitle = collapsedHeaderTitle.cloneNode(true) as HTMLElement;
+        const expandedHeaderTitle = this.expandedHeader.querySelector('h1') as HTMLElement;
+        const expandedTitleStyles = getComputedStyle(expandedHeaderTitle);
+        const expandedHeaderHeight = this.expandedHeader.clientHeight;
+        const expandedHeaderTitleBoundingBox = expandedHeaderTitle.getBoundingClientRect();
+        const [collapsedFontStyles, expandedFontStyles] = Array
+            .from(collapsedTitleStyles)
+            .filter(property => property.startsWith('font-') || property.startsWith('letter-'))
```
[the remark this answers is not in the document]
NoelDeMartin: Done.

src/core/directives/collapsible-header.ts
```diff
@@ -12,336 +12,335 @@
...
+            page.style.setProperty('--collapsible-header-progress', `${progress}`);
+            page.classList.toggle('is-frozen', frozen);
+            page.classList.toggle('is-collapsed', collapsed);
 
-    /**
-     * @inheritdoc
-     */
-    ngOnDestroy(): void {
-        this.loadingObserver.off();
+            Object
```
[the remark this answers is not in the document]
NoelDeMartin: As mentioned above, this is intended so that the change is not so jarring at the end.

src/theme/components/collapsible-header.scss
NoelDeMartin: Thanks, I fixed the issue you mentioned, but this is still needed in some situations (for example, in the course home where the expanded header is outside of the scroll).

src/theme/components/collapsible-header.scss
NoelDeMartin: I have double-checked and it is being used, `transform: none` only applies when the expanded header is sitting within the scroll content.

src/theme/components/collapsible-header.scss
NoelDeMartin: As mentioned above, we need to set the height in some scenarios.

NoelDeMartin:
@crazyserver The diffs from history are a bit confusing because I did a couple of force-push, here's the diff since your last review: https://github.com/moodlehq/moodleapp/compare/a7b54534f0cae7a216366494cd63f26fc5e69a47..585f0b7afe9fe702a12ff8c42e1858b299ef0f55

src/theme/components/collapsible-header.scss
```diff
@@ -0,0 +1,86 @@
...
+
+            @include core-transition(height, 200ms, linear);
+            @include core-transition(transform, 200ms, linear);
+
+            .collapsible-header--floating-title {
+                opacity: 1;
+                width: calc(var(--collapsible-header-floating-title-width) + var(--collapsible-header-progress) * var(--collapsible-header-floating-title-width-delta));
+
+                @include core-transition(width, 200ms, linear);
```
crazyserver, line 44: You cannot have two rules for the same property. Please use "all" since core-transition is not prepared to have commas on the first variable.

src/theme/components/collapsible-header.scss
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    &.is-active {
+
+        .collapsible-header--expanded {
+            height: calc(var(--collapsible-header-expanded-height) * (1 - var(--collapsible-header-progress)));
+            transform: translateY(calc(var(--collapsible-header-expanded-y-delta) * var(--collapsible-header-progress)));
+            overflow: visible;
+
+            @include core-transition(height, 200ms, linear);
```
crazyserver, line 37: See comment bellow

src/theme/components/collapsible-header.scss
```diff
@@ -0,0 +1,86 @@
...
+                @include core-transition(transform, 200ms, linear);
+            }
+
+            .collapsible-header--original-title {
+                opacity: 0;
+            }
+
+            & > *:not(.collapsible-header--floating-title-wrapper),
+            .collapsible-header--floating-title-wrapper > *:not(.collapsible-header--floating-title):not(.collapsible-header--original-title) {
+                opacity: calc(1 - var(--collapsible-header-progress));
```
crazyserver, line 54: Please add a transition for the opacity here.

crazyserver (changes requested):
That's it! I think there are all minor changes :-) Great work!

src/theme/components/collapsible-header.scss
```diff
@@ -0,0 +1,86 @@
...
+    --collapsible-header-collapsed-height: 0px;
+    --collapsible-header-expanded-y-delta: 0px;
+    --collapsible-header-expanded-height: 0px;
+    --collapsible-header-floating-title-top: 0px;
+    --collapsible-header-floating-title-left: 0px;
+    --collapsible-header-floating-title-width: 0px;
+    --collapsible-header-floating-title-x-delta: 0px;
+    --collapsible-header-floating-title-width-delta: 0px;
+
+    .collapsible-header--expanded .collapsible-header--floating-title {
```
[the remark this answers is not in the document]
NoelDeMartin: Ok, I have updated the class names and opened the following issue https://tracker.moodle.org/browse/MOBILE-4004

src/core/singletons/components-registry.ts
```diff
@@ -0,0 +1,53 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Registry to keep track of component instances.
+ */
+export class CoreComponentsRegistry {
```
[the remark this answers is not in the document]
NoelDeMartin: Ok, I have deprecated the old ones and moved this to a separate commit.

src/theme/components/collapsible-header.scss
NoelDeMartin: Nice catch, I didn't think about that :sweat_smile: We cannot use `all` because it was causing a flickering when the title is substituted (because it's introduced with opacity: 0 to measure the offset). So what I did is refactor the `core-transition` mixin to accept an array of values, it should be working correctly now.

src/theme/components/collapsible-header.scss
NoelDeMartin: Same

src/theme/components/collapsible-header.scss
NoelDeMartin: Done.

crazyserver (approved):

PR #3144, opened 2022-02-28
Mobile 3149

crazyserver:


src/addons/mod/quiz/components/index/index.ts
```diff
@@ -159,7 +159,7 @@ export class AddonModQuizIndexComponent extends CoreCourseModuleMainActivityComp
         }
 
         // Prefetch the quiz.
-        this.showStatusSpinner = true;
+        const modal = await CoreDomUtils.showModalLoading();
```
dpalou, line 162: I don't like this change, the modal will block the user (he won't be able to come back for example).

dpalou (changes requested):

src/addons/mod/quiz/components/index/index.ts
crazyserver: Thanks Dani! Done!

dpalou (approved):

PR #3145, opened 2022-02-28
Mobile 3998

dpalou:


src/addons/mod/imscp/components/index/addon-mod-imscp-index.html
```diff
@@ -13,25 +9,36 @@
...
+                <ion-label [class.core-bold]="!item.href">
+                    <p>
+                        <span class="ion-padding-start" *ngFor="let i of getNumberForPadding(item.level)"></span>
+                        {{item.title}}
+                    </p>
+                </ion-label>
+            </ion-item>
+        </ion-list>
+
+        <ion-button class="ion-margin" expand="block" (click)="openImscp()">
```
crazyserver, line 34: Recently MOBILE-3149 has been integrated, can you add the buttons to the collapsible-footer here?

crazyserver (changes requested):
Additionally to previous comment: Probably we should open an issue to erase last viewed data from course/activity from the downloads manager.

src/addons/mod/imscp/components/index/addon-mod-imscp-index.html
dpalou: Done.

dpalou:
The last viewed item isn't downloaded data, it could be considered "session" data like the selected order in course overview, so I think it shouldn't be deleted.

crazyserver (approved):

PR #3146, opened 2022-02-28
Mobile 3814

crazyserver:


src/core/features/login/pages/site/site.scss
```diff
@@ -1,5 +1,9 @@
 @import "~theme/globals";
 
+:host {
+    --list-item--max-width: 600px;
```
dpalou, line 4: I didn't notice this variable has a double dash: "item--max". I guess it should be a single dash, right?

dpalou (changes requested):

src/core/features/login/pages/site/site.scss
crazyserver: Nice catch!!!

dpalou (approved):

PR #3172, opened 2022-03-11
Mobile 3814

crazyserver:


src/core/components/loading/loading.ts
```diff
@@ -77,50 +81,54 @@ export class CoreLoadingComponent implements OnInit, OnChanges, AfterViewInit {
...
+     */
+    async changeState(loaded: boolean): Promise<void> {
+        await CoreUtils.nextTick();
+
+        this.element.classList.toggle('core-loading-loaded', loaded);
+        this.content?.nativeElement.classList.add('core-loading-content', loaded);
+
+        await CoreUtils.nextTick();
+
+        // Trigger the event after a timeout since the elements inside ngIf haven't been added to DOM yet.
```
dpalou, line 113: This comment doesn't make sense anymore, there is no timeout. You should rephrase it to mention the nextTick, e.g. "Wait for next tick before triggering the event to make sure ngIf elements have been added to the DOM."

src/core/components/loading/loading.ts
```diff
@@ -77,50 +81,54 @@ export class CoreLoadingComponent implements OnInit, OnChanges, AfterViewInit {
...
+        CoreEvents.trigger(CoreEvents.CORE_LOADING_CHANGED, <CoreEventLoadingChangedData> {
+            loaded: !!this.hideUntil,
```
dpalou, line 115: You can use the "loaded" parameter instead of hideUntil, right? That way there's no need to put a double negation.

src/core/directives/collapsible-header.ts
```diff
@@ -111,30 +117,36 @@ export class CoreCollapsibleHeaderDirective implements OnInit, OnChanges, OnDest
...
+    protected initializePage(): void {
+        if (!this.collapsedHeader.parentElement) {
             throw new Error('[collapsible-header] Couldn\'t get page');
         }
 
         // Find element and prepare classes.
         this.page = this.collapsedHeader.parentElement;
-
         this.page.classList.add('collapsible-header-page');
+
```
dpalou, line 136: Do you think it could make sense to have a collapsible-header with an ngIf or similar so it's created after the DidEnter has been called? In that case the promise would never be resolved.

If it doesn't make sense then I would add some timeout to write an error in console if the DidEnter hasn't been called after a certain time.
If it could happen, then I guess we should add a timeout to resolve the promise if the DidEnter hasn't been called after a certain time.

src/core/singletons/components-registry.ts
```diff
@@ -36,14 +39,47 @@ export class CoreComponentsRegistry {
...
+            .map(element => CoreComponentsRegistry.resolve<T>(element));
+
+        await Promise.all(components.map(component => {
+            if (!component) {
+                return;
+            }
+
+            return component[fnName].apply(component, params);
+        }));
+        await CoreUtils.nextTick();
```
dpalou, line 80: Why is this nextTick needed? All components should be resolved without the nextTick, right?

src/core/singletons/components-registry.ts
```diff
@@ -48,15 +48,15 @@ export class CoreComponentsRegistry {
     }
 
     /**
-     * Waits all elements to be resolved.
+     * Waits all elements to be rendered.
      *
      * @param element Parent element where to search.
      * @param selector Selector to search on parent.
      * @param fnName Function that needs to get resolved.
```
dpalou, line 55: With the change in the function name I think you should also change the descrirption of the _fnName_ and _params_ parameters.

src/core/services/utils/dom.ts
```diff
@@ -91,6 +92,45 @@ export class CoreDomUtilsProvider {
...
+
+        // Disconnect observer for performance reasons.
+        const timeout = window.setTimeout(() => {
+            inDomPromise.reject();
+            observer.disconnect();
+        }, 5000);
+
+        const observer = new MutationObserver(() => {
+            root = element.getRootNode();
+            if (root == document) {
```
dpalou, line 122: Please use triple equal instead.

src/core/services/utils/dom.ts
```diff
@@ -91,6 +92,45 @@ export class CoreDomUtilsProvider {
         this.debugDisplay = debugDisplay != 0;
     }
 
+    /**
+     * Wait an element to be in dom of another element.
+     *
+     * @param element Element to wait.
+     * @return Promise resolved when added. It will be rejected after a timeout of 5s.
+     */
```
dpalou, line 100: Have you tried this function with shadow DOM? I'm not sure if getRootNode will always return the document when it's inside shadow DOM.

dpalou (changes requested):

src/core/directives/collapsible-header.ts
crazyserver: It should never happen but it's great to have it to avoid future problems... thanks!

src/core/singletons/components-registry.ts
crazyserver: I've added the comment: Wait for next tick to ensure components are completely rendered.

src/core/services/utils/dom.ts
crazyserver: I think it's not going to be used but, since you requested I changed the function so now it works with shadow dom too.

dpalou (approved):

PR #3177, opened 2022-03-15
Mobile 3814

crazyserver:


dpalou (approved):

src/core/directives/collapsible-header.ts
```diff
@@ -228,7 +228,7 @@ export class CoreCollapsibleHeaderDirective implements OnInit, OnChanges, OnDest
             throw new Error('[collapsible-header] Couldn\'t create floating title');
         }
 
-        this.page.classList.remove('is-active');
+        this.page.classList.remove('collapsible-header-is-active');
```
NoelDeMartin, line 231: How come you're changing this? I know we still don't have a convention for CSS class names, but at least I think we should be consistent within one component, so if you change this class you should also change is-frozen, is-collapsed, etc.

src/core/singletons/events.ts
```diff
@@ -117,6 +117,9 @@ export class CoreEvents {
     static readonly APP_LAUNCHED_URL = 'app_launched_url'; // App opened with a certain URL (custom URL scheme).
     static readonly FILE_SHARED = 'file_shared';
     static readonly KEYBOARD_CHANGE = 'keyboard_change';
+    /**
+     * @deprecated since app 4.0. Use CoreComponentsRegistry promises instead.
```
NoelDeMartin, line 121: I know this change hasn't been introduced in this commit, but I don't think it's the job of the components registry to know when a component has been rendered or not. Maybe we could move it somewhere else? Or even create a new function/singleton if necessary.

src/core/directives/collapsible-header.ts
crazyserver: Since it's added in a page it may become unrelated, we have other is-active but applied to the element it requested.

src/core/singletons/events.ts
crazyserver: I think it's not necessary to create a new class for that, but we can work on it later on.

src/core/directives/collapsible-header.ts
NoelDeMartin: True, I hadn't noticed that, in this case the component then is the `collapsible-header-page`, so I guess the proper class name should be `collapsible-header-page-is-active`, and the same applies to the rest because they are also placed in the page (`collapsible-header-page-is-frozen`, etc.).

src/core/singletons/events.ts
NoelDeMartin: ok :(

src/core/directives/collapsible-header.ts
crazyserver: True! Changed

NoelDeMartin (approved):

PR #3181, opened 2022-03-15
MOBILE-3833 core: Implement cancellable promise

NoelDeMartin:


src/core/classes/promise.ts
```diff
@@ -0,0 +1,69 @@
...
+ * Base class to use for implementing custom Promises.
+ */
+export abstract class CorePromise<T = unknown> implements Promise<T> {
+
+    protected nativePromise: Promise<T>;
+    protected resolvePromise!: (result: T) => void;
+    protected rejectPromise!: (error?: Error) => void;
+
+    constructor(nativePromise?: Promise<T>) {
+        this.nativePromise = nativePromise ?? new Promise((resolve, reject) => {
```
crazyserver, line 25: resolvePromise and rejectPromise can be null if nativePromise is passed here, but they are declared not nullable.

crazyserver (changes requested):

src/core/classes/promise.ts
NoelDeMartin: You're right! I didn't initially intend to make the `nativePromise` optional, but [there were some errors](https://github.com/moodlehq/moodleapp/runs/5557448943?check_suite_focus=true) with the way I was initializing the promise in the subclass and my refactor introduced this issue. I've gone back to making it required, and it should be fine now. I don't like how the `CorePromisedValue` constructor turned out, but I'm not sure how to make it more elegant so I guess that should work for now.

crazyserver (approved):

PR #3194, opened 2022-03-22
MOBILE-3833 courses: Add site home link handler & fix site listing styles

NoelDeMartin:


src/core/features/courses/services/handlers/site-link.ts
```diff
@@ -0,0 +1,59 @@
...
+import { CoreMainMenuHomeHandlerService } from '@features/mainmenu/services/handlers/mainmenu';
+import { CoreSiteHomeHomeHandler, CoreSiteHomeHomeHandlerService } from '@features/sitehome/services/handlers/sitehome-home';
+import { CoreNavigator } from '@services/navigator';
+import { makeSingleton } from '@singletons';
+
+/**
+ * Handler to treat links to site home.
+ */
+@Injectable({ providedIn: 'root' })
+export class CoreCoursesSiteLinkHandlerService extends CoreContentLinksHandlerBase {
```
dpalou, line 27: I think this handler should be in the _sitehome_ feature. We already have a link handler in there, _CoreSiteHomeIndexLinkHandlerService_ (that one is currently using hardcoded paths, it should be fixed).

You can either use a single handler for both URLs or keep them separate, as you prefer.

dpalou (changes requested):

src/core/features/courses/services/handlers/site-link.ts
NoelDeMartin: Thanks I hadn't noticed that, I've put both in the same handler.

NoelDeMartin:
@dpalou I've applied your suggestions and also added a new commit to fix some styles, please take another look.

PR #3212, opened 2022-03-30
Mobile 4017

alfonso-salces:


.vscode/settings.json
```diff
@@ -22,5 +22,6 @@
         "moodle.config.json": "jsonc",
         "moodle.config.*.json": "jsonc",
     },
+    "files.eol": "\n",
```
NoelDeMartin, line 25: I think this should go in the section above (called "Formatting").

NoelDeMartin (changes requested):

.vscode/settings.json
alfonso-salces: Done! thank you

src/core/classes/tests/error.test.ts
```diff
@@ -30,15 +32,14 @@ describe('CoreError', () => {
         } catch (e) {
             error = e;
         }
-
```
NoelDeMartin, line 33: I believe you've removed this line by mistake?

src/testing/utils.ts
```diff
@@ -245,3 +246,12 @@ export async function renderWrapperComponent<T>(
 
     return renderTemplate(component, `<${tag} ${inputAttributes}></${tag}>`, config);
 }
+/**
```
NoelDeMartin, line 249: I think the JSDocs should be added in the same commit as the method, it'd also be better to have an empty line between functions (I've added a new linting rule to enforce that in https://github.com/moodlehq/moodleapp/pull/3214, but it's not merged yet)

NoelDeMartin (changes requested):

src/core/classes/tests/error.test.ts
alfonso-salces: Nice catch! thanks you

src/testing/utils.ts
alfonso-salces: Done!

NoelDeMartin (approved):
I see you ended up squashing both commits, I think it was fine with two but I'll go ahead with merging anyways so that we get this done :).

PR #3215, opened 2022-03-31
Mobile 3833

crazyserver:


src/core/classes/tabs.ts
```diff
@@ -232,29 +225,83 @@ export class CoreTabsBaseComponent<T extends CoreTabBase> implements OnInit, Aft
...
-        this.slidesSwiper = await this.slides?.getSwiper();
-        this.slidesSwiper.once('progress', () => {
-            this.slidesSwiperLoaded = true;
-            this.calculateSlides();
-        });
+        if (!this.initialized || !this.slidesElement) {
+            return;
+        }
 
+        // Initialize slider.
```
dpalou, line 294: This comment isn't needed anymore in here.

src/core/classes/tabs.ts
```diff
@@ -339,17 +384,23 @@ export class CoreTabsBaseComponent<T extends CoreTabBase> implements OnInit, Aft
...
 
         this.maxSlides = 3;
-        let width = this.slidesSwiper.width;
+        await CoreUtils.nextTick();
+
+        let width: number = this.slidesElement.getBoundingClientRect().width;
         if (!width) {
-            this.slidesSwiper.updateSize();
-            width = this.slidesSwiper.width;
+            const slidesSwiper = await this.slides?.getSwiper();
```
dpalou, line 403: If _this.slides_ is optional it means slidesSwiper could be undefined. So the next calls to _slidesSwiper_ should use "?", or maybe you should stop the execution if not found, up to you.

src/core/components/tabs-outlet/core-tabs-outlet.html
```diff
@@ -6,10 +6,9 @@
                 <ion-icon *ngIf="showPrevButton" name="fas-chevron-left" [attr.aria-label]="'core.previous' | translate"></ion-icon>
             </ion-col>
             <ion-col class="ion-no-padding" size="10">
-                <ion-slides (ionSlideDidChange)="slideChanged()" [options]="slidesOpts" [dir]="direction" role="tablist"
-                    [attr.aria-label]="description">
+                <ion-slides [options]="slidesOpts" [dir]="direction" role="tablist" [attr.aria-label]="description">
                     <ng-container *ngFor="let tab of tabs">
-                        <ion-slide role="presentation" [id]="tab.id! + '-tab'" class="tab-slide" tabindex="-1"
+                        <ion-slide role="presentation" class="tab-slide" [id]=" tab.id! + '-tab'" tabindex="-1"
```
dpalou: There's an unneeded space in the id.

src/core/components/tabs-outlet/core-tabs-outlet.html
```diff
@@ -1,37 +1,36 @@
 <ion-tabs class="hide-header">
-    <ion-tab-bar slot="top" class="core-tabs-bar" [hidden]="!tabs || numTabsShown <= 1" #tabBar>
+    <ion-tab-bar slot="top" [hidden]="hideUntil && (!tabs || numTabsShown <= 1)">
```
dpalou, line 2: You removed the "core-tabs-bar" class, this could break remote themes. I'm not saying we shouldn't remove it, but we need to take into account the implications of doing so.

dpalou (changes requested):

dpalou (approved):

PR #3228, opened 2022-04-04
Mobile 3833

alfonso-salces:


crazyserver (changes requested):

src/addons/storagemanager/pages/course-storage/course-storage.scss
```diff
@@ -17,6 +16,12 @@
         font-size: 1.2rem;
     }
 
+    .hidden-content {
+        &.card-content-md {
```
crazyserver, line 20: all classes with md or ios are dependant on the platform: Android Material Design or iOS so this change would change the padding only on Material design. I sugggest here to use the tag (ion-card-content) name instead.

crazyserver (changes requested):

crazyserver (approved):

PR #3252, opened 2022-04-14
MOBILE-3833: Improve navigator canLeave

NoelDeMartin:


NoelDeMartin:
After discussing it with @dpalou, we decided to close this in favor of [MOBILE-4028](https://tracker.moodle.org/browse/MOBILE-4028). The main issue with the current approach is that the `CanLeave` guard is activated even if the user already confirmed, which may cause some issues. It'd be more correct to clean up the page before actually leaving to avoid asking for a second confirmation, but using a separate route is simpler.

PR #3262, opened 2022-04-21
[4.1] Mobile 4021 notifications: refactor to add split-view and swipe

alfonso-salces:


src/addons/notifications/pages/list/list.ts
```diff
@@ -56,18 +55,33 @@ export class AddonNotificationsListPage implements OnInit, OnDestroy {
...
+
     /**
      * @inheritdoc
      */
-    ngOnInit(): void {
-        this.fetchNotifications();
+    async ngAfterViewInit(): Promise<void> {
+        await this.fetchInitialNotifications();
+
+        this.notifications.start(this.splitView);
```
NoelDeMartin, line 78: If this is called, we should call `this.notifications.destroy()` inside the `ngOnDestroy` method to clean up references to the source and anything else the manager needs to clean up.

src/addons/notifications/classes/notifications-source.ts
```diff
@@ -0,0 +1,38 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { CoreRoutedItemsManagerSource } from '@classes/items-management/routed-items-manager-source';
+import { AddonNotifications } from '../services/notifications';
+import { AddonNotificationsHelper, AddonNotificationsNotificationToRender } from '../services/notifications-helper';
+
+export class AddonsNotificationsNotificationsSource extends CoreRoutedItemsManagerSource<AddonNotificationsNotificationToRender> {
+
+    protected async loadPageItems(page: number): Promise<{
```
NoelDeMartin, line 21: Missing docblocks here (both in the class and the methods). For methods, you can simply use:

```ts
/**
 * @inheritdoc
 */
```

NoelDeMartin (changes requested):

src/addons/notifications/pages/notification/notification.ts
```diff
@@ -110,6 +96,48 @@ export class AddonNotificationsNotificationPage implements OnInit {
...
+        try {
+            const pushNotification: AddonNotificationsNotificationData | undefined = CoreNavigator.getRouteParam('notification');
+
+            return this.loadNotifications(
+                pushNotification
+                    ? Number(pushNotification?.savedmessageid)
+                    : CoreNavigator.getRequiredRouteNumberParam('id'),
+            );
+        } catch (error) {
+            CoreNavigator.back();
```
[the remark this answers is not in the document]
alfonso-salces: Nice idea, moved to ngOnInit method! ✅

src/addons/notifications/pages/notification/notification.ts
```diff
@@ -55,42 +56,27 @@ export class AddonNotificationsNotificationPage implements OnInit {
...
-            const foundNotification = result?.notifications.find(notif => notif.id === notifId);
-            if (foundNotification) {
-                notification = AddonNotificationsHelper.formatNotificationText(foundNotification);
-            }
-        }
-
         if ('subject' in notification) {
             this.subject = notification.subject;
-            this.content = notification.mobiletext || notification.fullmessagehtml;
+            this.content = notification.mobiletext || notification?.fullmessagehtml;
```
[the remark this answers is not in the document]
alfonso-salces: Nice catch, that was added at the start trying some things and forgot remove it

src/addons/notifications/pages/list/list.ts
alfonso-salces: Done! ✅

src/addons/notifications/classes/notifications-source.ts
alfonso-salces: Done, thank you!

src/addons/notifications/pages/notification/notification.ts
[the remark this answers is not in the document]
NoelDeMartin: There are still a fair amount of changes that aren't necessary. I'm not against changing unrelated code if it's to clean it up, but in this case we're only changing some small details so I'd prefer to avoid introducing that type of changes.

NoelDeMartin (changes requested):
A couple of comments still haven't been addressed, the most important being that notification objects that are not present in the source will not work after the refactor. I think you should be able to write a Behat test as well to reproduce this scenario, and I think it's failing now after the refactor.

PR #3268, opened 2022-04-22
Mobile 3881

crazyserver:


src/addons/block/timeline/components/events/events.ts
```diff
@@ -89,20 +85,41 @@ export class AddonBlockTimelineEventsComponent implements OnChanges {
...
             if (end) {
-                return start <= event.timesort && event.timesort < end;
+                timeBetween = start <= event.timesort && event.timesort < end;
+            } else {
+                timeBetween = start <= event.timesort;
+            }
+
+            if (!timeBetween) {
+                return false;
+            }
```
NoelDeMartin, line 109: This can be refactored to avoid creating a "timeBetween" variable at all:

```js
if (start > event.timesort || (end && event.timesort >= end)) {
    return false;
}
```

NoelDeMartin (approved):
I only saw a small improvement and we can merge.

src/addons/block/timeline/components/events/events.ts
crazyserver: Nice! Thanks!

src/addons/block/timeline/components/timeline/timeline.ts
```diff
@@ -172,14 +174,27 @@ export class AddonBlockTimelineComponent extends CoreBlockBaseComponent implemen
...
+        } catch {
+            this.gradePeriodAfter = 0;
+            this.gradePeriodBefore = 0;
+        }
+
         // Do not filter courses by date because they can contain activities due.
         this.timelineCourses.courses = await CoreCoursesHelper.getUserCoursesWithOptions();
 
         if (this.timelineCourses.courses.length > 0) {
+            // Filter only in progress courses.
```
NoelDeMartin, line 189: I think this filter could be applied before the if. Ideally, we would only filter the courses once but I suppose the filter is split in two to avoid calling `AddonBlockTimeline.getActionEventsByCourses` if the list is empty. I'm not sure about the cost of calling that, if it's low I'd prefer to call a single filter. But if it's important to avoid it, then we should filter as much as possible before the if.

NoelDeMartin (approved):

src/addons/block/timeline/components/timeline/timeline.ts
crazyserver: Yep, it's done by this reason to reduce bandwidth and also caching non necessary data

NoelDeMartin (approved):

PR #3278, opened 2022-05-04
Mobile 4061: Include behat tests on app code

crazyserver:


NoelDeMartin (changes requested):
I think the commits are ok, but you should keep the history for the files you've moved. Using git, it is possible to move files with history from one repo to another. I did it in the [previous repository](https://github.com/moodlehq/moodle-local_moodlemobileapp/commits/7d1fd94c19b73db722c2eebdb4f4ef668b4de884), but I don't remember exactly how. Looking at my browser history, I visited these two questions in StackOverflow so they should be useful:

- [How to move a file from one git repository to another while preserving history](https://stackoverflow.com/questions/10524578/how-to-move-a-file-from-one-git-repository-to-another-while-preserving-history)
- [How to move some files from one git repo to another (not a clone), preserving history](https://stackoverflow.com/questions/1365541/how-to-move-some-files-from-one-git-repo-to-another-not-a-clone-preserving-hi)

crazyserver:
Thanks Noel! I did what you said and added also the tests from the plugin.

crazyserver:
To make your review easier:
- Forget about commits before 5th May
- Merge commit contains files from the old repo with renames.

src/tests/service/behat-blocking.ts
```diff
@@ -0,0 +1,242 @@
+// (C) Copyright 2015 Moodle Pty Ltd.
```
NoelDeMartin, line 1: This folder should be called `test/services` (with plural, not singular)

src/tests/tests.module.ts
```diff
@@ -0,0 +1,34 @@
...
+import { CoreAppProvider } from '@services/app';
+import { TestsBehatBlockingService } from './service/behat-blocking';
+import { BehatTestsWindow, TestsBehatRuntime } from './service/behat-runtime';
+
+function initializeBehatTestsWindow(window: BehatTestsWindow) {
+    // Make functions publicly available for Behat to call.
+    window.behatInit = TestsBehatRuntime.init;
+}
+
+@NgModule({
```
NoelDeMartin, line 25: I'm not sure about the naming and location of this module. Since it's setting up something for testing, but it doesn't contain tests, it should be located under `src/testing` instead (this folder already exists). At the same time, testing so far in the app made reference to unit/integration tests (jest), but now this will make reference to behat. Should it be called "BehatModule" then? I don't know, maybe it's ok to just call it `TestingModule`.

src/tests/service/behat-runtime.ts
```diff
@@ -0,0 +1,396 @@
...
+
+    /**
+     * Init behat functions and set options like skipping onboarding.
+     *
+     * @param options Options to set on the app.
+     */
+    static init(options?: TestsBehatInitOptions): void {
+        TestsBehatBlocking.init();
+
+        (window as BehatTestsWindow).behat = {
```
NoelDeMartin, line 35: Instead of doing this, wouldn't it be better to assign a `TestsBehatRuntime` instance to behat directly? In particular, it could be done like we're already exposing other services in the `prepare-automated-tests.ts` file.

The `TestsBehatBlocking.init()` and the rest of the initializing (setting config and env overrides) could be called directly from behat, as it was being done here: https://github.com/moodlehq/moodle-local_moodlemobileapp/blob/integration/tests/behat/behat_app.php#L562

Something to keep in mind as well is that sometimes, this may get called when the application has already been launched. In that case, things like `skipOnBoarding` will not have any effects because the onboarding will already be on screen. Looking at the code, that seems to be the case so I'm not entirely sure why it works (maybe this is executed before the app finishes loading? In that case, that's a race condition and we shouldn't depend on that).

local-moodleappbehat/tests/behat/behat_app.php
```diff
@@ -0,0 +1,800 @@
...
+
+    use behat_app_trait;
+
+    /** @var string URL for running Ionic server */
+    protected $ionicurl = '';
+
+    /** @var array Config overrides */
+    protected $appconfig = ['disableUserTours' => true];
+
+    protected $windowSize = '360x720';
```
NoelDeMartin, line 70: Php files use Moodle's styling guide, so this should be `$windowsize` instead. Maybe at some point we should configure [the codechecker](https://github.com/moodlehq/moodle-local_codechecker) as well. There are other similar style errors in the file.

local-moodleappbehat/tests/behat/behat_app.php
```diff
@@ -0,0 +1,800 @@
...
+    /**
+     * @Then I wait the app to restart
+     */
+    public function i_wait_the_app_to_restart() {
+        // Wait window to reload.
+        $this->spin(function() {
+            $result = $this->evaluate_script("return !window.behat;");
+
+            if (!$result) {
+                throw new DriverException('Behat already here.');
```
NoelDeMartin, line 140: What does `Behat already here` mean? How come this needed a refactor, it wasn't working before?

local-moodleappbehat/tests/behat/behat_app.php
```diff
@@ -0,0 +1,800 @@
...
+    /**
+     * Trigger swipe gesture.
+     *
+     * @When /^I swipe to the (left|right) in the app$/
+     * @param string $direction
+     */
+    public function i_swipe_in_the_app(string $direction) {
+        $method = 'swipe' . ucwords($direction);
+
+        $this->evaluate_script("window.behat.getAngularInstance('ion-content', 'CoreSwipeNavigationDirective').$method()");
```
NoelDeMartin, line 238: Why did you add `window.` here? It's not necessary, but I can see that it's used elsewhere in this file. Maybe if we want to standarize, I'd actually not use `window.` anywhere because scripts are already evaluated in the window scope.

local-moodleappbehat/tests/behat/behat_app_trait.php
```diff
@@ -0,0 +1,613 @@
...
+
+    /**
+     * Trigger Angular change detection.
+     */
+    protected function trigger_angular_change_detection() {
+        $this->getSession()->executeScript('ngZone.run(() => {});');
+    }
+
+    /**
+     * Evaludate a script that returns a Promise.
```
NoelDeMartin, line 435: Typo, this should say "Evaluate".

local-moodleappbehat/tests/behat/behat_app.php
```diff
@@ -0,0 +1,800 @@
...
+        $names = $this->getSession()->getWindowNames();
+        if (count($names) !== 2) {
+            throw new DriverException('Expected to see 2 tabs open, not ' . count($names));
+        }
+        // Make sure the browser tab is selected.
+        if ($this->getSession()->getWindowName() !== $names[1]) {
+            $this->getSession()->switchToWindow($names[1]);
+        }
+
+        $this->execute_script('window.close();');
```
NoelDeMartin, line 785: Similar to my comment about removing `window.`, we can also remove `;` at the end of scripts.

Also, I just noticed that sometimes we use `evaluate_script` and others we use `execute_script`. It seems like the only difference is that evaluating returns a value whilst executing doesn't. I think we should also standardize this: either using each one appropriately or using always evaluate (I can't think of any drawbacks).

local-moodleappbehat/tests/behat/behat_app_trait.php
```diff
@@ -0,0 +1,613 @@
...
+        }, false, 60);
+
+        try {
+            // Init Behat JavaScript runtime.
+
+            $initOptions = new StdClass();
+            $initOptions->skipOnBoarding = $options['skiponboarding'] ?? true;
+            $initOptions->configOverrides = $this->appconfig;
+
+            $this->execute_script('window.behatInit(' . json_encode($initOptions) . ');');
```
NoelDeMartin, line 311: As I mentioned in the service file, I don't think we need to do this. Here, we can check that `behat` is actually available in in the window and that should be enough.

local-moodleappbehat/tests/behat/behat_app.php
```diff
@@ -0,0 +1,800 @@
...
+     * @throws ExpectationException Problem with resizing window
+     */
+    public function i_launch_the_app(string $runtime = '') {
+        // Check the app tag was set.
+        if (!$this->has_tag('app')) {
+            throw new DriverException('Requires @app tag on scenario or feature.');
+        }
+
+        // Go to page and prepare browser for app.
+        $this->prepare_browser(['skiponboarding' => empty($runtime)]);
```
NoelDeMartin, line 128: I think the `skiponboarding` override could be configured similarly to `$appconfig` which is disabling user tours by default. The only difference is that that makes reference to environment build config, whilst this is runtime config. Maybe we should also rename some of the variables in this file to make that more clear.

local-moodleappbehat/tests/behat/behat_app.php
```diff
@@ -0,0 +1,800 @@
...
+
+    /** @var array Config overrides */
+    protected $appconfig = ['disableUserTours' => true];
+
+    protected $windowSize = '360x720';
+
+    /**
+     * Opens the Moodle App in the browser and logins (optional).
+     *
+     * @Given I enter the app
```
NoelDeMartin, line 75: I'm not sure if we should accept both nomenclatures here. "I enter" makes reference to an action, so it should be defined as a `@Then`. And "I entered" makes reference to some state, so it should be defined as a `@Given`. But the implementation should be different as well, an action should really perform the action using the UI whilst some state can be set using shortcuts. The fact that this has now become a shortcut with the same syntax may break some existing tests (if not ours, for some third party). So I think it should be better to keep "I enter" as it was before, and define shortcuts specifically with "I entered".

local-moodleappbehat/tests/behat/behat_app_trait.php
```diff
@@ -0,0 +1,613 @@
...
+
+    /**
+     * Handles the custom URL on the Moodle App (and waits to finish.)
+     *
+     * @param string $path To navigate.
+     * @param string $successXPath The XPath of the element to lookat after navigation.
+     */
+    protected function handle_url_and_wait_page_to_load(string $customurl, string $successXPath = '') {
+        // Instead of using evaluate_async_script, we wait for the path to load.
+        $this->evaluate_script("return window.behat.handleCustomURL('$customurl')");
```
NoelDeMartin, line 540: Instead of implementing shortcuts using actual app functionality, I think it'd be better to implement them directly on the app's side. For example, in this case instead of relying on custom urls we could do something like this in the app's side (pseudocode):


```
CoreSites.setCurrentSite({ siteUrl, token, ... });
CoreNavigator.navigate(path, { reset: true });
```

I guess we can leave it as is for now if changing that is too much work, but I think it's important to implement shortcuts directly instead of relying on existing functionality.

local-moodleappbehat/tests/behat/behat_app_trait.php
```diff
@@ -0,0 +1,613 @@
...
+        $urlscheme = $this->get_mobile_url_scheme();
+        $url = "$urlscheme://link=" . urlencode($CFG->behat_wwwroot.$path);
+
+        $this->handle_url_and_wait_page_to_load($url);
+    }
+
+    /**
+     * Handles the custom URL on the Moodle App (and waits to finish.)
+     *
+     * @param string $path To navigate.
```
NoelDeMartin, line 535: The name of the argument is `$customurl`, not `$path`

local-moodleappbehat/tests/behat/behat_app_trait.php
```diff
@@ -0,0 +1,613 @@
...
+            $listener->on_app_unload();
+        }
+    }
+
+
+    /**
+     * Trigger Angular change detection.
+     */
+    protected function trigger_angular_change_detection() {
+        $this->getSession()->executeScript('ngZone.run(() => {});');
```
NoelDeMartin, line 431: It's ok to encapsulate this given how much it's used, but I've come to believe that we should never need to call this manually. So maybe instead of doing this, we could review where it's necessary and fix it instead. If that's too much work for now, I'd at least create a follow up issue or TODO to improve it in the future.

local-moodleappbehat/tests/behat/behat_app.php
```diff
@@ -0,0 +1,800 @@
...
+    /** @var string URL for running Ionic server */
+    protected $ionicurl = '';
+
+    /** @var array Config overrides */
+    protected $appconfig = ['disableUserTours' => true];
+
+    protected $windowSize = '360x720';
+
+    /**
+     * Opens the Moodle App in the browser and logins (optional).
```
NoelDeMartin, line 73: This should say "logs in" not "logins". Or maybe better, "Opens the Moodle App in the browser and optionally logs in."

local-moodleappbehat/tests/behat/classes/performance_measure.php
```diff
@@ -18,6 +18,8 @@
 use Facebook\WebDriver\Exception\InvalidArgumentException;
 use Moodle\BehatExtension\Driver\WebDriver;
 
+require_once(__DIR__ . '/../behat_app.php');
```
NoelDeMartin, line 21: Why is this necessary now when it wasn't before?

src/tests/behat/performance.feature
```diff
@@ -81,23 +81,17 @@ Feature: Measure performance.
...
-    And I set the field "Your site" to "$WWWROOT" in the app
-    And I press "Connect to your site" in the app
-    And I log in as "student1"
+    Given I entered the app as "student1"
     Then I press "My courses" in the app
     And I should find "Course 1" in the app
 
-    When I reload the page
     And I start measuring "Open Activity"
+    When I reload the page
```
NoelDeMartin, line 89: We want to have the `I reload the page` step before we start measuring, because this is supposed to measure the performance of opening an activity, not the app startup.

src/tests/behat/performance.feature
```diff
@@ -81,23 +81,17 @@ Feature: Measure performance.
...
     And I wait the app to restart
     Then I should find "Course 1" in the app
 
-    When I enter the course "Course 1" in the app
-    And I press "Choice course 1" in the app
+    Given I entered the course "Course 1" in the app
```
NoelDeMartin, line 93: This is supposed to test the performance of opening a course, so it shouldn't use any shortcuts between measurements.

local-moodleappbehat/tests/behat/behat_app.php
```diff
@@ -0,0 +1,800 @@
...
+    /**
+     * User enters a course in the app.
+     *
+     * @Given I enter(ed) the course :coursename as :username in the app
+     * @Given I enter(ed) the course :coursename in the app
+     * @param string $coursename Course name
+     * @throws DriverException If the button push doesn't work
+     */
+    public function i_enter_the_course_in_the_app(string $coursename, ?string $username = null) {
+        global $DB;
```
NoelDeMartin, line 318: This isn't used.

src/core/features/mainmenu/tests/behat/mainmen-311.feature
```diff
@@ -0,0 +1,23 @@
+@app @javascript @lms_upto3.11
```
NoelDeMartin, line 1: This file should be called `mainmenu-311.feature` (it's missing the "u" at the end of "menu").

NoelDeMartin (changes requested):

local-moodleappbehat/tests/behat/behat_app.php
crazyserver: It was working but when failing console shows a general error. So now it explains something. However I changed the message.

local-moodleappbehat/tests/behat/behat_app.php
crazyserver: I'd leave the window then it's easier to understand that something has been added there.

src/tests/service/behat-runtime.ts
crazyserver: Main problem of doing this is that we'll be overriding XMLRequest and listening to mutations while developing, but probably we can separate this on two different functions.

local-moodleappbehat/tests/behat/behat_app.php
crazyserver: I think a discussion is needed here

local-moodleappbehat/tests/behat/behat_app.php
crazyserver: I think right now the declaration of Given, Then, When is something we have to check it works the same. Also I entered the app as doesn't exists previously so nothing changed on the current step.

local-moodleappbehat/tests/behat/behat_app_trait.php
crazyserver: Yeah, I annotated that for a future improvement.

local-moodleappbehat/tests/behat/behat_app_trait.php
crazyserver: Annotated

src/tests/behat/performance.feature
crazyserver: I know, and this is about the race condition you've mentioned before. It's causing the user tours to be shown when page is reloaded.

src/tests/behat/performance.feature
crazyserver: True, my bad!

src/tests/tests.module.ts
crazyserver: I called BehatTestingModule

PR #3282, opened 2022-05-17
MOBILE-4074 configs: Add default zoom level

alfonso-salces:


moodle.config.json
```diff
@@ -77,8 +77,8 @@
         }
     },
     "zoomlevels": {
-        "normal": 100,
-        "low": 110,
+        "low": 100,
+        "normal": 110,
```
dpalou, line 81: I just realized that, since the zoom level setting is stored in the DB by name, this change can have an impact to users who have changed the setting. E.g. imagine a user that changed the setting to the "middle A", the DB (CoreConfig) will have the value "low". After updating, the value "low" is now 100% so the zoom level will change.

IMO we should implement an upgrade step to change the value stored in the DB (if any), and you can also use it to change the word "normal" to "medium". So if the DB has "low" stored it should be changed to "medium" and viceversa.

I was also thinking that, since yesterday we said that the "Low" zoom should always be 100%, maybe we should rename it to "None" because 100% means there is no zoom instead of being a low zoom.

You can take a look to _CoreUpdateManagerProvider.initialize_ to see how to do the upgrade step.

src/types/config.d.ts
```diff
@@ -39,6 +39,7 @@ export interface EnvironmentConfig {
     wsservice: string;
     demo_sites: Record<string, CoreSitesDemoSiteData>;
     zoomlevels: Record<CoreZoomLevel, number>;
+    defaultZoomLevel?: CoreZoomLevel; // Define list of zoom percentages to select in settings.
```
dpalou, line 42: I think this comment could be improved, it makes me think the setting is a "list of zoom percentages" but it's actually a single zoom level.

dpalou (changes requested):

dpalou:
The commit name is wrong (and the PR title too), the issue number is MOBILE-4074.

src/types/config.d.ts
alfonso-salces: Nice catch, I forgot to edit this comment before. Now it's fine!

dpalou (approved):

PR #3295, opened 2022-05-27
Mobile 4069

dpalou:


src/core/singletons/tests/browser.test.ts
```diff
@@ -0,0 +1,68 @@
...
+        expect(CoreBrowser.hasCookie('first-cookie')).toBe(true);
+        expect(CoreBrowser.hasCookie('second-cookie')).toBe(true);
+        expect(CoreBrowser.hasCookie('third-cookie')).toBe(false);
+    });
+
+    it('gets a cookie', () => {
+        document.cookie = 'first-cookie=foo';
+        document.cookie = 'second-cookie=bar';
+
+        expect(CoreBrowser.getCookie('first-cookie')).toEqual('foo');
```
crazyserver, line 32: This is unrelated to the commit but it would be nice to add a setCookie function too.

crazyserver (approved):

src/core/singletons/tests/browser.test.ts
dpalou: I guess it wasn't created because it wasn't needed, I think we'll usually use setDevelopmentSetting instead.

PR #3297, opened 2022-05-30
MOBILE-4081 sites: add logout without require auth again

alfonso-salces:


src/core/services/sites.ts
```diff
@@ -1353,12 +1353,13 @@ export class CoreSitesProvider {
...
      * @return Promise resolved when done.
      */
-    protected async setSiteLoggedOut(siteId: string): Promise<void> {
+    async setSiteLoggedOut(siteId: string, isLoggedOut: boolean = true): Promise<void> {
         const site = await this.getSite(siteId);
 
-        site.setLoggedOut(true);
+        site.setLoggedOut(isLoggedOut);
 
         await this.sitesTable.update({ loggedOut: 1 }, { id: siteId });
```
NoelDeMartin, line 1364: I suppose if `isLoggedOut` is true, this should be updated to `0` instead of `1`.

NoelDeMartin (changes requested):

src/core/services/sites.ts
alfonso-salces: Oh, that's true. Fixed!

PR #3301, opened 2022-05-30
MOBILE-4081 core: Fix package-lock.json for npm 6

NoelDeMartin:
We were previously relying on npm 7 by mistake


dpalou:
Pinging @crazyserver to confirm if there is something in our scripts that need to be changed too. I searched "npm@7" in apps-scripts and I didn't find it, so maybe these changes are enough.

crazyserver:
Nothing has to be changed because we're using v6 on BMA's

PR #3302, opened 2022-05-30
MOBILE-4042 calendar: Swipe navigation in event

NoelDeMartin:


src/addons/calendar/pages/day/day.page.ts
```diff
@@ -816,4 +824,33 @@ class AddonCalendarDaySlidesItemsManagerSource extends CoreSwipeSlidesDynamicIte
...
+        }
+    }
+
+    /**
+     * Remember the list of events in a day to be used in a different context.
+     *
+     * @param day Day containing the events list.
+     */
+    private async rememberEventsList(day: PreloadedDay): Promise<void> {
+        day.eventsUuid = Math.random().toString();
```
dpalou, line 842: I guess the odds are really really low, but I guess it's possible to end up with two identical UUIDs. Maybe we should verify that it doesn't exist already, or create a static method in _AddonCalendarEventsSource_ to use a counter instead of a random number.

src/addons/calendar/pages/day/day.page.ts
```diff
@@ -816,4 +824,33 @@ class AddonCalendarDaySlidesItemsManagerSource extends CoreSwipeSlidesDynamicIte
...
+
+        const source = CoreRoutedItemsManagerSourcesTracker.getOrCreateSource(AddonCalendarEventsSource, [
+            day.eventsUuid,
+            day.filteredEvents,
+        ]);
+
+        this.eventsSources.push(source);
+
+        CoreRoutedItemsManagerSourcesTracker.addReference(source, this);
```
dpalou, line 851: As it's implemented right now, every time the user changes a filter we'll duplicate the sources of all the loaded days. Shouldn't it be better to check if a source already exists for that day and, if it does, just update the events?

E.g. instead of having the _eventsSources_ array, you could store the source in the "day" object to have access to it easily.

dpalou (changes requested):

dpalou (approved):

PR #3303, opened 2022-05-30
MOBILE-3988 core: Remove duplicated promise helper

NoelDeMartin:


src/core/classes/promised-value.ts
```diff
@@ -55,6 +55,13 @@ export class CorePromisedValue<T = unknown> extends CorePromise<T> {
         this.rejectPromise = rejectPromise;
     }
 
+    /**
+     * @deprecated The instance can be directly used as a promise.
```
dpalou, line 59: Please include the version when this was deprecated (4.1) so we can remove it in the future.

src/core/services/utils/utils.ts
```diff
@@ -1325,18 +1326,13 @@ export class CoreUtilsProvider {
     }
 
     /**
-     * Similar to AngularJS $q.defer().
+     * Create a deferred promise that can be resolved or rejected explicitly.
      *
      * @return The deferred promise.
+     * @deprecated Use CorePromisedValue instead.
```
dpalou, line 1332: Please include the version when this was deprecated (4.1) so we can remove it in the future.

dpalou (changes requested):

src/core/classes/promised-value.ts
NoelDeMartin: Done.

PR #3304, opened 2022-06-01
MOBILE-4061 behat: Check tags when launching the app

crazyserver:


local-moodleappbehat/tests/behat/behat_app_helper.php
```diff
@@ -644,18 +647,18 @@ public function skip_restricted_tags_scenarios() {
...
     }
 
     /**
      * Gets if version is incompatible with the @lms_from and @lms_upto tags.
      *
-     * @return bool If scenario has any version incompatible tag.
+     * @return string If scenario has any version incompatible tag, return it.
      */
-    protected function has_version_restrictions() : bool {
+    protected function has_version_restrictions() : string {
```
NoelDeMartin, line 661: Given that this now returns a string, I think it should be renamed to something like "get_restricted_version_tag" or something like that. Also in that case, I guess it should return null in case there isn't any (instead of an empty string).

NoelDeMartin (approved):

NoelDeMartin (approved):

PR #3306, opened 2022-06-02
Mobile 4085 - Improve performance in views with OnPush strategy.

alfonso-salces:


src/addons/storagemanager/pages/course-storage/course-storage.ts
```diff
@@ -402,6 +408,7 @@ export class AddonStorageManagerCourseStoragePage implements OnInit, OnDestroy {
         });
 
         this.deleteModules(modules, section);
+        this.cdRef.markForCheck();
```
NoelDeMartin, line 411: How come this has been added here when other calls to `deleteModules` don't explicitly call `markForCheck` afterwards? It seems like `deleteModules` already calls it intenally, so it shouldn't be necessary to call it here.

src/core/features/course/components/course-format/course-format.ts
```diff
@@ -64,6 +66,7 @@ import { CoreUserTourDirectiveOptions } from '@directives/user-tour';
     selector: 'core-course-format',
     templateUrl: 'course-format.html',
     styleUrls: ['course-format.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class CoreCourseFormatComponent implements OnInit, OnChanges, OnDestroy {
```
NoelDeMartin, line 71: I'm unsure where `markForCheck` should be called, and I'm wondering why it's missing at the end of `doRefresh`.

Looking at this PR, this approach seems very error prone, specially in our codebase because we have very complex components (such as this one). I suppose this is something we'll find during testing, but there are always edge cases we can miss and this type of thing is difficult to cover with automated testing. I don't have that much experience with Angular and I'm not sure how this is done in big codebases. Is it always ad-hoc on a per component basis? Or maybe this issue is alleviated using presentational components because just checking the properties that change is enough?

I don't know, my overall impression is that I don't like having to do this type of thing but at the same time I can't think of a better way. I'm just worried if we start doing this more often, how will we discern when it's necessary to call this function or not. For example, whilst looking at this PR I'm very lost as to why or when change detection is triggered. Maybe adding some comments would help, but I'm not much of a fan either.

In any case, this does improve the performance which is important, so I'm fine with merging this once my suggestions in other comments are addressed. But I have some reservations on how to use this at scale in the codebase, I think we need to have some clear conventions on how to tackle this or it can be a very dangerous source of bugs.

src/addons/storagemanager/pages/course-storage/course-storage.ts
```diff
@@ -71,14 +72,15 @@ export class AddonStorageManagerCourseStoragePage implements OnInit, OnDestroy {
     protected isDestroyed = false;
     protected isGuest = false;
 
-    constructor(protected elementRef: ElementRef) {
+    constructor(protected elementRef: ElementRef, private readonly cdRef: ChangeDetectorRef) {
```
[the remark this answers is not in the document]
alfonso-salces: I think your name is better than mine in this case, im agreed with you in abbreviations too, so I changed name in both files and know these properties are protected.

Thank you for your review!

src/addons/storagemanager/pages/course-storage/course-storage.ts
alfonso-salces: It's necessary because deleteModules is an async method which manipulate properties used into the view (totalSize, module.totalSize...), so this will execute before deleteModules starts and then when all promises finish will execute again for update sizes.

alfonso-salces: I checked this again and you are right, there is no reason to call markForCheck after deleteModules.

src/core/features/course/components/course-format/course-format.ts
NoelDeMartin: I'm integrating this for now, but we'll need to talk about this if we start using this strategy more often.

NoelDeMartin (approved):

PR #3307, opened 2022-06-02
MOBILE-4081: Improve notifications source and fix behat steps

NoelDeMartin:


dpalou (changes requested):

src/addons/notifications/services/notifications.ts
```diff
@@ -165,6 +165,7 @@ export class AddonNotificationsProvider {
      * @param notifications Current list of loaded notifications. It's used to calculate the offset.
      * @param options Other options.
      * @return Promise resolved with notifications and if can load more.
+     * @deprecated since 4.1. Use getNotificationsWithStatus instead.
```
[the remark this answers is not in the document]
NoelDeMartin: Thanks for pointing this out, but I think I'll leave as deprecated just in case.

src/addons/notifications/classes/notifications-source.ts
```diff
@@ -28,20 +51,74 @@ export class AddonsNotificationsNotificationsSource extends CoreRoutedItemsManag
...
-        // TODO this should be refactored to avoid using the existing items.
-        const { notifications, canLoadMore } = await AddonNotifications.getNotifications(page === 0 ? [] : this.getItems() ?? []);
+        const site = await CoreSites.getSite();
+        const results = await Promise.all(
+            site.isVersionGreaterEqualThan('4.0')
+                ? [this.loadNotificationsPage(page, AddonNotificationsGetReadType.BOTH)]
+                : [
+                    this.loadNotificationsPage(page, AddonNotificationsGetReadType.UNREAD),
+                    this.loadNotificationsPage(page, AddonNotificationsGetReadType.READ),
+                ],
```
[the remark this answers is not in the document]
NoelDeMartin: Good catch, I did this looking at the [feedback-attempts-source](https://github.com/moodlehq/moodleapp/blob/integration/src/addons/mod/feedback/classes/feedback-attempts-source.ts#L137) which has a similar issue (instead of read/unread, the endpoint returns anonymous/non anonymous results). However, in that case the component using the source already takes this into account and doesn't show the bottom part of the list until the top is completely loaded.

I have refactored the code and it should be working now, please take another look.

src/addons/notifications/classes/legacy-notifications-source.ts
```diff
@@ -0,0 +1,62 @@
...
+            const pageResults = await this.loadNotifications(AddonNotificationsGetReadType.UNREAD, page * pageLength);
+
+            pageUnreadCount = items.length;
+            items = items.concat(pageResults.notifications);
+            hasMoreItems = pageResults.hasMoreNotifications;
+        }
+
+        // If all unread notifications have been fetched, load read notifications first.
+        if (totalUnread() < (page + 1) * pageLength) {
+            const offset = page * pageLength - totalUnread();
```
dpalou, line 48: I think this _offset_ is wrong, it can be negative. Imagine I have 5 unread notifications and 5 read. When loading page 0, first we load the unread ones, we add 5 items and the _totals_ for unread is 5. So the offset is now: 0 * 20 - 5 = -5. I'm not sure if sending a negative value to the WS can cause problems or it will work as 0, but this offset is also used in _loadNotifications_ to calculate the _totals_ and that means we'll have a wrong value. Because the _totals_ for read will be:

offset + notifications.length = -5 + 5 = 0

but it should actually be 5. I think that forcing it to be 0 instead of negative will be enough to make it work fine.

src/addons/notifications/classes/legacy-notifications-source.ts
```diff
@@ -0,0 +1,62 @@
...
+        let hasMoreItems = false;
+        let pageUnreadCount = 0;
+        const pageLength = this.getPageLength();
+        const totalUnread = () => this.totals[AddonNotificationsGetReadType.UNREAD] ?? Number.MAX_VALUE;
+
+        // Load unread notifications first.
+        if (totalUnread() > page * pageLength) {
+            const pageResults = await this.loadNotifications(AddonNotificationsGetReadType.UNREAD, page * pageLength);
+
+            pageUnreadCount = items.length;
```
dpalou, line 41: I think this value will always be 0, right? Because it's calculated before modifying the _items_ array. Maybe you can get rid of this variable and just use _items.length_ when fetching the read notifications?

dpalou (changes requested):

src/addons/notifications/classes/legacy-notifications-source.ts
NoelDeMartin: Yeah good catch, I didn't think this properly. This only happens when reading a page that has notifications from both read and unread sources, so it could be fixed like this as well (and maybe it's more semantically correct):

`const offset = page * pageLength - totalUnread() + pageUnreadCount;`

But I agree it's safer to just do `Math.max(0, ...)`, and it will produce the same results in the end so that's what I've done.

src/addons/notifications/classes/legacy-notifications-source.ts
NoelDeMartin: Yeah you're right, this was also wrong. Instead of removing the variable, I've fixed it to read from the `pageResults` variable. It's true that your proposal of using `items.length` later on would work, but I think this is more readable.

dpalou (approved):

PR #3311, opened 2022-06-09
Mobile 4061

crazyserver:


src/addons/mod/data/tests/behat/entries.feature
```diff
@@ -0,0 +1,191 @@
...
+      | Field name | URL |
+      | Field description | URL link |
+    And I add a "Text input" field to "Web links" database and I fill the form with:
+      | Field name | Description |
+      | Field description | Link description |
+    And I log out
+
+  Scenario: Create entry
+    Given I entered the data activity "Web links" on course "Course 1" as "student1" in the app
+    And I should find "No entries in database" in the app
```
dpalou, line 37: Some connectors are wrong, e.g. this one should be "Then" instead of "And". There are also places where we have a "Then" and, after all the checks, we use "And" instead of "When". It happens in several places of this file and also other tests.

src/addons/mod/data/tests/behat/entries.feature
```diff
@@ -0,0 +1,191 @@
...
+    And I should find "Moodle Cloud" in the app
+    And I press "Delete" in the app
+    And I should find "Are you sure you want to delete this entry?" in the app
+    And I press "Cancel" in the app
+    And I should find "Moodle Cloud" in the app
+    And I press "Delete" in the app
+    And I should find "Are you sure you want to delete this entry?" in the app
+    And I press "Delete" in the app
+    And I should not find "Moodle Cloud" in the app
+    And I press "Add entries" in the app
```
dpalou, line 114: Can you add a comment explaining that now you'll test the same in single view? At first I thought you were replicating the same test.

src/addons/mod/glossary/tests/behat/basic_usage.feature
```diff
@@ -0,0 +1,254 @@
...
+    Then the header should be "Test glossary" in the app
+    And I should find "Eggplant" in the app
+    And I should find "Cucumber" in the app
+    And I should find "Potato" in the app
+
+    When I press "Potato" in the app
+    Then I should find "Potato" in the app
+    And I should find "To make chips" in the app
+
+  Scenario: Change filters (include search)
```
dpalou, line 49: This test might be duplicated, the navigation tests inside glossary are already checking search.

local-moodleappbehat/tests/behat/behat_app.php
```diff
@@ -195,7 +195,7 @@ public function i_load_more_items_in_the_app(bool $not = false) {
     public function i_swipe_in_the_app(string $direction) {
         $method = 'swipe' . ucwords($direction);
 
-        $this->evaluate_script("window.behat.getAngularInstance('ion-content', 'CoreSwipeNavigationDirective').$method()");
+        $this->execute_script("window.behat.getAngularInstance('ion-content', 'CoreSwipeNavigationDirective').$method()");
```
dpalou, line 198: As discussed in the chat, maybe we can have everything with _evaluate_script_ even if we don't need the return, that way all the calls will be the same. Also, if we need the return in the future we won't need to change the function name. But I don't have a strong opinion on this :P

src/testing/services/behat-runtime.ts
```diff
@@ -46,6 +55,10 @@ export class TestsBehatRuntime {
             scrollTo: TestsBehatRuntime.scrollTo,
             setField: TestsBehatRuntime.setField,
             handleCustomURL: TestsBehatRuntime.handleCustomURL,
+            setForceOffline: TestsBehatRuntime.setForceOffline,
```
dpalou, line 58: Maybe it would be better to just expose the services inside _window.behat_ (unless we need some extra code), that way we don't need to change this file if we need another function in the service. E.g. we could do something like:

_network: CoreNetwork.instance_,

That way you don't need to create a _setForceOffline_ function that only calls another function.

src/core/initializers/prepare-automated-tests.ts
```diff
@@ -23,12 +23,30 @@ import { Application, NgZone } from '@singletons';
 
 type AutomatedTestsWindow = Window & {
     appRef?: ApplicationRef;
-    appProvider?: CoreAppProvider;
     dbProvider?: CoreDbProvider;
+    /**
+     * @depreacated on 4.1 Use TestsBehatRuntime instead.
+     */
+    appProvider?: CoreAppProvider;
```
dpalou, line 30: AFAIK these values are only used in our internal functions (the steps implementation), the public behat API (the steps declaration) isn't affected by this so I think we can just delete them instead of deprecating them.

dpalou (changes requested):

src/core/initializers/prepare-automated-tests.ts
crazyserver: Removed

dpalou (approved):

PR #3315, opened 2022-06-11
MOBILE-4092 courses: Fix access to hidden courses

tasosb:


src/theme/theme.base.scss
```diff
@@ -1399,6 +1399,11 @@ html.ios div.fake-ion-item {
 
 }
 
+// Simulate item[disabled] but keep pointer-events for hidden courses.
```
crazyserver, line 1402: Can you move this css to the specific file on src/core/features/courses/components/course-progress/
and src/core/features/courses/components/course-list-item/

It will be duplicated but course-progress is deprecated and will be removed soon.

crazyserver (changes requested):
Thanks for providing this patch! Nice catch! It would be great if you can also add behat testing to this issue (I will do it if you are not able to). 

THanks!

PR #3317, opened 2022-06-13
MOBILE-4041 completion: Fix styles on non-synced completion

crazyserver:


crazyserver:
Automatic testing is not required because it's an style fix.

dpalou (approved):

PR #3318, opened 2022-06-13
MOBILE-4147 core,folder: Restore access to inline folders 

tasosb:
This PR restores access to inline folders that were previously inaccessible.
Inline folders are now displayed like the regular folders in the mobile app.
Changes:

1. New core feature that can be set to true if the mobile app can produce a view for the module requested but its URL isn't passed from the WebService.
2. If (1) is enabled and no URL is assigned to the module, the "Open In Browser" button is hidden in the module summary.
3. Enable feature (1) for the folder addon, and change its description so the inline folder passed in the description from the WebService is not shown.


NoelDeMartin:
It seems like this issue is still pending, and it's actually tracked by [MOBILE-4147](https://tracker.moodle.org/browse/MOBILE-4147) so I just updated the title and converted it to draft.

dpalou:
This branch was closed automatically because the v4.1.x branch was deleted. IMO there's no need to rebase the code to point to the main branch because we still need to dedicate some time to do a UX research about this issue, and we don't know when will we be able to do this.

Once we've been able to dedicate some time to [MOBILE-4147](https://tracker.moodle.org/browse/MOBILE-4147) we can decide what to do with the code of this PR.

PR #3320, opened 2022-06-15
Mobile 4081

crazyserver:


src/core/singletons/index.ts
```diff
@@ -190,7 +190,7 @@ export const NativeHttp = makeSingleton(HTTP);
 /**
  * @deprecated on 4.1 use CoreNetwork instead.
  */
-export const Network = makeSingleton(NetworkService);
+export const Network = CoreNetwork;
```
NoelDeMartin, line 193: I don't think this is necessary, if any part of the code wants to use this API it's possible to just import from `@services/network`. So Network can be removed entirely from this file (it doesn't seem like it's used anywhere anyways).

src/core/services/network.ts
```diff
@@ -39,7 +37,7 @@ export class CoreNetworkService extends NetworkService {
         this.checkOnline();
 
         if (CorePlatform.isMobile()) {
-            Network.onChange().subscribe(() => {
+            this.onChange().subscribe(() => {
```
NoelDeMartin, line 40: Technically not part of this PR, but one comment I have is that we should avoid doing things like registering listeners or subscriptions in constructors (or said differently, we should avoid side-effects in constructors). This could be done on an initializer instead, like many services are already doing in the `initialize-services.ts` file.

src/core/features/emulator/emulator.module.ts
```diff
@@ -142,17 +141,13 @@ import { CoreNetworkService } from '@services/network';
...
-            useFactory: (): Media => new Media(),
-        },
+        Media,
         {
             provide: MediaCapture,
             deps: [Platform],
             useFactory: (platform: Platform): MediaCapture => platform.is('cordova') ? new MediaCapture() : new MediaCaptureMock(),
         },
-        CoreNetworkService,
+        Network,
```
NoelDeMartin, line 150: I'm not sure why it's necessary to add this here (and I have the same question for the rest: `Push`,  `QRScanner`, etc.). The ones that are overriding an instance make sense, but I'm not sure about the rest. Maybe this was needed in Ionic 3 and we didn't remove them on the update?

NoelDeMartin (changes requested):

src/core/features/emulator/emulator.module.ts
crazyserver: Hmmm So probably we'd open an issue about that?

NoelDeMartin (approved):

PR #3325, opened 2022-06-17
MOBILE-4081 login: Load reconnect and credentials pages from service

alfonso-salces:


src/core/features/login/services/login-helper.ts
```diff
@@ -1435,6 +1437,24 @@ export class CoreLoginHelperProvider {
         }
     }
 
+    /**
+     * Get reconnect page route module.
+     *
+     * @returns Reconnect page route module.
+     */
+    async getReconnectRouteModule(): Promise<CoreLoginReconnectPageModule> {
+        return import('@features/login/pages/reconnect/reconnect.module').then(m => m.CoreLoginReconnectPageModule);
```
NoelDeMartin, line 1446: I think it'd be better to declare this as returning `Promise<unknown>` or something else, there a couple of reasons why I think that's better.

First, we should avoid importing module classes in this way because we already have many issues with circular dependencies and lazy modules not working as well as they should (I talked about it [in this issue](https://tracker.moodle.org/browse/MOBILE-1804?focusedCommentId=887454&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-887454)). So if we were to import the module, I think it should be a type import instead.

But secondly, and maybe more important, with this signature it would be difficult to override this method in subclasses (because they should return a subclass of the returned module). I think just returning `Promise<unknown>` should work because modules don't really have a parent class, and looking at the Angular types they use `any` so it should be equivalent.

NoelDeMartin (changes requested):

src/core/features/login/services/login-helper.ts
alfonso-salces: Hi, Im agreed with you, so now the return type is ` Promise<unknown>` 
Thanks for your review!

NoelDeMartin (approved):

PR #3330, opened 2022-06-22
MOBILE-4038 timeline: Refactor using OnPush

NoelDeMartin:


NoelDeMartin:
I converted this to a draft because I realized I still need to implement some things (in particular, refresh is not working). But I don't think the approach will change drastically, so if anyone wants to get started with the review go ahead.

NoelDeMartin:
Ok, it should be fine now, let me know what you think.

alfonso-salces (changes requested):

src/addons/block/timeline/classes/section.ts
```diff
@@ -0,0 +1,200 @@
...
+
+        const lastEventId = this.data$.value.lastEventId;
+        const { events, canLoadMore } = this.course
+            ? await AddonBlockTimeline.getActionEventsByCourse(this.course.id, lastEventId, this.search ?? '')
+            : await AddonBlockTimeline.getActionEventsByTimesort(lastEventId, this.search ?? '');
+
+        this.data$.next({
+            events: this.data$.value.events.concat(this.reduceEvents(events, this.overdue, this.dateRange)),
+            lastEventId: canLoadMore,
+            canLoadMore: typeof canLoadMore !== 'undefined',
```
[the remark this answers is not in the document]
NoelDeMartin: Done.

src/core/utils/rxjs.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+ * Observable operator that waits for a promise to resolve before emitting the result.
+ *
+ * @returns Operator.
+ */
+export function resolved<T>(): OperatorFunction<Promise<T>, T> {
+    return source => new Observable(subscriber => {
+        source.subscribe(promise => promise.then(value => subscriber.next(value)));
```
dpalou, line 39: I think you should do:

`return source.subscribe...`

Otherwise I think you will never unsubscribe from the `source`. If I understood it right, if you return the subscription then when we `unsubscribe` from the returned `Observable` it will also unsubscribe from `source`.

src/core/utils/rxjs.ts
```diff
@@ -0,0 +1,56 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { FormControl } from '@angular/forms';
+import { Observable, OperatorFunction } from 'rxjs';
+import { filter } from 'rxjs/operators';
+
```
dpalou, line 18: It would be nice to have unit tests for these utility functions. This might be useful to you:

https://github.com/moodlehq/moodleapp/commit/b9deedc3b2cd77eb505aef8d9f5c779398c7e920#diff-37edbc3ec48e47dbd9ca306be02050c9d0760b2010b0e5750d67f11ab1d442dfR19

Your patch will probably be integrated first, so when it is I'll move my functions and tests to your rxjs files.

src/core/utils/rxjs.ts
```diff
@@ -0,0 +1,56 @@
...
+ * on subscription.
+ *
+ * @param getValue Callback to calculate the starting value.
+ * @returns Operator.
+ */
+export function startWithAsync<T>(getValue: () => T): OperatorFunction<T, T> {
+    return source => new Observable(subscriber => {
+        subscriber.next(getValue());
+
+        source.subscribe(value => subscriber.next(value));
```
dpalou, line 54: As above, I think you should `return` the subscription.

src/core/utils/rxjs.ts
```diff
@@ -0,0 +1,56 @@
...
+}
+
+/**
+ * Same as the built-in startWith operator, but evaluates the starting value for each subscriber
+ * on subscription.
+ *
+ * @param getValue Callback to calculate the starting value.
+ * @returns Operator.
+ */
+export function startWithAsync<T>(getValue: () => T): OperatorFunction<T, T> {
```
dpalou, line 50: This name confused me at first. Since it contains the word "async" I thought it would have something to do with promises, but it doesn't. I was trying to find a better name but I can't think of any that convinces me, e.g. `startWithCallback`, `startWithDelayed`, etc. Maybe we can discuss it internally with the team to see if we find a better name.

dpalou (changes requested):

src/testing/services/behat-runtime.ts
```diff
@@ -349,6 +352,30 @@ export class TestsBehatRuntime {
         }
     }
 
+    static async pullToRefresh(): Promise<string> {
+        this.log('Action - pullToRefresh');
+
+        try {
+            const dashboard = this.getAngularInstance<CoreCoursesDashboardPage>(
```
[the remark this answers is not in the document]
NoelDeMartin: Initially I tried to get a reference to the refresher directly, that way we could make this more generic. But the `IonRefresher` component doesn't have any public method to trigger the PTR, so I couldn't find a way to do it like that :/.

So the only way I found is to call the method in the page component directly. Which I don't like, but I couldn't find a better way. One issue with this is that each page could have different method names so we'll have to implement PTR for each individual page... I think we can take another look at this in the future, but after spending some time trying to do it properly I decided to leave it like this for now.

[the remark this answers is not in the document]
crazyserver: Ok then, could you please add JS docs and clarify this is something TODO?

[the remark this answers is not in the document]
NoelDeMartin: Done

src/core/utils/rxjs.ts
NoelDeMartin: Done

src/core/utils/rxjs.ts
NoelDeMartin: After talking we decided to call it `startWithOnSubscribed`.

crazyserver:
I has my OK ;-)

crazyserver (approved):

dpalou (approved):

alfonso-salces (approved):

PR #3335, opened 2022-06-30
MOBILE-4027 h5pactivity: Don't show modal if user finish activity 10 secs ago

alfonso-salces:


src/addons/mod/h5pactivity/pages/index/index.ts
```diff
@@ -40,13 +42,25 @@ export class AddonModH5PActivityIndexPage extends CoreCourseModuleMainActivityPa
...
+                return true;
+            } catch {
+                return false;
+            }
         }
+
+        return true;
+    }
+
+    setCanleaveSafely(isDone: boolean): void {
```
dpalou, line 58: Please add JSDoc for this function.

src/addons/mod/h5pactivity/pages/index/index.ts
```diff
@@ -40,13 +42,25 @@ export class AddonModH5PActivityIndexPage extends CoreCourseModuleMainActivityPa
...
+        this.canLeaveSafely = isDone;
+        // When user finish an activity, he have 10 seconds to leave safely (without show alert).
+        setTimeout(() => {
```
dpalou, line 61: You might want to cancel a previous timeout (if any) before setting a new one. It's really unlikely that a user will finish an attempt, retry and finish it again before 10 secs, but I still think it's more correct to cancel the previous timeout.

dpalou (changes requested):

dpalou (approved):

PR #3336, opened 2022-07-04
MOBILE-4108 core: Don't call autologin WS for external URLs

dpalou:


src/core/directives/link.ts
```diff
@@ -42,10 +42,9 @@ export class CoreLinkDirective implements OnInit {
     @Input() capture?: boolean | string; // If the link needs to be captured by the app.
     @Input() inApp?: boolean | string; // True to open in embedded browser, false to open in system browser.
     /* Whether the link should be opened with auto-login. Accepts the following values:
-       "yes" -> Always auto-login.
-       "no" -> Never auto-login.
-       "check" -> Auto-login only if it points to the current site. Default value. */
-    @Input() autoLogin = 'check';
+       "yes" -> Auto-login only if it points to the current site. Default value.
+       "no" -> Never auto-login.  */
+    @Input() autoLogin = 'yes';
```
NoelDeMartin, line 47: Given that this is yes/no now, could we refactor it to a boolean instead? If it's for backwards compatibility, I think we should be able to accept both a string (yes/no) and a boolean, and document that using a string is deprecated.

src/core/directives/link.ts
```diff
@@ -184,10 +183,20 @@ export class CoreLinkDirective implements OnInit {
...
-        // It's an external link, we will open with browser. Check if we need to auto-login.
+        // Priority order is: core-link inApp attribute > forceOpenLinksIn setting > data-open-in HTML attribute.
+        let openInApp = this.inApp;
+        if (this.inApp === undefined) {
+            if (CoreConstants.CONFIG.forceOpenLinksIn == 'browser') {
+                openInApp = false;
+            } else if (CoreConstants.CONFIG.forceOpenLinksIn == 'app' || openIn == 'app') {
+                openInApp = true;
+            }
+        }
```
NoelDeMartin, line 194:
```suggestion
        const openInApp = this.inApp ?? (
            CoreConstants.CONFIG.forceOpenLinksIn !== 'browser' &&
            (CoreConstants.CONFIG.forceOpenLinksIn === 'app' || openIn === 'app')
        );
```

What do you think of doing this instead? I always prefer to avoid mutable variables if possible, and I always frown upon using if/else to set booleans.

Also, looking at the `inApp` attribute I noticed that it's declared as `boolean | string`. I think that's a bit misleading, because in `ngOnInit` we're casting it to a boolean, so I think we should use a derivated attribute instead that is typed as a `boolean` only. And additional issue is that the value is not casted if the input changes, but I guess we have that problem in many places throughout the app.

Of course, the best approach would be to just accept boolean properties, and move the responsibility of casting strings to booleans outside of this component. But I suppose that may require a lot of refactors and break the plugins API.

NoelDeMartin (commented):
I have some suggestions, but overall I think it's ready for merging.

src/core/directives/link.ts
dpalou: Done, now it prefers boolean but it also supports yes/no as it used to. I'm also supporting "true"/"false" (as strings) for coherence, I'll talk about this in the other comment.
I also changed this in the iframe.ts page.

src/core/directives/link.ts
dpalou: I changed how the variable is calculated, as you suggested. Now I'm also doing the cast to boolean in this expression to avoid having problems if the value changes, now the casted value isn't stored in the input.

IMO accepting only booleans can cause problems for plugin developers, most of them don't know Angular and they don't know the difference between [inApp]="true" and inApp="true" (we've found people with this type of problems in the past). I think that with my changes we will no longer have problems in this directive.

NoelDeMartin (approved):

PR #3339, opened 2022-07-06
Mobile 4086

dpalou:


src/core/services/database/sites.ts
```diff
@@ -15,14 +15,29 @@
...
 /**
  * Database variables for CoreSites service.
  */
 export const SITES_TABLE_NAME = 'sites_2';
 export const SCHEMA_VERSIONS_TABLE_NAME = 'schema_versions';
 
+/**
+ * Database variables for CoreSite class.
+ */
+export enum WSCacheTablesSuffixes {
```
NoelDeMartin, line 28: I think this enum should be renamed to something else, because although it is true that it's used as a suffix for WSCacheTables that is just *how it's used*, not *what it is*. I guess the difference is subtle, and maybe it doesn't matter, but I think it makes more sense.

Now, as to what to call it... I'm not sure :/. Maybe `WSTypes`, `WSGroups`, etc? I'm not sure, I guess one way to find a name is to answer the question: How are WS classified within these groups?

src/core/services/database/sites.ts
```diff
@@ -15,14 +15,29 @@
...
+    CORE = 'core',
+    BLOCK = 'block',
+    MOD = 'mod',
+    TOOL = 'tool',
+    OTHER = 'other',
+}
+export const WS_CACHE_OLD_TABLE = 'wscache_2';
+export const WS_CACHE_TABLES_COMMON = 'wscache_';
```
NoelDeMartin, line 36:
```suggestion
export const WS_CACHE_TABLES_PREFIX = 'wscache_';
```
What do you think of this? It may seem like I am contradicting my previous comment, but in this case I think this variable really *is* the table prefix, I don't think it carries any other meaning.

src/core/classes/site.ts
```diff
@@ -111,7 +115,7 @@ export class CoreSite {
     // Rest of variables.
     protected logger: CoreLogger;
     protected db?: SQLiteDB;
-    protected cacheTable: AsyncInstance<CoreDatabaseTable<CoreSiteWSCacheRecord>>;
+    protected cacheTables: Record<string, AsyncInstance<CoreDatabaseTable<CoreSiteWSCacheRecord>>> = {};
```
NoelDeMartin, line 118:
```suggestion
    protected cacheTables: Record<WSCacheTablesSuffixes, AsyncInstance<CoreDatabaseTable<CoreSiteWSCacheRecord>>> = {};
```
I don't think in practice this will have any relevant impact, because we're initializing the tables using the enum already. But I think it's good practice to do this whenever possible.

If you accept this change, you won't be able to initialize the object like this (which to be honest, is good because we don't want an empty object to be a valid value for this property). But since you're already initializing the object in the constructor, it should be fine. It's just that instead of populating the object with a forEach, you'll have to implement it with a reduce and assign the variable.

I'm thinking you may have some issues declaring types for the reduce though, because until all the values have been added it won't be complete. I guess we could do this "the right way" using (or implementing) some helpers to map enums into values or something, but I think it's ok if you just cast it manually.

src/core/classes/site.ts
```diff
@@ -1022,6 +1029,25 @@ export class CoreSite {
...
+            return WSCacheTablesSuffixes.MOD;
+        } else if (name.startsWith('tool_')) {
+            return WSCacheTablesSuffixes.TOOL;
+        } else if (name.startsWith('block_') || name.startsWith('core_block_')) {
+            return WSCacheTablesSuffixes.BLOCK;
+        } else if (name.startsWith('core_')) {
+            return WSCacheTablesSuffixes.CORE;
+        } else {
+            return WSCacheTablesSuffixes.OTHER;
+        }
```
NoelDeMartin, line 1048:
```suggestion
        if (name.startsWith('mod_')) {
            return WSCacheTablesSuffixes.MOD;
        } 
        
        if (name.startsWith('tool_')) {
            return WSCacheTablesSuffixes.TOOL;
        }
        
        if (name.startsWith('block_') || name.startsWith('core_block_')) {
            return WSCacheTablesSuffixes.BLOCK;
        }
        
        if (name.startsWith('core_')) {
            return WSCacheTablesSuffixes.CORE;
        }
        
        return WSCacheTablesSuffixes.OTHER;
```

I think this improves readability following a couple of rules of thumb. Although to be honest, it doesn't change that much so ignore it if you want.

src/core/classes/site.ts
```diff
@@ -1183,7 +1213,7 @@ export class CoreSite {
 
         this.logger.debug('Invalidate cache for key: ' + key);
 
-        await this.cacheTable.update({ expirationTime: 0 }, { key });
+        await Promise.all(Object.values(this.cacheTables).map(table => table.update({ expirationTime: 0 }, { key })));
```
NoelDeMartin, line 1216: I wonder if this operation and similar ones are executed too often, because in that case this could be detrimental to performance. We're running N queries instead of just one, so that ain't good.

What do you think of adding the table type at the start of the key? I know we already talked about this internally and you mentioned the keys were hashed, but maybe we should think more about this if that's creating so many issues. Maybe we can just have `{prefix}_{hash}`? That way, we'll also be able to tell if it's worth looking into the old cache table or not.

src/core/services/database/sites.ts
```diff
@@ -152,19 +166,7 @@ export const SITE_SCHEMA: CoreSiteSchema = {
             ],
             primaryKeys: ['component', 'id'],
         },
-    ],
-    async migrate(db: SQLiteDB, oldVersion: number): Promise<void> {
```
NoelDeMartin, line 156: How come this has been deleted? I guess it's very unlikely that someone will be updating the app from <4.0 to 4.1, but I'm not sure if we have an official policy of not supporting that. If we do, I guess we could just remove all migrations before 4.0, not only this one in this particular issue.

NoelDeMartin (changes requested):
I think it's ok overall, but I have some doubts about a couple of parts. And suggestions about naming, as always :sweat_smile:

src/core/services/database/sites.ts
dpalou: I changed it to WSGroups, now I'm no longer using the word suffix in the patch.

src/core/classes/site.ts
dpalou: Changed!

src/core/classes/site.ts
dpalou: Some of the functions affected by this change aren't used often, the ones that are used more often are the "invalidate" ones (like this one). I already told Juan to make tests using PTR (that's when we invalidate things) to see how this affects the results, if we see a negative impact we can consider possible ways to mitigate it.

The problem about modifying the data stored to include the group is that it doesn't work for data already stored because we don't know the WS that the cached data belongs to.

src/core/services/database/sites.ts
dpalou: I deleted this because it was introduced in 3.9.3, so this would only cause problems to people upgrading from 3.9.2 to 4.1 (that's 2 years without updating the app, the cached data is probably no longer valid lol).

PR #3343, opened 2022-07-11
Mobile 4031

dpalou:


src/core/services/utils/dom.ts
```diff
@@ -1639,12 +1635,15 @@ export class CoreDomUtilsProvider {
...
      */
-    async showToastWithOptions(options: ToastOptions): Promise<HTMLIonToastElement> {
-        // Set some default values.
-        options.duration = options.duration ?? 2000;
-        options.position = options.position ?? 'bottom';
+    async showToastWithOptions(options: ShowToastOptions): Promise<HTMLIonToastElement> {
+        // Convert some values and set default values.
+        const toastOptions: ToastOptions = {
+            ...options,
+            duration: CoreConstants.CONFIG.toastDurations[options.duration] ?? options.duration ?? 2000,
```
crazyserver, line 1642: 2000 should be ToastDuration.SHORT

crazyserver (changes requested):

src/core/services/utils/dom.ts
dpalou: Hi Pau, ToastDuration.SHORT is not a number, is "short". This is a fallback in case we cannot find the corresponding number in the config, it shouldn't happen.

PR #3348, opened 2022-07-14
MOBILE-4081 courses: Load 'my.module' from service

alfonso-salces:
This solve black ion-toolbar after use QR scanner too.

### DIFF
--- a/src/core/features/courses/services/courses-helper.ts
+++ b/src/core/features/courses/services/courses-helper.ts
@@ -339,6 +339,15 @@ export class CoreCoursesHelperProvider {
         return startDate > Date.now();
     }
 
+    /**
+     * Retrieves my courses page module.
+     *
+     * @returns My courses page module.
+     */
+    async getMyRouteModule(): Promise<unknown> {
+        return import('../pages/my/my.module').then(m => m.CoreCoursesMyCoursesPageModule);
+    }
+
 }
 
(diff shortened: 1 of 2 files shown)


src/core/features/courses/services/courses-helper.ts
```diff
@@ -339,6 +339,15 @@ export class CoreCoursesHelperProvider {
         return startDate > Date.now();
     }
 
+    /**
+     * Import my courses page module and returns it.
```
NoelDeMartin, line 343: This is too literal and exposes implementation details (what if I don't want to import the module inside this function?). This is just a simple getter (albeit asynchronous), so I think it's fine to just say something like "Retrieve my courses page module" or whatever we do for getters (we don't have a standard format as far as I know, but maybe we should).

src/core/services/utils/utils.ts
```diff
@@ -1702,6 +1702,10 @@ export class CoreUtilsProvider {
         }
 
         delete this.qrScanData;
+        // Clear dirty style after hide QR scanner.
+        setTimeout(() => {
+            document.body.removeAttribute('style');
```
NoelDeMartin, line 1707: Looking at the code I'm not sure what this is supposed to be clearing, and whatever it is I think it should be implemented differently because adding inline styles to the body element is too heavy handed, and could cause conflicts with other parts of the code.

NoelDeMartin (approved):

PR #3350, opened 2022-07-26
MOBILE-4113 geolocation: Fix screen locked before allowing

dpalou:


src/core/services/geolocation.ts
```diff
@@ -133,6 +135,30 @@ export class CoreGeolocationProvider {
         return CoreUtils.promiseWorks(Diagnostic.getLocationAuthorizationStatus());
     }
 
+    /**
+     * Request and return the location authorization status for the application.
+     *
+     * @return Promise resolved with authorization status.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    protected requestLocationAuthorization(): Promise<any> {
```
NoelDeMartin, line 144: Given that we're not using the result for anything, wouldn't it be better to just declare this as `Promise<void>`? That way, we don't have to disable eslint rules.

Also, I have to say I'm not much of a fan of saying "Promise resolved with..." in the documentation for async methods. At this point, I think promises and async functions are enough of a language construct that it's not necessary to mention it explicitly in the docs. For example, if you have a method `user.getAge(): number` you can just say that it returns "A user's age", no need to say "A user's age number".

src/core/services/geolocation.ts
```diff
@@ -133,6 +135,30 @@ export class CoreGeolocationProvider {
...
+        // In iOS, the modal disappears when the screen is locked and the promise never ends. Treat that case.
+        return new Promise((resolve, reject) => {
+            const pauseSubscription = CorePlatform.pause.subscribe(() => {
+                pauseSubscription?.unsubscribe();
+                reject(new CoreSilentError()); // Don't display an error, just finish the process.
+            });
+
+            Diagnostic.requestLocationAuthorization().then(resolve, reject).finally(() => {
+                pauseSubscription.unsubscribe();
+            });
```
NoelDeMartin, line 158:
```suggestion
            // Don't display an error, just finish the process.
            const unsubscribe = CoreSubscriptions.once(CorePlatform.pause, () => reject(new CoreSilentError()));

            Diagnostic.requestLocationAuthorization().then(resolve, reject).finally(() => unsubscribe());
```

I recently found a similar use-case where I wanted to use `CoreSubscriptions.once`, but couldn't because the subscription could be cancelled elsewhere. What do you think of returning an unsubscription callback as well, to improve this type of scenario?

dpalou: Sounds good, done!

NoelDeMartin (approved):

PR #3354, opened 2022-07-29
Mobile 4034

dpalou:


src/core/features/login/pages/change-password/change-password.ts
```diff
@@ -75,4 +81,38 @@ export class CoreLoginChangePasswordPage {
...
+            delete this.urlLoadedObserver;
+            delete this.browserClosedObserver;
+        });
+    }
+
+    /**
+     * @inheritdoc
+     */
+    ngOnDestroy(): void {
+        // The page shouldn't be destroyed if IAB is opened, but stop listening just in case.
```
NoelDeMartin, line 113: I think this comment is not necessary, and it can be removed from the complete profile page as well. It doesn't matter why the page is being destroyed, whenever it happens observers should be cleared to avoid leaks.

src/app/app.component.ts
```diff
@@ -110,7 +111,7 @@ export class AppComponent implements OnInit, AfterViewInit {
             CoreLoginHelper.passwordChangeForced(data.siteId!);
         });
         CoreEvents.on(CoreEvents.USER_NOT_FULLY_SETUP, (data) => {
```
NoelDeMartin, line 113: I realize this is not part of this task, but I think it'd be a good opportunity to move this type of things into more appropriate locations. For example, in this case, I think this would be better in the login module initializer.

src/app/app.component.ts
```diff
@@ -110,7 +111,7 @@ export class AppComponent implements OnInit, AfterViewInit {
...
-            CoreLoginHelper.openInAppForEdit(data.siteId!, '/user/edit.php', 'core.usernotfullysetup');
+            CoreUserHelper.openCompleteProfile(data.siteId!);
```
NoelDeMartin, line 114: I'm not 100% sure of what happened before this was called, but reading the issue description in jira it seems like an error has been shown saying something like "User not fully set-up", right?.

If that's the case, since now we're showing a dedicated page before opening the browser, I think we can omit that error message and convey any information needed in the page.

src/core/features/user/services/user-helper.ts
```diff
@@ -60,6 +62,25 @@ export class CoreUserHelperProvider {
...
+     *
+     * @param siteId The site ID.
+     */
+    async openCompleteProfile(siteId: string): Promise<void> {
+        const currentSite = CoreSites.getCurrentSite();
+        if (!currentSite || siteId !== currentSite.getId()) {
+            return; // Site that triggered the event is not current site.
+        }
+
+        // If current page is already change password, stop.
```
NoelDeMartin, line 76:
```suggestion
        // If current page is already complete profile, stop.
```

src/core/features/user/pages/complete-profile/complete-profile.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    constructor() {
+        this.logoutLabel = CoreLoginHelper.getLogoutLabel();
+    }
+
+    /**
+     * Show a help modal.
+     */
+    showHelp(): void {
+        // @todo: Change this message in MOBILE-4059.
```
NoelDeMartin, line 48:
```suggestion
        // @todo MOBILE-4059: Change this message.
```

It's better to follow this format to improve searchability.

src/core/features/user/user.module.ts
```diff
@@ -51,6 +52,21 @@ export const CORE_USER_SERVICES: Type<unknown>[] = [
...
+            {
+                path: 'completeprofile',
+                loadChildren: () => import('./pages/complete-profile/complete-profile.module')
+                    .then(m => m.CoreUserCompleteProfilePageModule),
+            },
+        ],
+        canActivate: [CoreRedirectGuard],
+        canLoad: [CoreRedirectGuard],
+    },
+];
```
NoelDeMartin, line 68:
```suggestion
const appRoutes: Routes = [
    {
        path: 'user',
        loadChildren: () => import('@features/user/user-app-lazy.module').then(m => m.CoreUserAppLazyModule),
    },
];
```

Instead of creating a page module, I think it'd be better to create a lazy module instead. In practice, it's very similar, but conceptually I think it's better because we only want to define page modules for pages that are used in multiple routes (which is not the case here).

src/core/features/login/tests/behat/basic_usage.feature
```diff
@@ -90,3 +90,55 @@ Feature: Test basic usage of login in app
       | minimumversion | 11.0.0 | tool_mobile |
     When I enter the app
     Then I should find "App update required" in the app
+
+  Scenario: Force password change
+    Given I force a password change for user "student1"
+    When I enter the App
```
NoelDeMartin, line 96:
```suggestion
    When I enter the app
```

src/core/features/user/tests/behat/basic_usage.feature
```diff
@@ -0,0 +1,56 @@
...
+  Background:
+    Given the following "users" exist:
+      | username | firstname | lastname |
+      | student1 | Student   | Student  |
+
+  Scenario: Complete missing fields
+    Given the following "custom profile fields" exist:
+      | datatype | shortname  | name           | required |
+      | text     | food       | Favourite food | 1        |
+    When I enter the App
```
NoelDeMartin, line 13:
```suggestion
    When I enter the app
```

src/core/features/user/tests/behat/basic_usage.feature
```diff
@@ -0,0 +1,56 @@
+@user @app @javascript
```
NoelDeMartin, line 1:
```suggestion
@core @core_user @app @javascript
```

src/app/app.component.ts
dpalou: I added a new commit for this. I only moved the obvious listeners, the other ones I wasn't 100% sure were to put them so for now I've left them there, we can move them in another issue.

FYI you'll see I added a new todo, this is because I compared this with the Ionic 3 code and I saw a missing block of code, I'm not sure if it was removed on purpose or it was a mistake.

src/app/app.component.ts
dpalou: I thought about this and I wanted to change it, but then I hesitated and I left it as it was. Since you agree with me, I changed it so now no error will be displayed. It's handled here:

https://github.com/moodlehq/moodleapp/pull/3354/files#diff-cddd06a2367d72c4464ab87b2689b80871e6b75b5b237783b0ef08671b07e16aR678

NoelDeMartin (approved):

PR #3368, opened 2022-08-31
Mobile 4081

crazyserver:


src/core/services/utils/utils.ts
```diff
@@ -562,18 +562,30 @@ export class CoreUtilsProvider {
...
 
+        // Create a map first to avoid problems with not sorted.
         list.forEach((node: TreeNode<T>, index): void => {
             const id = node[idFieldName];
+
+            if (id === undefined) {
+                this.logger.error(`Node with incorrect ${idFieldName}:${id} found on formatTree`);
+            }
+
+            if (node.children == undefined) {
```
dpalou, line 571: You should use triple equal here. You can also replace this if with:

node.children = node.children ?? [];

src/core/services/utils/utils.ts
```diff
@@ -639,52 +637,50 @@ export class CoreUtilsProvider {
...
-            keys.forEach((key) => {
-                if (key.indexOf('assets.countries.') === 0) {
-                    const code = key.replace('assets.countries.', '');
-                    countries[code] = Translate.instant(key);
-                }
-            });
+        const keys = await this.getCountryKeysList();
+
+        // Now get the code and the translated name.
+        const countries = {};
```
dpalou, line 645: This variable doesn't have a type.

dpalou (changes requested):

dpalou (approved):

PR #3369, opened 2022-09-01
Mobile 4081

crazyserver:


gulp/jira.js
```diff
@@ -208,14 +208,14 @@ class Jira {
...
-                exec('mdk config show tracker.username', (err, username) => {
+                exec('mdk config show tracker.username', (error, username) => {
                     if (username) {
                         resolve({
                             url: url.replace('\n', ''),
                             username: username.replace('\n', ''),
                         });
                     } else {
-                        reject(err | 'Username not found.');
+                        reject(error | 'Username not found.');
```
dpalou, line 218: There is another error here, it should be "||" (right now it's doing a bitwise OR).

src/addons/coursecompletion/services/handlers/user.ts
```diff
@@ -57,7 +57,7 @@ export class AddonCourseCompletionUserHandlerService implements CoreUserProfileH
      * @inheritdoc
      */
     async isEnabledForUser(user: CoreUserProfile, context: CoreUserDelegateContext,  contextId: number): Promise<boolean> {
```
dpalou, line 59: Because of this change, we now have a bunch of functions declared as async but they don't do any await inside so the async isn't needed. AFAIK this doesn't cause any problem, and I guess it's not a problem in terms of performance either, so we could leave it like this if you want.

dpalou (changes requested):

gulp/jira.js
crazyserver: good catch!

src/addons/coursecompletion/services/handlers/user.ts
crazyserver: hmm that's true, but as you said there's no disadvantage of declaring it that way. Thanks anyway!

PR #3371, opened 2022-09-02
MOBILE-4081 course: Load 'course-summary.module' from service

alfonso-salces:


src/core/features/course/pages/index/index.ts
```diff
@@ -288,17 +287,7 @@ export class CoreCourseIndexPage implements OnInit, OnDestroy {
      * Open the course summary
      */
     openCourseSummary(): void {
-        if (!this.course) {
```
dpalou, line 291: I think it would be better to leave this if guard here, and make the course parameter in the function to be required.

dpalou (changes requested):

dpalou (approved):

PR #3377, opened 2022-09-07
MOBILE-4081 reconnect: Create component for identity providers

alfonso-salces:


src/core/classes/site.ts
```diff
@@ -2470,6 +2470,8 @@ export type CoreSiteIdentityProvider = {
     name: string; // The identity provider name.
     iconurl: string; // The icon URL for the provider.
     url: string; // The URL of the provider.
+    icon?: string; // Icon of the provider.
+    action?: () => unknown; // Action to execute on button click.
```
NoelDeMartin, line 2474: This is a type used in a Web Service response (`tool_mobile_get_public_config`), so I would refrain from reusing it in this way. With this change, it's implying that the webservice can return these, which is wrong and impossible (a webservice response cannot return a closure).

If you want to reuse it, I guess you can use extend it in some way:

```
type MyType = CoreIdentityProvider & {
    icon?: string; // Icon of the provider.
    action?: () => unknown; // Action to execute on button click.
};
```

src/core/features/login/pages/reconnect/reconnect.html
```diff
@@ -76,17 +76,20 @@ <h1>{{ 'core.login.reconnect' | translate }}</h1>
             {{ 'core.login.forgotten' | translate }}
         </ion-button>
 
+        <!-- Custom identity providers -->
+        <core-identity-providers></core-identity-providers>
```
NoelDeMartin, line 80: I think the naming is a bit confusing here. This component is supposed to render identity providers, but then you're rendering a list of identity providers below. So does that mean that identity providers are rendered twice? (I know they aren't but that's what it looks like).

I think just with a change in naming, everything will be more clear. For example, instead of calling these "identity providers" they could be called "login methods", which is already how they are referred to in the service. Although I'm still not sure at a glance what's the difference between an "identity provider" and a "login method", they sound the same to me. But using a different name, at least we can understand that they are different things. And hopefully reading the documentation of the code involved with each we'll be able to discern them.

src/core/features/login/services/login-methods.ts
```diff
@@ -0,0 +1,32 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+import { Injectable } from '@angular/core';
+import { CoreSiteIdentityProvider } from '@classes/site';
+import { makeSingleton } from '@singletons';
+
+@Injectable({ providedIn: 'root' })
+export class CoreLoginMethodsService {
```
NoelDeMartin, line 19: Is it necessary to create a new service just for this? Why not add a new method in the `CoreLoginHelper` service?

In general, I'm in favor of separating things by responsibility, but I'm not sure if I like it taken to this extreme. I don't see a future where anything else is added to this class. Just a suggestion though.

The service is also missing a documentation block, which should explain what this service's for (even though it could be something as simple as "login methods manager" :sweat_smile:).

NoelDeMartin (changes requested):

NoelDeMartin (approved):

PR #3378, opened 2022-09-09
Integration

MRafiqAsim:
Hi, I am facing an issue with v4.0.1 and integration, As Google Play stop accepting releases with target APIs less than 31, So I built the release with target APIs 31, and 33, to test everything is ok. I found that the push notifications stop working in android while working for iOS. 

is this is already addressed, you need to work on ?


ardiandeari:
> Hi, I am facing an issue with v4.0.1 and integration, As Google Play stop accepting releases with target APIs less than 31, So I built the release with target APIs 31, and 33, to test everything is ok. I found that the push notifications stop working in android while working for iOS.
> 
> is this is already addressed, you need to work on ?

Same problem here! Even tried to update to 31 but yet the notifications plugin seems to crash the app.

MRafiqAsim:
> > Hi, I am facing an issue with v4.0.1 and integration, As Google Play stop accepting releases with target APIs less than 31, So I built the release with target APIs 31, and 33, to test everything is ok. I found that the push notifications stop working in android while working for iOS.
> > is this is already addressed, you need to work on ?
> 
> Same problem here! Even tried to update to 31 but yet the notifications plugin seems to crash the app.

Hi,

Anybody facing this issue can sort it out by following below steps:

in package.json, do these three changes (update values , do not add new ):
"@ionic-native/push": "^5.36.0",
"@moodlehq/phonegap-plugin-push": "4.0.0-moodle.1",
"@moodlehq/phonegap-plugin-push": {
"ANDROID_SUPPORT_V13_VERSION": "28.0.0",
"FCM_VERSION": "23.+",
"IOS_FIREBASE_MESSAGING_VERSION": "~> 6.32.2",
"ANDROIDX_CORE_VERSION": "1.6.+"
},

now clean everything and re-run the build process from start. after this, the notification will work for SDK target 31 but only for when the app is in the background. in the foreground, the app will crash on notifications.

to fix that, you need to replace two files at the following path inside platforms /android :
path: app->src->main->[java](https://moodle.org/mod/glossary/showentry.php?eid=24&displayformat=dictionary)->de->appplant->cordova->plugin->notification
files :Builder.java
Notification.java

NoelDeMartin:
Hey, thanks for opening the PR.

We have already fixed this in [MOBILE-4123](https://tracker.moodle.org/browse/MOBILE-4123) (you can see the fix in this PR: https://github.com/moodlehq/moodleapp/pull/3366), but we aren't going to merge this into master until we do the next release, so I'm closing this PR.

In any case, you can continue using the `integration` branch for your own apps. But keep in mind that it is still under development, so do it at your own risk.

PR #3391, opened 2022-09-20
Mobile 4081

alfonso-salces:
This pull request has the following two changes:

- Add possibility of add toggle inside site preferences page.

- Add event to use it when main home loads.

### DIFF
--- a/src/core/features/settings/pages/site/site.html
+++ b/src/core/features/settings/pages/site/site.html
@@ -17,14 +17,17 @@
         </ion-refresher>
         <core-loading [hideUntil]="handlers.loaded">
             <ion-list>
-                <ion-item *ngFor="let handler of handlers.items" [ngClass]="['core-settings-handler', handler.class]"
-                    [attr.aria-label]="handler.title | translate" detail="true" (click)="handlers.select(handler)" button
-                    [attr.aria-current]="handlers.getItemAriaCurrent(handler)">
+                <ion-item *ngFor="let handler of handlerItems" [ngClass]="['core-settings-handler', handler.class]"
+                    [attr.aria-label]="handler.title | translate" (click)="handlers.select(handler)" [button]="!handler.toggle"
+                    [detail]="!handler.toggle" [attr.aria-current]="handlers.getItemAriaCurrent(handler)">
                     <ion-icon [name]="handler.icon" slot="start" *ngIf="handler.icon" aria-hidden="true">
                     </ion-icon>
                     <ion-label>
                         <p class="item-heading">{{ handler.title | translate}}</p>
                     </ion-label>
+                    <ion-toggle *ngIf="handler.toggle" [(ngModel)]="handler.toggleChecked"
+                        (ionChange)="handler.toggle($event.detail.checked)">
+                    </ion-toggle>
                 </ion-item>
             </ion-list>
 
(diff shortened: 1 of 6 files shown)


src/core/features/settings/pages/site/site.html
```diff
@@ -18,13 +18,16 @@ <h1>{{ 'core.settings.preferences' | translate}}</h1>
...
-                    [attr.aria-label]="handler.title | translate" detail="true" (click)="handlers.select(handler)" button
-                    [attr.aria-current]="handlers.getItemAriaCurrent(handler)">
+                    [attr.aria-label]="handler.title | translate" (click)="handlers.select(handler)" [button]="!handler.toggle"
+                    [detail]="!handler.toggle" [attr.aria-current]="handlers.getItemAriaCurrent(handler)">
                     <ion-icon [name]="handler.icon" slot="start" *ngIf="handler.icon" aria-hidden="true">
                     </ion-icon>
                     <ion-label>
                         <p class="item-heading">{{ handler.title | translate}}</p>
                     </ion-label>
+                    <ion-toggle *ngIf="handler.toggleChecked !== undefined" [(ngModel)]="handler.toggleChecked"
```
NoelDeMartin, line 28:
```suggestion
                    <ion-toggle *ngIf="handler.toggle" [(ngModel)]="handler.toggleChecked"
```

This should be more readable and we're already using the same logic on the lines above.

src/core/features/settings/services/settings-delegate.ts
```diff
@@ -52,6 +52,16 @@ export interface CoreSettingsHandlerData {
...
+
+    /**
+     * Toggle checked.
+     */
+    toggleChecked?: boolean;
+
+    /**
+     * Method for emit events to the handler.
+     */
+    toggle?(checked: boolean): void;
```
NoelDeMartin, line 64: I suppose for a toggle handler, both `toggle` and `toggleChecked` are required, right? Also, `page` is probably not needed, but according to the current implementation it'll be required. I'm thinking we could define the specific types, and type guards should take care of inferring what's optional or not.

```ts
interface CoreSettingsHandlerBaseData {
    title: string;
    class?: string;
    icon?: string;
}

interface CoreSettingsToggleHandlerData extends CoreSettingsHandlerBaseData {
    toggleChecked: boolean;
    toggle(checked: boolean): void;
}

interface CoreSettingsPageHandlerData extends CoreSettingsHandlerBaseData {
    page: string;
    params?: Params;
}

type CoreSettingsHandlerData = CoreSettingsPageHandlerData | CoreSettingsToggleHandlerData;
```

NoelDeMartin (commented):
I think this is ready for merging, but I have some suggestions, let me know what you think.

src/core/features/settings/services/settings-delegate.ts
alfonso-salces: I had to tried to implement something like your suggestion and it's works but I have to do some narrowing in settings-handlers-source.ts and in site page (site.ts). Maybe that can help to clear settings handlers of unused properties, but I don't just like narrowing.
I have to say, that maybe Im missing other way of implementation for that which don't involves narrowing.

NoelDeMartin (approved):

PR #3395, opened 2022-10-04
MOBILE-4081 core: Remove most .bind() usage

dpalou:


src/addons/qbehaviour/deferredcbm/services/handlers/deferredcbm.ts
```diff
@@ -55,8 +55,9 @@ export class AddonQbehaviourDeferredCBMHandlerService implements CoreQuestionBeh
             question,
             componentId,
             siteId,
-            this.isCompleteResponse.bind(this),
-            this.isSameResponse.bind(this),
+            (question, answers, component, componentId) => this.isCompleteResponse(question, answers, component, componentId),
+            (question, prevAnswers, prevBasicAnswers, newAnswers, newBasicAnswers, component, componentId) =>
+                this.isSameResponse(question, prevAnswers, prevBasicAnswers, newAnswers, newBasicAnswers, component, componentId),
```
NoelDeMartin, line 60:
```suggestion
            (...args) => this.isCompleteResponse(...args),
            (...args) => this.isSameResponse(...args),
```

In most places, I guess it's ok to add arguments explicitly because it's more readable. But if there are a lot of arguments, such as here, I think this approach is better. If the arguments are typed, as it happens here, this is still type safe.

src/core/services/file.ts
```diff
@@ -834,9 +834,9 @@ export class CoreFileProvider {
             return this.copyOrMoveExternalFile(from, to, copy);
         }
 
-        const moveCopyFn: MoveCopyFunction = copy ?
-            (isDir ? File.copyDir.bind(File.instance) : File.copyFile.bind(File.instance)) :
-            (isDir ? File.moveDir.bind(File.instance) : File.moveFile.bind(File.instance));
+        const moveCopyFn = (path: string, name: string, newPath: string, newName: string) => copy ?
+            (isDir ? File.copyDir(path, name, newPath, newName) : File.copyFile(path, name, newPath, newName)) :
+            (isDir ? File.moveDir(path, name, newPath, newName) : File.moveFile(path, name, newPath, newName));
```
NoelDeMartin, line 839:
```suggestion
        const moveCopyFn: MoveCopyFunction = (...args) => copy ?
            (isDir ? File.copyDir(...args) : File.copyFile(...args)) :
            (isDir ? File.moveDir(...args) : File.moveFile(...args));
```

src/core/services/file.ts
```diff
@@ -854,14 +854,16 @@ export class CoreFileProvider {
         try {
             const entry = await moveCopyFn(this.basePath, from, this.basePath, to);
 
-            return entry;
+            return <FileEntry | DirectoryEntry> entry;
```
NoelDeMartin, line 857: How do you know that the entry won't be an `Entry`?

src/core/services/file.ts
NoelDeMartin, line 839:
```suggestion
        const moveCopyFn = (...args: [string, string, string, string]) => copy ?
            (isDir ? File.copyDir(...args) : File.copyFile(...args)) :
            (isDir ? File.moveDir(...args) : File.moveFile(...args));
```

This is another way of doing it, although in both cases there's something funky going on with the response type, maybe we should take another look.

dpalou: I used this approach now.

src/core/services/file.ts
dpalou: I actually don't know why the Ionic functions are typed as `Entry`, AFAIK moving a directory will always return a `DirectoryEntry` and moving a file will always return a `FileEntry`. At least all the tests I did in the past worked like this, and I did some tests now and it works like that too in browser, Android and iOS. I've also seen that the JSDoc of `moveDir` says:

> Returns a Promise that resolves to the new DirectoryEntry object or rejects with an error.

But then the type is `DirectoryEntry|Entry`.

In any case, the cast was already there before but it was part of the `MoveCopyFunction` (it didn't throw an error because of `bind`) so maybe it's safe to keep casting it. Or if you prefer we can check if the field has some property that is only present in the child types before casting, but in that case I don't know what should we do if we ever find a case where we receive an `Entry`. Throw an error?

dpalou:
Thanks Noel, I fixed most things (and added a comment about the type cast). I also added 2 new commits, one to fix a bug in H5P and another one to fix files in Android.

Can you please review it again?

NoelDeMartin (approved):

PR #3397, opened 2022-10-05
MOBILE-4140 message: Hide overflow to wrap text inside message box

alfonso-salces:


src/core/components/message/message.scss
```diff
@@ -45,6 +45,7 @@
         .main {
             padding: 8px;
             flex-grow: 1;
+            overflow: hidden;
```
NoelDeMartin, line 48:
```suggestion
            overflow: hidden;
            text-overflow: ellipsis;
            white-space: nowrap;
```

Maybe this would be better, so that the cut is not as abrupt? This is using the "truncate" technique, for example as documented in TailwindCSS: https://tailwindcss.com/docs/text-overflow#truncate

NoelDeMartin: And I just noticed we already use it in some places with the class `text-truncate`.

NoelDeMartin (approved):

PR #3409, opened 2022-10-17
MOBILE-4120 messages: Fix wrong messages displayed in split view

dpalou:


src/addons/messages/messages-lazy.module.ts
```diff
@@ -19,7 +19,7 @@ import { buildTabMainRoutes } from '@features/mainmenu/mainmenu-tab-routing.modu
 import { AddonMessagesIndexGuard } from './guards';
 
 export const AddonMessagesDiscussionRoute: Route = {
-    path: 'discussion',
+    path: 'discussion/:conversationId/:userId',
```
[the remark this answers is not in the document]
dpalou: Thanks Noel, as we discussed in chat I ended up using the routes `'discussion/user/:userId'`, and `'discussion/:conversationId'`. I've declared them in the `messages-lazy module` instead of putting it in a `messages-discussions-lazy.module.ts` because it's more coherent with the rest of pages in that module, we can change it later if we stop using page modules in this addon.

Another thing that I changed because of this is that I created a new function `CoreNavigator. getRelativePathToParent` to make it easier to navigate when there is a split view, now the "../" are calculated automatically with this function instead of doing it manually.

The whole patch has changed quite a lot so I recommend reviewing it all again.

src/core/services/navigator.ts
```diff
@@ -700,6 +700,28 @@ export class CoreNavigatorService {
...
     }
 
+    /**
+     * Get the relative path to a parent path.
+     * E.g. if parent path is '/foo' and current path is '/foo/bar/baz' it will return '../../'.
+     *
+     * @param parentPath Parent path.
+     * @return Relative path to the parent, empty if same path or parent path not found.
+     */
+    getRelativePathToParent(parentPath: string): string {
```
NoelDeMartin, line 710: Looking at the rest of the code, I realize this is necessary because the discussion pages are not using list managers. If they were, this would be done automatically.

I realize we've dedicated enough time to this issue, so we can proceed merging this for now but I'd at least add a TODO here mentioning that this method shouldn't be necessary if split-view navigation were implemented using list managers.

src/addons/messages/messages-lazy.module.ts
```diff
@@ -18,11 +18,18 @@ import { Route, RouterModule, ROUTES, Routes } from '@angular/router';
 import { buildTabMainRoutes } from '@features/mainmenu/mainmenu-tab-routing.module';
 import { AddonMessagesIndexGuard } from './guards';
 
-export const AddonMessagesDiscussionRoute: Route = {
-    path: 'discussion',
-    loadChildren: () => import('./pages/discussion/discussion.module')
-        .then(m => m.AddonMessagesDiscussionPageModule),
-};
+export const AddonMessagesDiscussionRoutes: Route[] = [
```
NoelDeMartin, line 21: This should be named using all caps given that it's a constant, not a type.

NoelDeMartin (approved):
I just found a couple of small things we could improve, but we are ready for merging.

src/core/singletons/tests/navigator.test.ts
```diff
@@ -0,0 +1,36 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { mockSingleton } from '@/testing/utils';
+import { CoreNavigator } from '@services/navigator';
+
+describe('CoreNavigator singleton', () => {
+
+    it('calculates relative paths to parent paths', () => {
```
NoelDeMartin, line 20: I just realized we already have tests for the navigator under `src/core/services/tests/navigator.test.ts`, this should be moved there.

dpalou:
Thanks Noel, I fixed these last things you raised.

PR #3411, opened 2022-10-19
Mobile 4081

crazyserver:


src/core/components/horizontal-scroll-controls/horizontal-scroll-controls.ts
```diff
@@ -44,13 +44,17 @@ export class CoreHorizontalScrollControlsComponent {
     /**
      * Scroll the target in the given direction.
      *
-     * @param direction Scroll direction.
+     * @param ev Click event
+    * @param direction Scroll direction.
```
dpalou, line 48: Wrong indentation.

src/core/features/mainmenu/services/handlers/mainmenu.ts
```diff
@@ -34,12 +32,9 @@ export class CoreMainMenuHomeHandlerService implements CoreMainMenuHandler {
...
      */
     async isEnabled(): Promise<boolean> {
-        const siteId = CoreSites.getCurrentSiteId();
+        await CoreMainMenuHomeDelegate.waitForReady();
 
-        const dashboardEnabled = await CoreDashboardHomeHandler.isEnabledForSite(siteId);
-        const siteHomeEnabled = await CoreSiteHomeHomeHandler.isEnabledForSite(siteId);
-
-        return dashboardEnabled || siteHomeEnabled;
+        return CoreMainMenuHomeDelegate.hasHandlers(true);
```
dpalou, line 37: I think this change is wrong. Imagine the following case:

1. App starts and the handlers begin initializing.
2. This handler waits for the home handlers to be ready.
3. Home handlers finish initializing, let's imagine they're all disabled.
4. This handler finishes and it detects handlers are disabled so Home tab is not displayed.
5. Now we receive an event to update the handlers (site info updated, site plugins loaded, etc.). The handlers are calculated again.
6. This handler's `isEnabled` starts again, and `waitForReady` is resolved immediately because the init promise isn't resetted. So it still says there are no handlers.
7. One of the home handlers finishes recalculating and now it's enabled. But the home tab is not displayed because `isEnabled` finished in the previous step.

I opened an issue to do what you're trying to do here, but it won't be included in 4.1:

https://tracker.moodle.org/browse/MOBILE-4138

It's not a trivial change, so I'd leave it for 4.2 unless you really need it for 4.1. If you really need it for 4.1 then it would be better to move the commit to the issue above.

dpalou (changes requested):

src/core/features/mainmenu/services/handlers/mainmenu.ts
crazyserver: OK I left a partial solution here.

dpalou (approved):

PR #3421, opened 2022-10-26
MOBILE-4145: Remove REQUEST_INSTALL_PACKAGES permission

NoelDeMartin:


src/core/services/file.ts
```diff
@@ -1031,20 +1018,6 @@ export class CoreFileProvider {
...
-     * @param path Path.
-     * @return Path without a slash in the first position.
-     */
-    removeStartingSlash(path: string): string {
-        if (path[0] == '/') {
-            return path.substring(1);
-        }
-
-        return path;
-    }
```
dpalou, line 1046: This function is public, so we should deprecate it instead of deleting it. People can use `CoreText.removeStartingSlash` instead.

src/core/lang.json
```diff
@@ -18,6 +18,7 @@
     "cannotconnecttrouble": "We're having trouble connecting to your site.",
     "cannotconnectverify": "<strong>Please check the address is correct.</strong>",
     "cannotdownloadfiles": "File downloading is disabled. Please contact your site administrator.",
+    "cannotinstallapk": "For security reasons, it is not allowed to install unknown apps from the Moodle App. Please open the file using a browser.",
```
dpalou, line 21: Did you review this string with the UX copy team? I wrote this text for the temporary patch (4.0.2) but we didn't have time to review it with Júlia.

dpalou (approved):

PR #3431, opened 2022-11-03
Mobile 4100

dpalou:


crazyserver:
Changes are ok, but you should add some testing if there are generators. Please, add a new commit for that.

PR #3436, opened 2022-11-10
MOBILE-4177 forum: Empty box with auto height

crazyserver:


dpalou (changes requested):
Thanks Pau,

this will fix the problem for forum, but I think this problem affects any empty box inside a split view left pane. I think the height wasn't calculated right in glossary either for example. Maybe we should put this fix inside the split-view CSS, probably only for left pane?

crazyserver:
Actually the problem is with any page: The height of the box is 100% so it will occupy the same height as her parent.
The thing is that when other elements are in page which is not very common because when it's empty, usually nothing else is shown, the element goes down. If I fix that for all the left panels, it will be at the top on those pages.

One solution is to check the top position of the element and remove the height 100% when this position is more than a threshold. But this is a JS solution... That do you think?

crazyserver:
After some chat, I've changed the approach to match splitview problem.

dpalou (approved):

PR #3444, opened 2022-11-15
MOBILE-4059: Fix issues found during testing

NoelDeMartin:


src/core/features/login/services/login-helper.ts
```diff
@@ -1515,6 +1533,40 @@ export class CoreLoginHelperProvider {
...
+    buildExceededAttemptsHTML(canContactSupport: boolean, canRecoverPassword: boolean): SafeHtml | string | null {
+        const safeHTML = (html: string) => DomSanitizer.sanitize(SecurityContext.HTML, html) ?? '';
+        const recoverPasswordHTML = (messageKey: string) => {
+            const placeholder = '%%RECOVER_PASSWORD%%';
+            const message = safeHTML(Translate.instant(messageKey, { recoverPassword: placeholder }));
+            const recoverPassword = safeHTML(Translate.instant('core.login.exceededloginattemptsrecoverpassword'));
+
+            return DomSanitizer.bypassSecurityTrustHtml(
+                message.replace(placeholder, `<a href="#" role="button" style="color:inherit">${recoverPassword}</a>`),
+            );
```
dpalou, line 1552: Why are you using a placeholder and then replacing it immediately? Wouldn't it be better to just use the final text?

```suggestion
            const recoverPassword = safeHTML(Translate.instant('core.login.exceededloginattemptsrecoverpassword'));
            const message = safeHTML(Translate.instant(messageKey, {
                recoverPassword: `<a href="#" role="button" style="color:inherit">${recoverPassword}</a>`,
            }));

            return DomSanitizer.bypassSecurityTrustHtml(message);
```

dpalou (changes requested):

src/core/features/login/services/login-helper.ts
NoelDeMartin: This is necessary because I want to replace the placeholder with some html. If I don't replace it after sanitazing it removes the `style="color:inherit"` part.

Which by the way, I'm adding inline because it doesn't work if I do it in the component CSS :/. I guess in order to work I'd have to use `::ng-deep` or something like that. But I think I prefer inline styles given that it's text inserted dynamically, so it won't appear in the template file anyways. What do you think?

dpalou: Ok, let's leave it like this then.

dpalou (approved):

PR #3445, opened 2022-11-15
Mobile 3936

crazyserver:


src/addons/mod/assign/assign.module.ts
```diff
@@ -77,6 +81,20 @@ const routes: Routes = [
...
+                        await ApplicationInit.donePromise;
+
+                        CoreCourseHelper.navigateToModule(
+                            notification.instanceId,
+                            {
+                                siteId: notification.siteId,
+                            },
+                        );
+                    },
+                );
```
dpalou, line 97: I like that each module registers their own "click handler" in the module, but we have a lot of duplicated code. We could create a function to abstract this code, something like:

`CoreCourseHelper.registerModuleReminderClick(AddonModAssignProvider.COMPONENT);`

And that helper function would register the click and so on. What do you think?

dpalou (changes requested):

src/addons/mod/assign/assign.module.ts
crazyserver: Your true!!! Done!

dpalou (approved):

PR #3463, opened 2022-11-23
Mobile 4081

dpalou:


NoelDeMartin (approved):

src/addons/notifications/services/handlers/push-click.ts
```diff
@@ -141,4 +141,5 @@ export type AddonNotificationsNotificationData = CorePushNotificationsNotificati
     contexturl?: string; // URL related to the notification.
     savedmessageid?: number; // Notification ID (optional).
     id?: number; // Notification ID (optional).
+    date?: string; // Notification date.
```
NoelDeMartin, line 144: Is this a timestamp? If so, why is it a string? At the very least, I would document that it should be a number, because initially I thought this was supposed to be a date string like `2022-11-23T11:53:41.496Z`.

And where does this come from, it is from a plugin or we send it from the airnotifier or the LMS?

NoelDeMartin (changes requested):

src/addons/notifications/services/handlers/push-click.ts
dpalou: I updated the comment to specify what it is.

I don't know where it comes from. I don't think LMS is adding it, I'd say it's probably added by either Airnotifier or FCM.

NoelDeMartin (approved):

PR #3464, opened 2022-11-23
Mobile 4081

crazyserver:


src/core/features/settings/pages/space-usage/space-usage.ts
```diff
@@ -76,31 +92,59 @@ export class CoreSettingsSpaceUsagePage implements OnInit, OnDestroy {
...
+        let siteUrl = '';
 
-        const settingsHelper = CoreSettingsHelper.instance;
+        const currentSiteId = CoreSites.getCurrentSiteId();
+
+        if (currentSiteId) {
+            const index = sites.findIndex((site) => site.id === currentSiteId);
+
+            const site = sites.splice(index, 1)[0];
+            this.accountsList.currentSite = await this.getSiteWithUsage(site);
```
dpalou, line 106: I think it would be better to calculate this along with the rest of sites. Right now you're blocking first to calculate the current site size, and then to calculate the rest of sites. For performance maybe it would be better to calculate them all at once in parallel.

dpalou (changes requested):

src/core/features/settings/pages/space-usage/space-usage.ts
crazyserver: Thanks! Done!

dpalou (approved):

PR #3465, opened 2022-11-23
MOBILE-4081 theme: Fix background mode in rich text editor

alfonso-salces:


src/theme/theme.light.scss
```diff
@@ -402,4 +402,6 @@ html {
     } @else {
         --core-mainpage-sitename-display: inline;
     }
+
+    --rte-editor-background: var(--ion-background-color);
```
crazyserver, line 406: I'd use here :
--rte-editor-background: var(--core-input-background); 

Probably with that, the changes on theme.dark are not needed.

crazyserver (changes requested):

crazyserver (approved):

PR #3478, opened 2022-11-25
Mobile 4081

crazyserver:


src/addons/coursecompletion/services/handlers/completionstatus-link.ts
```diff
@@ -0,0 +1,80 @@
...
+                    { params: pageParams, siteId },
+                );
+            },
+        }];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    async isEnabled(siteId: string, url: string, params: Record<string, string>, courseId?: number): Promise<boolean> {
```
dpalou, line 67: You aren't using the courseId param so I wouldn't declare it, that way the code is clearer IMO.

dpalou (changes requested):

src/addons/coursecompletion/services/handlers/completionstatus-link.ts
crazyserver: Done! Also fixed the checks

dpalou (approved):

PR #3480, opened 2022-11-28
translation mistake

saman222:


saman222:
hi
I changed some setting. Is it ok?

NoelDeMartin:
Hey thanks for opening the PR. As I told you in #3479, we develop using the `integration` branch so PRs are only merged using that branch. I also noticed that you were missing the declaration in the `langindex.json` file which is necessary when adding new strings.

I realize that you may not understand all of this if it's your first time contributing to the app, so I just went ahead and fixed that in a separate PR #3482. I added you as Co-author to keep the attribution.

I'm closing this now as the fix will be included in the other PR, thanks for contributing!

saman222:
excuse me @NoelDeMartin for can't working with github correctly but i find another translation mistake in:
https://github.com/moodlehq/moodleapp/blob/227be9fc27b8cfd43cb1f4ca5ae54605601b846f/src/core/services/utils/utils.ts#L1636
and in :
https://github.com/moodlehq/moodleapp/blob/227be9fc27b8cfd43cb1f4ca5ae54605601b846f/src/core/services/utils/utils.ts#L680

https://github.com/moodlehq/moodleapp/blob/227be9fc27b8cfd43cb1f4ca5ae54605601b846f/src/core/services/utils/utils.ts#L708

I hope this post is not disrespectful to you. i can't do that. if is possible give me teaching link how to do that.

NoelDeMartin:
Thanks @saman222 for pointing that out.

We've decided not to translate these messages because they are errors and shouldn't show up so often in the UI, maybe never if they are caught somewhere else in the code. In the rare case that they appear in the UI, we prefer not to translate them because they are error messages that a developer would understand; but we don't want to introduce these in [AMOS](https://lang.moodle.org/) to avoid confusion and reduce the workload.

The other one you found was in the UI and not an error message, that's why we translated that one.

PR #3485, opened 2022-11-29
MOBILE-4176: Grades improvements

NoelDeMartin:


src/core/directives/fa-icon.ts
```diff
@@ -17,6 +17,11 @@ import { CoreLogger } from '@singletons/logger';
 import { Http } from '@singletons';
 import { CoreConstants } from '@/core/constants';
 
+/**
+ * Object use to store whether icons exist or not during development.
+ */
+const DEV_ICONS_STATUS: Record<string, Promise<boolean>> = {};
```
dpalou, line 23: IIRC we decided to use readonly properties instead of const variables :)

dpalou (changes requested):

dpalou (approved):

PR #3487, opened 2022-11-29
[4.2] MOBILE-4202 split-view: Scroll to current element when swipe

alfonso-salces:
Scroll to current element when swipe in split-view page.

### DIFF
--- a/src/core/classes/items-management/items-manager.ts
+++ b/src/core/classes/items-management/items-manager.ts
@@ -103,6 +103,10 @@ export abstract class CoreItemsManager<
      * @param item Item, null if none.
      */
     setSelectedItem(item: Item | null): void {
+        if (item === this.selectedItem) {
+            return;
+        }
+
         this.selectedItem = item;
 
         this.listeners.forEach(listener => listener.onSelectedItemUpdated?.call(listener, item));
--- a/src/core/classes/items-management/list-items-manager.ts
+++ b/src/core/classes/items-management/list-items-manager.ts
@@ -22,6 +22,7 @@ import { CoreUtils } from '@services/utils/utils';
 
 import { CoreRoutedItemsManagerSource } from './routed-items-manager-source';
 import { CoreRoutedItemsManager } from './routed-items-manager';
+import { CoreDom } from '@singletons/dom';
 
 /**
  * Helper class to manage the state and routing of a list of items in a page.
@@ -129,6 +130,7 @@ export class CoreListItemsManager<
         }
 
         await this.navigateToItem(item, { reset: this.resetNavigation() });
+        setTimeout(async () => await this.scrollToCurrentElement(), 100);
     }
 
     /**
@@ -196,8 +198,31 @@ export class CoreListItemsManager<
         super.updateSelectedItem(route);
 
         const selectDefault = CoreScreen.isTablet && this.selectedItem === null && this.splitView && !this.splitView.isNested;
-
         this.select(selectDefault ? this.getDefaultItem() : this.selectedItem);
+    }
+
+    /**
+     * Scroll to current element in split-view list.
+     */
+    protected async scrollToCurrentElement(): Promise<void> {
+        if (CoreScreen.isMobile) {
+            return;
+        }
+
+        const element = this.splitView?.nativeElement ?? document;
+        const currentItem = element.querySelector<HTMLElement>('[aria-current="page"]');
+
+        if (!currentItem) {
+            return;
+        }
+
+        const isElementInViewport = CoreDom.isElementInViewport(currentItem, 1, this.splitView?.nativeElement);
+
+        if (isElementInViewport) {
+            return;
+        }
+
+        currentItem.scrollIntoView({ behavior: 'smooth' });
     }
 
     /**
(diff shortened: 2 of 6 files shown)


src/core/classes/items-management/swipe-navigation-items-manager.ts
```diff
@@ -107,6 +108,7 @@ export class CoreSwipeNavigationItemsManager<
         }
 
         await this.navigateToItem(item, { animationDirection, replace: true });
+        this.scrollToCurrentElement(delta);
```
NoelDeMartin, line 111: I think you're calling this here instead of inside of `navigateToItem` because you are using the `delta` to check whether to scroll to the next or the previous, but I don't understand why. What's the problem to scroll to the current item?

If the problem is that the "current item" is not actually the current item, but the one that will be selected next, I don't think this is the proper way to do it. I can see two ways.

The first one is to actually way for the selected item to be selected (I'm not sure if that can be achieved with nextTick or something else). The other option would be to actually find the item that will be selected next.

In both case, I think it's important to get the actually selected item rather than using the current algorithm. First, because delta may be more bigger than 1 or smaller than -1. And second, because the current implementation is looking at any element matching `[aria-current="page"]`, and for all we know that element may not even be one of the items in the split view. So I think we may need to get more creative with this solution.

Also, the manager in charge of the actual items seems to be the list manager, not the swipe items manager. So I think we need to take another look to solving this properly. Maybe we can do a call tomorrow and pair on this if you want.

NoelDeMartin (changes requested):

src/core/classes/items-management/list-items-manager.ts
```diff
@@ -196,8 +200,32 @@ export class CoreListItemsManager<
         super.updateSelectedItem(route);
 
         const selectDefault = CoreScreen.isTablet && this.selectedItem === null && this.splitView && !this.splitView.isNested;
+        const selectedItem = selectDefault ? this.getDefaultItem() : this.selectedItem;
+        this.select(selectedItem);
+    }
```
NoelDeMartin, line 205:
```suggestion
        this.select(selectDefault ? this.getDefaultItem() : this.selectedItem);
    }
```

I think it's better to leave this as this for a cleaner diff (it's probably changed because we had a different approach before).

NoelDeMartin (changes requested):
Overall I think it's good now, but some E2E tests related with the split-view are failing so you should rebase to make sure that they are not related with these changes.

src/core/classes/items-management/routed-items-manager.ts
```diff
@@ -147,15 +147,6 @@ export abstract class CoreRoutedItemsManager<
...
 
-    /**
-     * @inheritdoc
-     */
-    protected onSourceItemsUpdated(items: Item[]): void {
-        super.onSourceItemsUpdated(items);
-
-        this.updateSelectedItem();
-    }
-
```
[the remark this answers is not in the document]
alfonso-salces: I do that because if I scroll to the end of the list and load more items is scrolling to the selected item because we are calling to `this.updateSelectedItem()`.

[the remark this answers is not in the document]
NoelDeMartin: Hm I see, can you try instead of removing it only call `this.updateSelectedItem()` if the current item is null or it doesn't exist on the list of loaded items?

The idea of this override was to automatically select an item that was selected before the items load, which could happen for example opening a split-view with the url on a selected item. I think it's important to keep that behaviour, or it affects the UX of split-views because they could potentially open without a selected item which is weird.

src/core/classes/items-management/items-manager.ts
```diff
@@ -103,6 +103,10 @@ export abstract class CoreItemsManager<
      * @param item Item, null if none.
      */
     setSelectedItem(item: Item | null): void {
+        if (item && this.selectedItem && this.getItemId(item) === this.getItemId(this.selectedItem)) {
```
NoelDeMartin, line 106:
```suggestion
        if (item === this.selectedItem) {
```

I think this is what's breaking the tests, so with this we should be ready to merge :). Although you may need to rebase because there's been a while since we looked into this. But I think it should be fine, because we haven't changed any of the related files that I'm aware.

NoelDeMartin (changes requested):

src/core/classes/items-management/routed-items-manager.ts
```diff
@@ -152,6 +152,11 @@ export abstract class CoreRoutedItemsManager<
      */
     protected onSourceItemsUpdated(items: Item[]): void {
         super.onSourceItemsUpdated(items);
+        const selectedItem = this.selectedItem;
+
+        if (selectedItem !== null && items.some(item => this.getItemId(item) === this.getItemId(selectedItem))) {
```
NoelDeMartin, line 157:
```suggestion
        if (selectedItem !== null && items.some(item => item === selectedItem)) {
```

We also need to compare by reference rather than id here. I've reviewed the code to make sure we're not forgetting it anywhere else, and I think it should be fine after fixing this line.

NoelDeMartin (changes requested):

NoelDeMartin (approved):

PR #3503, opened 2022-12-05
MOBILE-4176 grades: Keep new lines

NoelDeMartin:


src/core/services/utils/text.ts
```diff
@@ -277,10 +277,12 @@ export class CoreTextUtilsProvider {
...
      *
      * @param text The text to be cleaned.
-     * @param singleLine True if new lines should be removed (all the text in a single line).
+     * @param options Processing options.
+     * @param options.singleLine True if new lines should be removed (all the text in a single line).
+     * @param options.trim True if text should be trimmed.
      * @returns Clean text.
      */
-    cleanTags(text: string | undefined, singleLine?: boolean): string {
+    cleanTags(text: string | undefined, options: { singleLine?: boolean; trim?: boolean } = {}): string {
```
dpalou, line 285: This is a breaking change, we should document it in upgrade.txt.

dpalou (changes requested):

src/core/services/utils/text.ts
NoelDeMartin: Thanks, I didn't remember we had that file. I also added other upgrade notes we had neglected.

dpalou (approved):

PR #3523, opened 2023-01-13
[4.3] MOBILE-4232 database: Create in memory database for testing

alfonso-salces:


src/core/classes/database/testing-database-table.ts
```diff
@@ -0,0 +1,220 @@
...
+     */
+    async initialize(): Promise<void> {
+        const records = await this.getMany();
+
+        this.records = records.reduce((data, record) => {
+            const primaryKey = this.serializePrimaryKey(this.getPrimaryKeyFromRecord(record));
+            data[primaryKey] = record;
+
+            return data;
+        }, {});
```
NoelDeMartin, line 47: I think this can be removed, given that the `getMany` method is reading attributes from `this.records` already, so this won't ever get any new records.

src/core/classes/database/testing-database-table.ts
```diff
@@ -0,0 +1,220 @@
...
+    /**
+     * Augment CoreDatabaseConfiguration interface with data specific to this table.
+     *
+     * @see https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation
+     */
+    export interface CoreDatabaseConfiguration {
+        lazyCacheLifetime: number;
+    }
+
+}
```
NoelDeMartin, line 220: This should be removed.

src/core/classes/database/testing-database-table.ts
```diff
@@ -0,0 +1,220 @@
...
+    CoreDatabaseConditions,
+    GetDBRecordPrimaryKey,
+    CoreDatabaseReducer,
+    CoreDatabaseQueryOptions,
+} from './database-table';
+
+/**
+ * Wrapper used to caching records in testing environment.
+ */
+export class CoreTestingDatabaseTable<
```
NoelDeMartin, line 28: Overall I think this looks fine, because everything has been taken from the eager database table, right?

Keeping that in mind, what do you think about extracting the code rather than duplicating it? I think it should be possible to create an instance of this table inside of the eager table, and use it to do everything in memory (the eager table class would still need to perform writes on the real database though).

I don't see anything specific to testing in this database table, so maybe it could be called something like `CoreInMemoryDatabaseTable` instead? Which I realize is a name that already exists, but maybe that wasn't named properly. Looking at that class, maybe it should be named `CoreSingletonDatabaseTable` because that's what it cares about.

Later on, if needed, we could have a specific testing database table with testing-specific functionality, but I don't think it's necessary yet. There's nothing wrong with continuing to call the strategy "testing" though.

NoelDeMartin (changes requested):

src/core/classes/database/eager-database-table.ts
```diff
@@ -32,9 +33,20 @@ export class CoreEagerDatabaseTable<
     DBRecord extends SQLiteDBRecordValues = SQLiteDBRecordValues,
     PrimaryKeyColumn extends keyof DBRecord = 'id',
     PrimaryKey extends GetDBRecordPrimaryKey<DBRecord, PrimaryKeyColumn> = GetDBRecordPrimaryKey<DBRecord, PrimaryKeyColumn>
-> extends CoreInMemoryDatabaseTable<DBRecord, PrimaryKeyColumn, PrimaryKey> {
+> extends CoreSingletonDatabaseTable<DBRecord, PrimaryKeyColumn, PrimaryKey> {
 
     protected records: Record<string, DBRecord> = {};
```
NoelDeMartin, line 38:
```suggestion

```

NoelDeMartin: Other than this, the initialize method should populate the inMemoryDb.

src/core/classes/database/singleton-database-table.ts
```diff
@@ -0,0 +1,73 @@
...
+     * @inheritdoc
+     */
+    async initialize(): Promise<void> {
+        await super.initialize();
+
+        const activeTables = CoreSingletonDatabaseTable.ACTIVE_TABLES.get(this.database) ?? new Set();
+
+        if (activeTables.has(this.tableName)) {
+            const message = `Trying to create multiple instances of an in-memory table for '${this.tableName}', ` +
+                'use singletons instead.';
```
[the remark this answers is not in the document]
NoelDeMartin: Still not fixed.

src/core/classes/tests/database-table.test.ts
```diff
@@ -221,11 +268,12 @@ describe('CoreDatabaseTable with lazy caching', () => {
 });
 
 describe('CoreDatabaseTable with no caching', () => {
-
+    CoreConstants.CONFIG.databaseOptimizations = { cachingStrategy: CoreDatabaseCachingStrategy.None };
```
NoelDeMartin, line 271: I think this shouldn't be here.

src/core/classes/database/eager-database-table.ts
```diff
@@ -32,48 +32,43 @@ export class CoreEagerDatabaseTable<
...
-        this.records = records.reduce((data, record) => {
+        const updatedRecords = records.reduce((data, record) => {
             const primaryKey = this.serializePrimaryKey(this.getPrimaryKeyFromRecord(record));
 
             data[primaryKey] = record;
 
             return data;
         }, {});
+
+        this.inMemoryDb.setRecords(updatedRecords);
```
NoelDeMartin, line 57: Maybe instead of implementing this `setRecords` method, you can just call `.insert` for each record. If you do it like this, classes calling `setRecords` have to know internals of the inMemoryDb in order to set the keys properly.

NoelDeMartin (changes requested):

PR #3524, opened 2023-01-15
feat(core): update angular 11

phyr0s:
The application is very outdated and little by little it must be updated to the latest version of Angular and Ionic available


NoelDeMartin:
Hey, thanks for opening the PR.

We're aware that we're using an outdated version of Angular, but updating is not so trivial as changing the versions for our app. We have some features such as [site plugins](https://moodledev.io/general/app/development/plugins-development-guide) that break when updating to Angular 12, so we have to decide what to do about that before upgrading.

I realize you suggested to update to Angular 11, not 12, but that's also an outdated version so when we update we'll probably do it to a newer version.

You can track the progress of this issue in [MOBILE-3947](https://tracker.moodle.org/browse/MOBILE-3947).

phyr0s:
@NoelDeMartin  to update you need to go version by version. When you arrived to version 12 then you can update ionic to v6, toolkit and cordova builders. Also @ionic-native to @awesome-cordova-plugins

PR #3525, opened 2023-01-16
MOBILE-4069 book: Add basic usage behat tests

dpalou:


dpalou:
Basically, lack of time. I already spent more time than I had expected to doing this, so I decided to stop here and leave the TODOs for later. E.g. we still don't have a step to swipe in ion-slides.

NoelDeMartin:
Ok, in that case I think would remove the TODOs altogether. We already have many things that are not tested, so I think it's ok that these TODOs are implicit.

NoelDeMartin (approved):
After discussing it with the rest of the team we decided to keep the TODOs, so I'm merging it.

PR #3527, opened 2023-01-20
MOBILE-4234 mod_data: Correctly handle 0 in number fields

dpalou:


src/addons/mod/data/pages/edit/edit.ts
```diff
@@ -368,9 +368,14 @@ export class AddonModDataEditPage implements OnInit {
                             }
                         });
                     }
+
                     this.jsData!.errors = this.errors;
 
                     this.scrollToFirstError();
+
+                    if (updateEntryResult.generalnotifications?.length) {
+                        throw new CoreError(updateEntryResult.generalnotifications[0]);
```
NoelDeMartin, line 377: It doesn't seem like this notification is necessarily an error, so maybe we should show an alert instead of throwing? And it can potentially return more than one notification, so maybe we could show all of them together in the same alert or something.

NoelDeMartin (approved):

PR #3533, opened 2023-01-25
Mobile 4077 - Support user custom reports

alfonso-salces:


src/core/features/reportbuilder/components/report-detail/report-detail.html
```diff
@@ -0,0 +1,54 @@
...
+
+    <core-loading [hideUntil]="state.loaded">
+
+        <ng-container *ngIf="state.report?.data?.rows && state.report?.data?.headers && state.report?.details">
+
+            <ion-refresher slot="fixed" [disabled]="!state.loaded" (ionRefresh)="refreshReport($event.target)">
+                <ion-refresher-content pullingText="{{ 'core.pulltorefresh' | translate }}"></ion-refresher-content>
+            </ion-refresher>
+
+            <ng-container *ngIf="layout === 'card' || (isMobile && layout === 'adaptative')">
```
NoelDeMartin, line 11: Instead of doing this, maybe we could compute the "effective layout" in the controller and have this variable be only `card` or `table`? I get that this is a property, but in another comment I suggest that this shouldn't be a component at all, and even if it is we should be able to create a computed variable with another name (maybe effectiveLayout?)

src/core/features/reportbuilder/components/report-detail/report-detail.ts
```diff
@@ -0,0 +1,210 @@
...
+    async ngOnInit(): Promise<void> {
+        this.isMobile = CoreScreen.isMobile;
+        await this.getReport();
+        this.updateState({ loaded: true });
+    }
+
+    /**
+     * Get report data.
+     */
+    async getReport(): Promise<void> {
```
NoelDeMartin, line 67: I think this should be called `loadReport` instead.

src/core/features/reportbuilder/pages/list/list.html
```diff
@@ -0,0 +1,29 @@
...
+            <h1>{{ 'core.reportbuilder.menu' | translate }}</h1>
+        </ion-title>
+    </ion-toolbar>
+</ion-header>
+<ion-content *ngIf="state$ | async as state">
+    <ion-refresher slot="fixed" [disabled]="!state.loaded" (ionRefresh)="refreshReports($event.target)">
+        <ion-refresher-content pullingText="{{ 'core.pulltorefresh' | translate }}"></ion-refresher-content>
+    </ion-refresher>
+    <core-loading [hideUntil]="state.loaded">
+        <ion-list *ngIf="reports.items">
```
NoelDeMartin, line 16: What happens if the user has no reports? Would this show an empty screen, or is this page unreachable? Maybe we need to handle the empty state here with some text.

src/core/features/reportbuilder/pages/list/list.module.ts
```diff
@@ -0,0 +1,34 @@
...
+];
+
+@NgModule({
+    imports: [
+        RouterModule.forChild(routes),
+        CoreSharedModule,
+    ],
+    declarations: [CoreReportBuilderListPage],
+})
+export class CoreReportBuilderListModule {}
```
NoelDeMartin, line 34: We've decided to avoid page modules if page components are not reused in multiple modules, so this should be removed and the page can be declared in the lazy module directly.

src/core/features/reportbuilder/pages/report/report.module.ts
```diff
@@ -0,0 +1,36 @@
...
+
+@NgModule({
+    imports: [
+        CoreReportBuilderComponentsModule,
+        CoreSharedModule,
+        RouterModule.forChild(routes),
+    ],
+    declarations: [CoreReportBuilderReportPage],
+})
+export class CoreReportBuilderReportModule {}
```
NoelDeMartin, line 36: Same as the other page module, we should remove it.

src/core/features/reportbuilder/pages/list/list.ts
```diff
@@ -0,0 +1,133 @@
...
+} from '@features/reportbuilder/services/reportbuilder';
+import { IonRefresher } from '@ionic/angular';
+import { CoreNavigator } from '@services/navigator';
+import { CoreDomUtils } from '@services/utils/dom';
+import { CoreUtils } from '@services/utils/utils';
+import { BehaviorSubject } from 'rxjs';
+
+@Component({
+    selector: 'core-report-builder-list',
+    styles: [''],
```
NoelDeMartin, line 32:
```suggestion
```

src/core/features/reportbuilder/tests/behat/reportbuilder.feature
```diff
@@ -0,0 +1,145 @@
+@app @javascript @core_reportbuilder @alfonsodebug
```
NoelDeMartin, line 1:
```suggestion
@app @javascript @core_reportbuilder
```

src/core/features/reportbuilder/pages/report/report.html
```diff
@@ -0,0 +1,16 @@
...
+        </ion-buttons>
+        <ion-title *ngIf="name && sourceName">
+            <h1> {{ name }} </h1>
+            <p class="subheading"> {{ sourceName }} </p>
+        </ion-title>
+    </ion-toolbar>
+</ion-header>
+
+<ion-content>
+    <core-report-builder-report-detail [isBlock]="false" [reportId]="reportId" (onReportLoaded)="loadTitle($event)">
```
NoelDeMartin, line 14: It seems like this component is only used here and it fills the entire page, so what's the reason to extract a separate component?

NoelDeMartin (changes requested):

scripts/langindex.json
```diff
@@ -1652,6 +1652,9 @@
   "core.courses.totalcoursesearchresults": "local_moodlemobileapp",
   "core.currentdevice": "local_moodlemobileapp",
   "core.custom": "form",
+  "core.reportbuilder.menu": "tool_reportbuilder",
```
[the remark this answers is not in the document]
alfonso-salces: Nice catch, I renamed it by mistake.
The string that i was using before is `core.reportbuilder.customreports`, but I agree that maybe we should show to the user "Reports" so I will use `core.reportbuilder.reportstab`.

src/core/features/reportbuilder/pages/report/report.html
alfonso-salces: Because it can be loaded as block.

NoelDeMartin (approved):

PR #3538, opened 2023-01-31
Mobile 4166

dpalou:


src/theme/theme.base.scss
```diff
@@ -1796,3 +1796,86 @@ ion-modal.core-modal-no-background {
         display: none;
     }
 }
+
+/**
+ * VideoJS modifications.
+ * Styles extracted from the end of https://github.com/moodle/moodle/blob/master/media/player/videojs/styles.css
```
NoelDeMartin, line 1802: Instead of adding this here, I think it'd be better to create a new `src/theme/components/videojs.scss` file.

I'm also not a fan of including this type of comments saying where the code was taken from. If the code is 100% copy-pasted and will never change, I guess it's ok. But as soon as it changes in some way or functionality is added or modified, the comment becomes misleading. Instead, I think it's better to include the code in a separate commit and add this information in the commit message (even co-authoring authors who worked on the original repository as well, if we're doing this to keep attribution).

src/addons/filter/mediaplugin/services/handlers/mediaplugin.ts
```diff
@@ -41,25 +47,100 @@ export class AddonFilterMediaPluginHandlerService extends CoreFilterDefaultHandl
...
+        const mediaElements = Array.from(container.querySelectorAll<HTMLVideoElement | HTMLAudioElement>('video, audio'));
+
+        mediaElements.forEach((mediaElement) => {
+            if (CoreDom.mediaUsesJavascriptPlayer(mediaElement)) {
+                this.useVideoJS(mediaElement);
+            } else {
+                // Remove the VideoJS classes and data if present.
+                mediaElement.classList.remove('video-js');
+                mediaElement.removeAttribute('data-setup');
+                mediaElement.removeAttribute('data-setup-lazy');
```
NoelDeMartin, line 69: I'm not sure why this is necessary, when were this classes added in the first place if not by the call above?

src/addons/filter/mediaplugin/services/handlers/mediaplugin.ts
```diff
@@ -41,25 +47,100 @@ export class AddonFilterMediaPluginHandlerService extends CoreFilterDefaultHandl
...
+            }
+        });
+    }
+
+    /**
+     * Use video JS in a certain video.
```
NoelDeMartin, line 75:
```suggestion
     * Use video JS in a certain video or audio.
```

src/core/directives/external-content.ts
```diff
@@ -266,7 +280,7 @@ export class CoreExternalContentDirective implements AfterViewInit, OnChanges, O
             return;
         }
 
-        let urls = inlineStyles.match(/https?:\/\/[^"') ;]*/g);
+        let urls = Array.from(inlineStyles.match(/https?:\/\/[^"') ;]*/g) ?? []);
         if (!urls || !urls.length) {
             return;
         }
```
NoelDeMartin, line 286:
```suggestion
        const urls = CoreUtils.uniqueArray(Array.from(inlineStyles.match(/https?:\/\/[^"') ;]*/g) ?? []));
        if (!urls.length) {
```

src/core/services/app.ts
```diff
@@ -266,6 +266,24 @@ export class CoreAppProvider {
...
+    }
+
+    /**
+     * Checks if the app is running in an iPhone device.
+     *
+     * @returns Whether the app is running in an iPhone device.
+     */
+    isIPhone(): boolean {
+        return CoreApp.isIOS() && CorePlatform.is('iphone');
+    }
```
NoelDeMartin, line 285: I think it's better to move these two methods to `CorePlaform`, to decouple code and avoid circular dependencies (this `AppService` is imported in too many places). It'd be ideal if we had all these type of methods in there, but I haven't done it yet.

src/core/singletons/dom.ts
```diff
@@ -532,6 +534,74 @@ export class CoreDom {
         });
     }
 
+    /**
+     * Get all source URLs and types for a video or audio.
+     *
+     * @param mediaElement Audio or video element.
+     * @returns List of sources.
+     */
+    static getMediaSources(mediaElement: HTMLVideoElement | HTMLAudioElement): CoreMediaSource[] {
```
NoelDeMartin, line 543: What do you think of moving all these new methods to a `CoreMedia` service? In fact, you're already naming the type `CoreMediaSource` instead of `CoreDOMMediaSoure`, so it goes in line with that.

src/addons/filter/mediaplugin/classes/videojs-ogvjs.ts
```diff
@@ -0,0 +1,695 @@
...
+import { CoreObject } from '@singletons/object';
+import { OGVPlayer, OGVCompat, OGVLoader } from 'ogv';
+import videojs, { PreloadOption, TechSourceObject, VideoJSOptions } from 'video.js';
+
+export const Tech = videojs.getComponent('Tech');
+
+/**
+ * OgvJS Media Controller for VideoJS - Wrapper for ogv.js Media API.
+ *
+ * Code adapted from https://github.com/HuongNV13/videojs-ogvjs.
```
NoelDeMartin, line 25: Same as I mentioned for the styles, I would remove this comment from the code and add it to the initial commit with proper attribution, co-authoring, etc.

In this case, we actually see what I was saying of the modifications because we're adding the fullscreen functionality.

src/core/singletons/object.ts
```diff
@@ -41,6 +41,42 @@ export class CoreObject {
         return JSON.stringify(a) === JSON.stringify(b);
     }
 
+    /**
+     * Object.defineProperty but "lazy", which means that the value is only set after
+     * it retrieved the first time, rather than being set right away.
+     *
+     * Code extracted from https://github.com/HuongNV13/videojs-ogvjs/blob/main/src/plugin.js
```
NoelDeMartin, line 48: Same comment about this type of documentation.

Also, I'm not sure we'll ever use this outside of VideoJS, so maybe I'd just declare this function there and not bother exporting it.

src/addons/filter/mediaplugin/classes/videojs-ogvjs.ts
```diff
@@ -0,0 +1,695 @@
...
+ */
+export class VideoJSOgvJS extends Tech {
```
NoelDeMartin, line 27: I haven't reviewed much of the initial code because I think it was almost the same as the original repository, but let me know if you prefer that I take a deeper look.

src/addons/filter/mediaplugin/mediaplugin.module.ts
```diff
@@ -26,7 +28,12 @@ import { AddonFilterMediaPluginHandler } from './services/handlers/mediaplugin';
         {
             provide: APP_INITIALIZER,
             multi: true,
-            useValue: () => CoreFilterDelegate.registerHandler(AddonFilterMediaPluginHandler.instance),
+            useValue: () => {
+                CoreFilterDelegate.registerHandler(AddonFilterMediaPluginHandler.instance);
+
+                OGVLoader.base = 'assets/lib/ogv';
+                Tech.registerTech('OgvJS', VideoJSOgvJS);
```
NoelDeMartin, line 35: Maybe this could be encapsulated in the `videojs-ovgjs` file? If not called directly, we can encapsulate it within an `initializeVideoJSOvgJS` method or something.

src/addons/filter/mediaplugin/classes/videojs-ogvjs.ts
```diff
@@ -0,0 +1,695 @@
...
+    protected triggerReady!: () => void;
+    protected on!: (name: string, callback: (e?: Event) => void) => void;
+
+    /**
+     * Create an instance of this Tech.
+     *
+     * @param options The key/value store of player options.
+     * @param ready Callback function to call when the `OgvJS` Tech is ready.
+     */
+    constructor(options: VideoJSOptions & { playerId: string }, ready: () => void) {
```
NoelDeMartin, line 76:
```suggestion
    constructor(options: VideoJSTechOptions, ready: () => void) {
```

I think it's better to extract this into a different type, even if it's only used once and not exported from this file. When looking at the code, I was confused where this `playerId` comes from and it took me a while to realize that this is actually [called internally by VideoJS](https://github.com/videojs/video.js/blob/b306ce614e70e6d3305348d1b69e1434031d73ef/src/js/player.js#L1194).

src/addons/filter/mediaplugin/classes/videojs-ogvjs.ts
```diff
@@ -0,0 +1,695 @@
...
+        return this.el_.readyState;
+    }
+
+    /**
+     * Does the player support native fullscreen mode or not. (Mobile devices)
+     *
+     * @returns Whether it supports full screen.
+     */
+    supportsFullScreen(): boolean {
+        return !!this.playerId;
```
NoelDeMartin, line 623: `playerId` is typed as a required string, so won't this always be true? (unless it's an empty string, but I'm not sure if that's possible).

src/addons/filter/mediaplugin/classes/videojs-ogvjs.ts
```diff
@@ -0,0 +1,695 @@
...
+        }
+
+        const container = player.el();
+        this.parentElement = container.parentElement;
+        if (!this.parentElement) {
+            // Shouldn't happen, it means the element is not in DOM. Do not support full screen in this case.
+            return;
+        }
+
+        this.parentElement.replaceChild(this.placeholderElement, container);
```
NoelDeMartin, line 652: Maybe you've already tried this, but I'm bringing it up just in case. Wouldn't it be possible to use `position: fixed` to work around this without the need to move the element around?

NoelDeMartin (changes requested):

src/addons/filter/mediaplugin/services/handlers/mediaplugin.ts
dpalou: They're added by LMS. LMS also uses VideoJS, and we receive the HTML already "filtered" so it includes this classes and attributes.

src/theme/theme.base.scss
dpalou: Moved to a new file.

I divided the CSS in 2 blocks: the styles specific of the app, and the ones copied from Moodle without changing anything. I also changed the link to point to the specific version of the CSS file.

src/core/services/app.ts
dpalou: I added a new commit to move these functions (and some existing ones) to CorePlatform.

src/core/singletons/dom.ts
dpalou: I like it, done!

src/addons/filter/mediaplugin/classes/videojs-ogvjs.ts
dpalou: As we talked in chat, I changed the link to point to the exact version of the file and I added the list of changes done.

src/core/singletons/object.ts
dpalou: I moved the function to the videojs-ogvjs file and removed the comment about the original code (that file already has a comment like that).

src/addons/filter/mediaplugin/classes/videojs-ogvjs.ts
dpalou: IMO there's no need to, as you say the code is almost the same. Now there's a list of changes in the class JSDoc, so you might want to review those parts.

src/addons/filter/mediaplugin/classes/videojs-ogvjs.ts
dpalou: Actually, since this playerId is added to the options by VideoJS internally (and I'm not sure if this is documented) I preferred to change the declaration to optional.

src/addons/filter/mediaplugin/classes/videojs-ogvjs.ts
dpalou: Yes, this is the first thing we tried but we didn't manage to make it work so we ended up using this solution.

NoelDeMartin (approved):

PR #3541, opened 2023-02-07
Mobile 4065

crazyserver:


src/core/singletons/dom.ts
```diff
@@ -526,10 +544,27 @@ export class CoreDom {
         });
 
         element.addEventListener('keyup', (event) => {
-            if ((event.key == ' ' || event.key == 'Enter')) {
-                callback(event);
+            if (event.key === ' ' || event.key === 'Enter') {
+                event.preventDefault();
+                event.stopPropagation();
+
+                element.click();
```
dpalou, line 551: Why are you calling element.click instead of just calling the callback?

dpalou (changes requested):

crazyserver:
Thanks Dani! Done!

dpalou (approved):

PR #3543, opened 2023-02-07
Mobile 4069

dpalou:


local_moodleappbehat/tests/behat/behat_app.php
```diff
@@ -237,13 +237,18 @@ public function i_load_more_items_in_the_app(bool $not = false) {
...
+     * @When /^I swipe to the (left|right) (within (".+") (".+") )?in the app$/
      * @param string $direction Swipe direction
+     * @param bool $hasLocator Whether a reference locator is used.
+     * @param string $text Locator text.
+     * @param string $container Container selector.
      */
-    public function i_swipe_in_the_app(string $direction) {
-        $method = 'swipe' . ucwords($direction);
+    public function i_swipe_in_the_app(string $direction, bool $hasLocator = false, string $text = '', string $container = '') {
+        $result = $this->zone_js("swipe('$direction'" . ($hasLocator ? ", { text: $text, container: $container }" : '') . ')');
```
NoelDeMartin, line 247: Instead of taking arguments individually, I think it's better to use the `parse_element_locator` method to parse it, like we're doing in other steps. This should also simplify the implementation of the `swipe` method in the JS side, by reusing the methods we already have to find elements and just use any matches to scope the angular instances search.

It's true that doing this would accept steps like `I swipe to the right within "Course" within "Courses" "ion-item" in the app`, but conversely it also accepts using things like "near" and such which may be useful. Maybe if you want to remove the potential redundancy, we can define it as  `I swipe to the right inside "..." in the app` instead or something.

NoelDeMartin (commented):
It looks good overall, but I found one thing that can be improved.

local_moodleappbehat/tests/behat/behat_app.php
dpalou: Thanks Noel, at first I didn't use it for 2 reasons: because it could look weird to have 2 "within" (as you said), and because we probably don't need as much specificity for this. But it's true that this way is more coherent with other steps, so I changed it.

I renamed the step to "`I swipe to the right in ...`" instead of **within**.

NoelDeMartin (approved):

PR #3548, opened 2023-02-15
MOBILE-4251 Feedback: Avoid displaying the preview button to students

alfonso-salces:


src/addons/mod/feedback/pages/form/form.ts
```diff
@@ -109,6 +109,12 @@ export class AddonModFeedbackFormPage implements OnInit, OnDestroy, CanLeave {
 
         await this.fetchData();
 
+        if (!this.access || this.access.isempty || (!this.access.canedititems && !this.access.canviewreports)) {
+            CoreNavigator.back();
```
dpalou, line 113: Shouldn't we display an error in this case? In LMS they display the 'nopermissions' error based on the capability that is missing, and the final message says something like:

> Sorry, but you do not currently have permissions to do that (Edit items).

We can either use the same message, or use a similar one like 'nopermissiontoaccesspage':

> You don't have permission to access this page.

dpalou (changes requested):

src/addons/mod/feedback/lang.json
```diff
@@ -23,6 +23,7 @@
     "non_respondents_students": "Non-respondent students ({{$a}})",
     "not_selected": "Not selected",
     "not_started": "Not started",
+    "nopermissiontoaccesspage": "You don't have permission to access this page.",
```
dpalou, line 26: Sorry for being a bit annoying, but since this string is generic (is not specific to feedback) IMO it'd be better to put it in src/core/lang.json so it can be reused in other places if needed.

dpalou (changes requested):

dpalou (approved):

PR #3550, opened 2023-02-16
Mobile 4065

crazyserver:


dpalou:
Hi @crazyserver , the code looks good to me but I have a question: I cannot see the relation between the following commit name and the changes done in the commit:

Add aria-role region to headings order

In the commit you just add a labelledby, so I don't know why the commit name talks about aria-role.

src/core/features/block/components/block/core-block.html
```diff
@@ -1,4 +1,5 @@
 <!-- Only render the block if it's supported. -->
-<ion-card *ngIf="loaded && componentClass && block && block.visible" class="{{class}}">
+<ion-card *ngIf="loaded && componentClass && block && block.visible" class="{{class}}" role="region"
```
crazyserver, line 2: Here it adds the role

dpalou (approved):

PR #3560, opened 2023-02-23
[4.2] Mobile 4245 - Add new config.json setting to allow to specify staging sites for testing purposes

alfonso-salces:


src/core/features/login/services/login-helper.ts
```diff
@@ -379,9 +380,29 @@ export class CoreLoginHelperProvider {
...
+     */
+    async getAvailableSites(): Promise<(CoreLoginSiteInfo | CoreLoginSiteInfoStaging)[]> {
+        const hasEnabledStagingSites = await CoreSettingsHelper.hasEnabledStagingSites();
+
+        return hasEnabledStagingSites
+            ? CoreConstants.CONFIG.sites
+            : CoreConstants.CONFIG.sites.filter(site => !this.isStagingSite(site));
+    }
+
+    isStagingSite(site: CoreLoginSiteInfo | CoreLoginSiteInfoStaging): boolean {
```
dpalou, line 404: This function doesn't have JSDoc.

src/core/features/login/services/login-helper.ts
```diff
@@ -379,9 +380,29 @@ export class CoreLoginHelperProvider {
...
      * @returns Fixed site or list of fixed sites.
+     * @deprecated 4.2.0 use CoreConstants.CONFIG.sites instead.
      */
-    getFixedSites(): string | CoreLoginSiteInfo[] {
-        return CoreConstants.CONFIG.siteurl;
+    async getFixedSites(): Promise<string | CoreLoginSiteInfo[]> {
+        const isUniqueFixedSite = await this.isUniqueFixedSite();
+
+        return isUniqueFixedSite ? CoreConstants.CONFIG.sites[0].url : this.getAvailableSites();
+    }
```
dpalou, line 389: Since this function is deprecated I'd just return either all sites or only the not-staging ones without checking the setting, that way the function can stay synchronous and we don't break existing code.

The same happens in `isFixedUrlSet`.

src/core/services/sites.ts
```diff
@@ -2107,6 +2107,18 @@ export type CoreLoginSiteInfo = {
...
+     * Is staging site.
+     */
+    staging: boolean;
+
+    /**
+     * Class to apply to site item.
+     */
+    className: string;
+};
+
```
dpalou, line 2121: Instead of creating a new type, maybe we can just add these properties in CoreLoginSiteInfo as optional? That way we don't need to do `'staging' in site`, we can just check `!!site.staging`.

Also, I guess the className can be useful for all sites, not just staging ones.

src/core/features/login/pages/reconnect/reconnect.ts
```diff
@@ -181,8 +181,12 @@ export class CoreLoginReconnectPage implements OnInit, OnDestroy {
         }
 
         this.isBrowserSSO = !this.isOAuth && CoreLoginHelper.isSSOLoginNeeded(this.siteConfig.typeoflogin);
-        this.showScanQR = CoreLoginHelper.displayQRInSiteScreen() ||
-            CoreLoginHelper.displayQRInCredentialsScreen(this.siteConfig.tool_mobile_qrcodetype);
+
+        const displayQRInCredentialsScreen = await CoreLoginHelper.displayQRInCredentialsScreen(
+            this.siteConfig.tool_mobile_qrcodetype,
+        );
```
dpalou, line 187: We usually put awaits in a separate line to make it more obvious that it has async code, but in this case you will be calling an asynchronous code that isn't needed if `CoreLoginHelper.displayQRInSiteScreen()` is true. So I'd either do:

`this.showScanQR = CoreLoginHelper.displayQRInSiteScreen() || (await ...);`

Or:

```
this.showScanQR = CoreLoginHelper.displayQRInSiteScreen();
if (!this.showScanQR) {
    this.showScanQR = await ...
}
```

src/core/features/login/services/login-helper.ts
```diff
@@ -1299,12 +1322,14 @@ export class CoreLoginHelperProvider {
      * @param qrCodeType QR Code type from public config, assuming enabled if undefined.
      * @returns Whether the QR reader should be displayed in credentials screen.
      */
-    displayQRInCredentialsScreen(qrCodeType = CoreSiteQRCodeType.QR_CODE_LOGIN): boolean {
+    async displayQRInCredentialsScreen(qrCodeType = CoreSiteQRCodeType.QR_CODE_LOGIN): Promise<boolean> {
```
dpalou, line 1325: A couple of existing functions have changed to be async, and this can break existing code. I'd add a note in upgrade.txt explaining this, something like:

* CoreLoginHelperProvider.getAddSiteRouteInfo and CoreLoginHelperProvider.displayQRInCredentialsScreen are now async.

src/core/features/login/pages/reconnect/reconnect.ts
```diff
@@ -115,7 +116,7 @@ export class CoreLoginReconnectPage implements OnInit, OnDestroy {
             this.isOAuth = site.isOAuth();
 
             // Show logo instead of avatar if it's a fixed site.
-            this.showSiteAvatar = !!this.userAvatar && !CoreLoginHelper.getFixedSites();
+            this.showUserAvatar = !!this.userAvatar && !CoreConstants.CONFIG.sites.length;
```
dpalou, line 119: Now that we can have staging and production sites in the same list, shouldn't you use `getAvailableSites` in several places where you are using `CoreConstants.CONFIG.sites` (like this one)? Maybe it no longer makes sense to deprecate `hasSeveralFixedSites`, this function could call getAvailableSites internally (and then check its length).

dpalou (changes requested):

src/core/features/settings/services/settings-helper.ts
```diff
@@ -477,6 +478,27 @@ export class CoreSettingsHelperProvider {
...
+        const modalText = enabled ? 'core.settings.confirmenablestagingsites' : 'core.settings.confirmdisablestagingsites';
+
+        try {
+            await CoreDomUtils.showConfirm(Translate.instant(modalText));
+        } catch {
+            return;
+        }
+
+        await CoreNavigator.navigate('/');
+        window.location.reload();
```
dpalou, line 499: Oh, one more thing I forgot to add. I guess you do this to reload the list of sites in the add site page. Maybe you should only do this if the user is not logged in? Because if the user has already authenticated then the add site page won't be loaded, so maybe there's no need to reload.

dpalou (changes requested):

dpalou (changes requested):

src/core/features/settings/pages/dev/dev.ts
```diff
@@ -41,6 +43,9 @@ export class CoreSettingsDevPage implements OnInit {
     pluginStylesCount = 0;
     sitePlugins: CoreSitePluginsBasicInfo[] = [];
     userToursEnabled = true;
+    stagingSitesCount = 0;
+    enableStagingSites?: boolean;
+    listenStagingSitesChanges = false;
```
dpalou, line 48: I checked if using FormBuilder and FormControl avoided this problem, but it didn't. It also triggers the onChange callback when changing the value.

I also checked how we do it in other places of the app and I found [this code](https://github.com/moodlehq/moodleapp/blob/integration/src/addons/messages/pages/settings/settings.page.ts#L55). Every time the value successfully changes we store it in a variable named "previous". In the onChange callback, if the new value is the same as the one stored in "previous" then you don't need to do anything (this happens when you revert the value).

Both solutions are similar, I prefer the one in the link above (using a "previous" or "last" variable) but it's a personal opinion, if you prefer to keep the "listen" flag feel free to do so.

dpalou (changes requested):

src/core/features/settings/pages/dev/dev.ts
alfonso-salces: Hi @dpalou!
To solve this with FormControl / FormGroup you only have to call `setValue` or `patchValue` with the options `{ emitEvent: false }`, this is an [example](https://stackblitz.com/edit/angular-uxftfs?file=src%2Fmain.ts) of what I meant.

I personally prefer work with reactive forms over template driven forms (ngModel) because of the validation outside the template and the reactivity with RxJS. I think that we are using ngModel because is similar to how AngularJS works with forms before. Anyways, in that case I will keep using ngModel over FormControl.

I will keep using the `listenStagingSitesChanges` flag.

Thanks for your review! 😄

dpalou (approved):

dpalou (approved):

PR #3565, opened 2023-02-24
Mobile 4077 - Open in browser

alfonso-salces:


src/core/features/reportbuilder/components/report-summary/report-summary.html
```diff
@@ -39,8 +39,8 @@ <h1>
         <ion-label>
             <p>
                 <ion-icon name="fas-info-circle" aria-hidden="true"></ion-icon>
-                <core-format-text [text]="'core.reportbuilder.filtersapplied' | translate: { $a: reportUrl }" [filter]="false"
-                    [clean]="true"></core-format-text>
+                <core-format-text core-link [href]="reportUrl" [text]="'core.reportbuilder.filtersapplied' | translate: { $a: reportUrl }"
+                    [filter]="false" [clean]="true"></core-format-text>
```
dpalou, line 43:
```suggestion
                <core-format-text [text]="'core.reportbuilder.filtersapplied' | translate: { $a: reportUrl }"
                    [filter]="false"></core-format-text>
```

IMO it's better to just use the link inside the text, it's better in terms of a11y (otherwise it's not clear that the text is clickable). In my suggestion I just removed the "clean=true" part, that way the link isn't removed and format-text applies the core-link directive to it.

dpalou (changes requested):

dpalou (approved):

PR #3579, opened 2023-03-14
MOBILE-4193 core: Consolidate module icons sources

NoelDeMartin:


dpalou (approved):

src/core/features/grades/services/grades-helper.ts
```diff
@@ -692,6 +692,10 @@ export class CoreGradesHelperProvider {
                     module[1],
                     CoreDomUtils.convertToElement(text).querySelector('img')?.getAttribute('src') ?? undefined,
                 );
+                row.imageIsShape = await CoreCourseModuleDelegate.moduleIconIsShape(
+                    module[1],
+                    CoreDomUtils.convertToElement(text).querySelector('img')?.getAttribute('src') ?? undefined,
```
dpalou, line 697: I think it's better to put this line in a variable and use it both in `getModuleIconSrc` and `moduleIconIsShape`. I know it's only used twice, but since it involves converting to HTML and doing a querySelector IMO it's better to do it only once.

dpalou (changes requested):
Sorry, I approved the changes too soon, without looking at the second commit ^^' I only found a minor thing.

Also, in the testing instructions of the issue, can you explain how to configure the LTI activity so it's displayed in Timeline/Calendar and also how to configure it to use a custom icon?

dpalou (approved):

PR #3581, opened 2023-03-16
Mobile 4279

crazyserver:


src/core/directives/fa-icon.ts
```diff
@@ -121,6 +128,41 @@ export class CoreFaIconDirective implements AfterViewInit, OnChanges {
...
+    }
+
+    /**
+     * Read a icons json file.
+     *
+     * @returns Promise resolved when loaded.
+     */
+    protected async readIconsJson(): Promise<void> {
+        if (CoreFaIconDirective.iconsPromise !== undefined) {
+            return;
```
dpalou, line 156: IMO this can cause problems. E.g. if the read file is in progress you will have an empty list of aliases but `readIconsJson` will resolve immediately.

IMO it's better to use a `CorePromisedValue` here. Something like:

```
protected static aliases: CorePromisedValue<Record<string, string>>;

...

protected async checkIconAlias(iconName: string): Promise<string> {
    const aliases = await CoreFaIconDirective.aliases;
    if (aliases[iconName]) {
    ...
}
```

The implementation of `readIconsJson` would be quite similar to the current one, but resolving the `CorePromisedValue` when the file has been loaded. If the file fails to load we should probably display an error in console and resolve with an empty list of aliases.

I couldn't use the suggestion tool because your PR is too big and the "Files" tab freezes.

dpalou (changes requested):

dpalou:
Besides the comment above, can you add a note in upgrade.txt specifying that Font-Awesome has been updated? IMO it'd be nice to specify that old icon names will still work but it's recommended to update them to the new names.

crazyserver:
Thanks Dani! Solved!

src/core/directives/fa-icon.ts
```diff
@@ -121,6 +135,45 @@ export class CoreFaIconDirective implements AfterViewInit, OnChanges {
...
+
+        return iconName;
+    }
+
+    /**
+     * Read a icons json file.
+     *
+     * @returns Promise resolved when loaded.
+     */
+    protected static async readIconsJson(): Promise<Record<string, string>> {
```
dpalou, line 161: Isn't "protected static" a bit weird? Shouldn't it be just "static"?

dpalou, line 161: Also, I think it would be better to name this function "getIconsAliases" or something like this, it's easier to understand what it does.

src/core/directives/fa-icon.ts
```diff
@@ -121,6 +135,45 @@ export class CoreFaIconDirective implements AfterViewInit, OnChanges {
...
+        if (CoreFaIconDirective.aliases !== undefined) {
+            return CoreFaIconDirective.aliases;
+        }
+
+        CoreFaIconDirective.aliases = new CorePromisedValue();
+
+        const request = await Http.get('assets/fonts/font-awesome/aliases.json', {
```
dpalou, line 168: I think you can specify the return type: `Http.get<Record<string, string>>`. That way you don't need to cast it later.

dpalou, line 168: IMO this variable should be called `aliases`, you're doing an await so the variable will hold the result of the request, it's not the request promise.

src/core/directives/fa-icon.ts
```diff
@@ -121,6 +135,45 @@ export class CoreFaIconDirective implements AfterViewInit, OnChanges {
...
+            responseType: 'json',
+        }).toPromise();
+
+        CoreFaIconDirective.aliases.resolve(request as Record<string, string>);
```
dpalou, line 172: I know it shouldn't happen, but if the read file fails the `CorePromisedValue` will never be resolved. I think it'd be better to have a `catch`, and if the request fails then we can resolve the aliases with an empty object.

src/core/directives/fa-icon.ts
```diff
@@ -121,6 +135,45 @@ export class CoreFaIconDirective implements AfterViewInit, OnChanges {
...
+
+        return CoreFaIconDirective.aliases;
```
dpalou, line 174: Here you already have the list of aliases (because you did an `await`), so I'd just return them instead of returning the `CorePromisedValue`.

dpalou (changes requested):

dpalou (approved):

PR #3582, opened 2023-03-16
MOBILE-4250 course: Remove module name from course module info component

crazyserver:


src/core/components/mod-icon/mod-icon.ts
```diff
@@ -31,7 +31,8 @@ const fallbackModName = 'external-tool';
 })
 export class CoreModIconComponent implements OnInit, OnChanges {
 
-    @Input() modname?: string; // The module name. Used also as component if set.
+    @Input() modname = ''; // The module name. Used also as component if set.
+    @Input() fallbackTranslation = ''; // The module name. Used also as component if set.
```
dpalou, line 35: This comment is wrong.

src/core/features/course/services/course.ts
```diff
@@ -1509,17 +1509,17 @@ export class CoreCourseProvider {
...
-        }
-
+    translateModuleName(moduleName: string, fallback?: string): string {
         const langKey = 'core.mod_' + moduleName;
         const translated = Translate.instant(langKey);
 
-        return translated !== langKey ? translated : moduleName;
+        return translated !== langKey ?
+            translated :
+            (fallback ?? moduleName);
```
dpalou, line 1522: I think it's better to do `fallback || moduleName` instead. In the mod-icon component you're initializing `fallbackTranslation` to an empty string, so an empty string will be used instead of `moduleName`.

dpalou (changes requested):

crazyserver:
Thanks Dani! Changed!

dpalou (approved):

PR #3583, opened 2023-03-20
MOBILE-4188: Implement teacher gradebook

NoelDeMartin:


src/core/features/grades/services/handlers/course-participants-option.ts
```diff
@@ -0,0 +1,102 @@
...
+    async prefetch(course: CoreCourseAnyCourseData): Promise<void> {
+        let offset = 0;
+        let canLoadMore = true;
+
+        do {
+            const result = await CoreUser.getParticipants(course.id, offset, undefined, undefined, true);
+
+            for (const participant of result.participants) {
+                await CoreUtils.ignoreErrors(CoreGrades.getCourseGradesTable(course.id, participant.id, undefined, true));
+            }
```
dpalou, line 93: Each iteration of this `for` will block the next one, IMO it'd be better to run this in parallel using a `Promise.all`.

Also, did you talk with Juan if we want to prefetch ALL pages? It can be a lot of WS requests.

dpalou (changes requested):

src/core/features/grades/services/grades.ts
```diff
@@ -388,6 +389,19 @@ export class CoreGradesProvider {
...
+     * Check whether the current user can view all the grades in the course.
+     *
+     * @param courseId Course id.
+     * @returns Whether the current user can view all the grades.
+     */
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    async canViewAllGrades(courseId: number): Promise<boolean> {
+        // @todo MDL-77585: implement method.
+
+        return CoreConstants.CONFIG.isTeacher ?? true;
```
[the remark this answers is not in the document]
NoelDeMartin: Yes, that's it.

The thing is that I wanted to get this integrated because we wanted to test some things before moving this to testing, in particular whether to show the search box by default or not. This should never end up like this in the release, so as long as it doesn't affect other issues (which I think it doesn't), I thought this would be the best default to allow us testing the new page.

But I guess I can also compile and apk myself and send it around, if we only want it for testing :shrug:

src/core/features/grades/services/handlers/course-option.ts
```diff
@@ -35,13 +35,16 @@ export class CoreGradesCourseOptionHandlerService implements CoreCourseOptionsHa
     /**
      * @inheritdoc
      */
-    invalidateEnabledForCourse(courseId: number, navOptions?: CoreCourseUserAdminOrNavOptionIndexed): Promise<void> {
+    async invalidateEnabledForCourse(courseId: number, navOptions?: CoreCourseUserAdminOrNavOptionIndexed): Promise<void> {
         if (navOptions && navOptions.grades !== undefined) {
             // No need to invalidate anything.
-            return Promise.resolve();
+            return;
```
dpalou, line 41: Before this patch, if `navOptions.grades` was defined we didn't need to invalidate user courses because `showGradebook` doesn't call `getUserCourse` in that case (it's done in `isPluginEnabledForCourse`). But we still need to call `invalidateCourseGradesPermissionsData`, because `canViewAllGrades` is used even if `grades` is defined.

So IMO we should aways call `invalidateCourseGradesPermissionsData` (we could do it only if `navOptions.grades` is undefined or true, but this WS barely returns data so IMO it's ok to always do it). And we should only call `invalidateUserCourses` is if `navOptions.grades` is undefined.

src/core/features/grades/services/handlers/course-participants-option.ts
```diff
@@ -0,0 +1,81 @@
...
+export class CoreGradesCourseParticipantsOptionHandlerService implements CoreCourseOptionsHandler {
+
+    name = 'CoreGradesParticipants';
+    priority = 400;
+
+    /**
+     * @inheritdoc
+     */
+    invalidateEnabledForCourse(courseId: number): Promise<void> {
+        return CoreGrades.invalidateCourseGradesPermissionsData(courseId);
```
dpalou, line 39: This function should be implemented just like the one in `CoreGradesCourseOptionHandlerService`, because the implementation of `isEnabledForCourse` is almost the same.

src/core/features/user/pages/participants/participants.page.ts
```diff
@@ -39,12 +41,11 @@ export class CoreUserParticipantsPage implements OnInit, AfterViewInit, OnDestro
     searchQuery: string | null = null;
     searchInProgress = false;
     searchEnabled = false;
-    showSearchBox = false;
     fetchMoreParticipantsFailed = false;
 
     @ViewChild(CoreSplitViewComponent) splitView!: CoreSplitViewComponent;
 
-    constructor() {
+    constructor(protected route: ActivatedRoute) {
```
dpalou, line 48: This route isn't used, right?

dpalou (changes requested):

NoelDeMartin:
Thanks @dpalou for reviewing, in addition to fixing what you mentioned, I've also amended the behat tests to actually test the new functionality. So now they should really fail until MDL-77585 is released in the weekly.

PS: Sorry about the noise with the force-pushes, I messed up doing some changes but it should be ok now :sweat_smile:

src/core/features/grades/tests/behat/navigation.feature
```diff
@@ -278,19 +277,42 @@ Feature: Grades navigation
     Given I entered the course "Course 2" as "teacher1" in the app
     And I change viewport size to "1200x640"
 
-    # Course grades
-    When I press "Participants" in the app
+    # User grades
```
dpalou, line 280: These new instructions should be in a Scenario with @lms_from4.2, since they won't work in older sites.

Also, I think it would be nice to keep the old instructions to test that teachers are still able to view the users grades from their profile.

dpalou (changes requested):

dpalou (approved):

PR #3590, opened 2023-03-24
MOBILE-4203 timezone: Convert 99 to a readable timezone

dpalou:


src/core/features/user/tests/behat/basic_usage.feature
```diff
@@ -59,3 +59,36 @@ Feature: Test basic usage of user features
...
+      | username | firstname | lastname | timezone      |
+      | student2 | John      | Smith    | Asia/Shanghai |
+    And the following "courses" exist:
+      | fullname | shortname |
+      | Course 1 | C1        |
+    And the following "course enrolments" exist:
+      | user     | course | role    |
+      | student1 | C1     | student |
+      | student2 | C1     | student |
+    # This capability override is needed until MDL-77468 is integrated.
```
NoelDeMartin, line 78:
```suggestion
    # @todo MDL-77468 This capability override is needed until the issue is integrated.
```

NoelDeMartin (changes requested):

src/core/features/user/pages/about/about.ts
```diff
@@ -120,6 +121,12 @@ export class CoreUserAboutPage implements OnInit, OnDestroy {
 
             this.user.address = CoreUserHelper.formatAddress('', user.city, user.country);
 
+            const defaultTimezone = CoreSites.getCurrentSite()?.getStoredConfig('timezone');
+            this.displayTimezone = !!defaultTimezone;
+            if (this.displayTimezone && this.user.timezone === '99') {
```
NoelDeMartin, line 126: What does "99" mean? Even if it's only used here, I think we should extract it to a constant, helper method, or even add some comment explaining it.

Other than that, I don't understand why we only display the timezone if there is a default one. If a user has a timezone set, shouldn't we show it regardless of the site's default?

dpalou: I moved it to a constant. I left it hardcoded because it was only used here and LMS doesn't use a constant for this (I thought maybe there was a reason for this, but I guess there is none). '99' means that the user wants to use the server's timezone instead of forcing a different one.

We decided to display the timezone only if we know the value of the server's timezone, because otherwise 99 was displayed and it looks weird. With this commit we are translating '99' to a readable value like "Australia/Perth", and we can do that thanks to MDL-77643. If we don't know the server's timezone then we only have 2 options: display "99", or only display the timezone if the user's timezone is not the server timezone (which can be weird because you would see it in some users but not in others). We didn't like any of those options, so we decided to display it only in 4.2+ sites.

NoelDeMartin (approved):

PR #3595, opened 2023-03-31
MOBILE-4292 multilang2: Create new multilang filter

alfonso-salces:


src/addons/filter/multilang2/services/handlers/multilang2.ts
```diff
@@ -0,0 +1,114 @@
...
+    /**
+     * This function filters the received text based on the language
+     * tags embedded in the text, and the current user language or
+     * 'other', if present.
+     *
+     * @param text The text to filter.
+     * @returns string The filtered text for this multilang block.
+     */
+    async filter(text: string): Promise<string> {
+        const currentLang = await CoreLang.getCurrentLanguage();
```
dpalou, line 45: This line should be after the if guard, there is no need to obtain the current language if the text doesn't include "mlang".

src/addons/filter/multilang2/services/handlers/multilang2.ts
```diff
@@ -0,0 +1,114 @@
...
+        if (text.indexOf('mlang') === -1) {
+            return text;
+        }
+
+        this.replacementDone = false;
+
+        if (!this.parentCache[currentLang] && CoreLang.getParentLanguage(currentLang)) {
+            const parentlangs = CoreLang.getParentLanguage(currentLang);
+            this.parentCache[currentLang] = parentlangs;
+        }
```
dpalou, line 56: In the app, getting the parent language is fast because there is only 1 and the translations table is already loaded. In LMS this function is recursive and it might require loading more languages, I guess that's why they use a memory cache. My point is that we probably don't need to use a memory cache in the app, I would either:

1- Call `getParentLanguage` in `replaceLangs`. Better readability (less arguments), but it can be done a lot of times.
2- Store it in a local variable (`const parentLanguage`) and pass it as a parameter to `replaceLangs`. This way it will only be done once per text.

I would go for option 2, but feel free to use 1 if you prefer it.

src/addons/filter/multilang2/services/handlers/multilang2.ts
```diff
@@ -0,0 +1,114 @@
...
+
+        const search = /{\s*mlang\s+((?:[a-z0-9_-]+)(?:\s*,\s*[a-z0-9_-]+\s*)*)\s*}(.*?){\s*mlang\s*}/g;
```
dpalou, line 58: Please use the following flags along with "g":

- Flag "i" to make the regex case insensitive. In PHP they use the flag "i" too.
- Flag "m" to allow multiline matches. In PHP they use the flag "s", I guess they're equivalent.

src/addons/filter/multilang2/services/handlers/multilang2.ts
```diff
@@ -0,0 +1,114 @@
...
+        const result = text.replace(
+            search,
+            (substring, language, content) => this.replaceLangs(currentLang, [substring, language, content]),
```
dpalou, line 61:
```suggestion
            (subString, language, content) => this.replaceLangs(currentLang, [subString, language, content]),
```

Changed the name of a variable to make it camel case.

src/addons/filter/multilang2/services/handlers/multilang2.ts
```diff
@@ -0,0 +1,114 @@
...
+            return text; // Error during regex processing, keep original text.
+        }
+
+        if (this.replacementDone) {
+            return result;
+        }
+
+        const result2 = text.replace(
+            search,
+            (substring, language, content) => this.replaceLangs('other', [substring, language, content]),
```
dpalou, line 74:
```suggestion
            (subString, language, content) => this.replaceLangs('other', [subString, language, content]),
```

src/addons/filter/multilang2/services/handlers/multilang2.ts
```diff
@@ -0,0 +1,114 @@
...
+     * specified filtering language, it returns the text of the
+     * block. Otherwise it returns an empty string.
+     *
+     * @param replacelang A string that specifies the language used to filter the matches.
+     * @param langblock An array containing the matching captured pieces of the
+     * regular expression. They are the languages of the tag, and the text associated with those languages.
+     *
+     * @returns replaced string.
+     */
+    protected replaceLangs(replacelang: string, langblock: string[]): string {
```
dpalou, line 92: There are a few variables in this function that don't use camel case: replacelang, langblock, blocklangs, blocktext, etc.

src/addons/filter/multilang2/services/handlers/multilang2.ts
```diff
@@ -0,0 +1,114 @@
...
+        // Normalize languages.
+        const blocklangs = langblock[1].replace(/-|_/g, '').toLowerCase().split(',');
+        const blocktext = langblock[2];
```
dpalou, line 95:
```suggestion
        const blocklangs = (langblock[1] ?? '').replace(/-|_/g, '').toLowerCase().split(',');
        const blocktext = langblock[2] ?? '';
```

Right now this isn't needed, but IMO this way the code won't crash if the code changes and this parameters are no longer supplied, because TS wouldn't detect that.

src/addons/filter/multilang2/services/handlers/multilang2.ts
```diff
@@ -0,0 +1,114 @@
...
+     *
+     * @param replacelang A string that specifies the language used to filter the matches.
+     * @param langblock An array containing the matching captured pieces of the
+     * regular expression. They are the languages of the tag, and the text associated with those languages.
+     *
+     * @returns replaced string.
+     */
+    protected replaceLangs(replacelang: string, langblock: string[]): string {
+        // Normalize languages.
+        const blocklangs = langblock[1].replace(/-|_/g, '').toLowerCase().split(',');
```
dpalou, line 94: Why are you removing both "-" and "_"? In PHP they replace "-" with "_" because the Moodle langs use "_", but in the app we use "-" so I think we should do the opposite: replace "_" with "-":

`.replace(/_/g, '-')`

That way if something uses something like "en_us" it will be changed to "en-us".

dpalou, line 94: Also, in PHP they also remove whitespaces, I think you should do it too:

`.replace(/ /g, '')`

src/addons/filter/multilang2/services/handlers/multilang2.ts
```diff
@@ -0,0 +1,114 @@
...
+        const blocktext = langblock[2];
+        const parentlangs = this.parentCache[replacelang];
+
+        for (const blocklang of blocklangs) {
+            /* We don't check for empty values of blocklang as they simply don't
+             * match any language and they don't produce any errors or warnings.
+             */
+            if (blocklang === replacelang || parentlangs?.includes(blocklang)) {
```
dpalou, line 102:
```suggestion
            if (blockLang === replaceLang || parentLangs === blockLang) {
```

In the app we'll only have 1 parent lang, it will be something like "en", and it should exactly match the value.

dpalou (changes requested):

dpalou:
Besides the comments above, a change was done in LMS to return the data already formatted: [MDL-77726](https://tracker.moodle.org/browse/MDL-77726). I think we should also adapt the code in this issue to use `displayvalue` instead of `value` if it exists. Please do it in a separate commit :)

dpalou (changes requested):

dpalou (approved):

PR #3597, opened 2023-04-04
MOBILE-4187 app: Add url class on HTML tag

crazyserver:


src/app/app.component.ts
```diff
@@ -253,14 +270,55 @@ export class AppComponent implements OnInit, AfterViewInit {
...
+
+    /**
+     * Converts the provided URL into a CSS class that be used within the page.
+     * This is primarily used to add the siteurl to the body tag as a CSS class.
+     * Extracted from LMS url_to_class_name function.
+     *
+     * @param url Url.
+     * @returns Class name
+     */
+    protected urlToClassName(url: string): string {
```
NoelDeMartin, line 291: I'm not sure if this method should be declared here, in fact most of the existing code in this file should probably be moved elsewhere (maybe an initializer or two? as many as responsibilities).

I say that for two reasons. First to avoid bloating this file given that it's the entry point to the app (so it should be as simple as possible). And second, because the `<html>` element really sits above the rendered App component, so it's outside its responsibility.

NoelDeMartin (commented):
I spoted some things to improve, but overall I think it's fine as it is and could be merged.

src/app/app.component.ts
```diff
@@ -253,14 +270,55 @@ export class AppComponent implements OnInit, AfterViewInit {
      *
      * @param prefix Prefix of to the class.
```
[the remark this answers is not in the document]
crazyserver: ok

src/app/app.component.ts
crazyserver: We can move protected (and static)  functions to dom singleton? However, I preffer not to move too much code and I would leave it here right now.

About where the code comes from, it has been already documented on urlToClassName function.

src/app/app.component.ts
```diff
@@ -113,31 +116,45 @@ export class AppComponent implements OnInit, AfterViewInit {
                 const info = site.getInfo();
                 if (info) {
                     // Add version classes to body.
-                    this.removeVersionClass(MOODLE_VERSION_PREFIX);
+                    this.removeModeClasses(MOODLE_VERSION_PREFIX);
+                    this.removeModeClasses(MOODLE_SITE_URL_PREFIX);
+
                     this.addVersionClass(MOODLE_VERSION_PREFIX, CoreSites.getReleaseNumber(info.release || ''));
+                    this.addSiteUrlClass(info.siteurl);
```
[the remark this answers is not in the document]
crazyserver: Change the function.

src/app/app.component.ts
NoelDeMartin: Well it doesn't need to be protected if we move it elsewhere, but ok let's leave it as is.

NoelDeMartin (approved):

PR #3607, opened 2023-04-13
MOBILE-4291 AddonUserProfile: Add margin-right in checkbox

alfonso-salces:


src/addons/userprofilefield/checkbox/component/addon-user-profile-field-checkbox.html
```diff
@@ -13,8 +13,8 @@
 
 <!-- Edit. -->
 <ion-item *ngIf="edit && field && field.shortname && form" [formGroup]="form">
-    <ion-label>
-        <span class="label-text" [core-mark-required]="required">{{ field.name }}</span>
+    <ion-label class="label">
```
crazyserver, line 16: I think you don't need any class to change styles here. This is optional amend.

src/addons/userprofilefield/checkbox/component/checkbox.scss
```diff
@@ -2,8 +2,16 @@
...
 
-    .label-text {
-        text-overflow: var(--text-overflow);
-        white-space: var(--white-space);
+    .item-heading {
+        white-space: normal;
+    }
+
+    .label {
+        padding-right: .5rem;
```
crazyserver, line 10: Please beware of RTL you can use IIRC @include padding-horizontal(null, .5rem); (you should @import "~theme/globals"; first).

crazyserver (changes requested):

crazyserver (approved):

PR #3608, opened 2023-04-13
MOBILE-4277 course: Show modules in course index

NoelDeMartin:
Including the ones that were previously hidden, such as text and label.

### DIFF
--- a/src/core/features/course/components/course-index/course-index.ts
+++ b/src/core/features/course/components/course-index/course-index.ts
@@ -18,9 +18,10 @@ import {
     CoreCourseModuleCompletionTracking,
     CoreCourseProvider,
 } from '@features/course/services/course';
-import { CoreCourseHelper, CoreCourseSection } from '@features/course/services/course-helper';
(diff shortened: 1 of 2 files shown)


dpalou (approved):

src/core/features/course/components/course-index/course-index.ts
```diff
@@ -79,7 +79,7 @@ export class CoreCourseCourseIndexComponent implements OnInit {
             .filter((section) => !CoreCourseHelper.isSectionStealth(section))
             .map((section) => {
                 const modules = section.modules
-                    .filter((module) => !CoreCourseHelper.isModuleStealth(module, section) && !module.noviewlink)
+                    .filter((module) => !CoreCourseHelper.isModuleStealth(module, section))
```
dpalou, line 82: According to the issue description, we should only display labels in course index in 4.2+ sites.

dpalou (changes requested):

dpalou (approved):

PR #3611, opened 2023-04-18
MOBILE-4292 base-profilefield: Translate fields to display

alfonso-salces:


src/core/features/user/classes/base-profilefield-component.ts
```diff
@@ -44,11 +46,23 @@ export abstract class CoreUserProfileFieldBaseComponent implements OnInit {
...
      */
-    ngOnInit(): void {
+    async ngOnInit(): Promise<void> {
         if (!this.field) {
             return;
         }
 
+        if (this.field.name) {
+            this.field.name = await AddonFilterMultilangHandler.filter(<string> this.field.name);
+            this.field.name = await AddonFilterMultilang2Handler.filter(<string> this.field.name);
```
dpalou, line 56: IMO we should try to avoid injecting these services directly, it's best to use the `core-format-text` directive in the template where this field is rendered since it will also check if filters are enabled or not.

If the WebService isn't filtering the data you need to add `[wsNotFiltered]="true"` to `core-format-text`. Please notice the `displayvalue` is filtered by the WS but `value` isn't, so you should probably do something like:

```
protected initForNonEdit(field: CoreUserProfileField): void {
    this.value = field.displayvalue ?? field.value;
    this.wsNotFiltered = field.displayvalue !== undefined && field.displayvalue !== null;
}
```

dpalou, line 56: Related to the comment above, I just saw that the `displayvalue` field is wrongly typed, it should be optional because we won't receive it in 4.1 and older sites:

`displayvalue?: string; // @since 4.2. Formatted value of the custom field.`

Looking at the Moodle code I think it will never be null, but in the code of the first comment I put null too to make it coherent with the assignation.

dpalou (changes requested):

dpalou (changes requested):

dpalou:
Just for the record, I saw that we still use the field.name without filtering in some places (like placeholders). However, this only happens in signup page and the signup WS already filters the name, so this issue won't affect the users. I opened a MDL issue to fix this in the WebServices: https://tracker.moodle.org/browse/MDL-77967

src/addons/userprofilefield/checkbox/component/addon-user-profile-field-checkbox.html
```diff
@@ -1,7 +1,11 @@
 <!-- Render (no edit). -->
 <ion-item *ngIf="!edit && field && field.name">
     <ion-label>
-        <p class="item-heading">{{ field.name }}</p>
+        <p class="item-heading">
```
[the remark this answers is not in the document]
alfonso-salces: I just forgot add it to this component.
Nice catch!

dpalou (approved):

PR #3618, opened 2023-04-19
MOBILE-4284: Implement custom question numbers in quiz

NoelDeMartin:


src/core/features/question/services/database/question.ts
```diff
@@ -41,18 +42,6 @@ export const QUESTION_SITE_SCHEMA: CoreSiteSchema = {
                     type: 'INTEGER',
                     notNull: true,
                 },
-                {
-                    name: 'componentid',
```
dpalou, line 45: IMO it's ok to remove the question number and the userid, but I'm not sure about componentid. This value lets us know to which quiz do the questions belong. Right now it's not used because we always retrieve the questions from quiz (and we know the attempt ID in there), but if we ever need to do this the opposite way we won't be able to do that because the WSs that let us know the quiz ID from the attempt ID have restrictions (e.g. mod_quiz_get_attempt_review only works for finished attempts).

Actually there is a workaround now, because the answers stored in the app's DB have the componentid too. But please see my next comment because maybe we remove it from there too for coherence.

src/core/features/question/services/database/question.ts
```diff
@@ -41,18 +42,6 @@ export const QUESTION_SITE_SCHEMA: CoreSiteSchema = {
...
-                    type: 'INTEGER',
-                },
-                {
-                    name: 'userid',
```
dpalou, line 49: I just saw that the question answers also have the userid, shouldn't we remove it from there too?

src/core/features/question/services/question.ts
```diff
@@ -314,7 +314,7 @@ export class CoreQuestionProvider {
      * @returns Question component ID.
      */
     getQuestionComponentId(question: CoreQuestionQuestionParsed, componentId: string | number): string {
-        return componentId + '_' + question.number;
+        return componentId + '_' + question.questionnumber;
```
dpalou, line 317: Damn, I missed this case when reviewing the impact of changing the number.

This is a breaking change, this component ID is used to retrieve the files stored in offline for essay questions. If it changes, the files won't be found. I guess there is no way to keep the old behaviour because the WS no longer returns the number as it used to.

If I understand it correctly, if the teacher changes the "number" displayed to the user this value will change, right? In that case maybe we should stop using the question number and use the question ID instead because it shouldn't change. However, doing this means that the user will lose the offline files when updating the app, we might need to change the folder names when upgrading the app to match the new values. We can discuss this in chat.

src/core/features/question/services/question.ts
```diff
@@ -594,14 +594,16 @@ export type CoreQuestionQuestionWSData = {
     lastactiontime?: number; // The timestamp of the most recent step in this question attempt.
     hasautosavedstep?: boolean; // Whether this question attempt has autosaved data.
     flagged: boolean; // Whether the question is flagged or not.
-    // eslint-disable-next-line id-blacklist
-    number?: number; // Question ordering number in the quiz.
+    questionnumber?: string; // Question ordering number in the quiz.
```
dpalou, line 597:
```suggestion
    questionnumber?: string; // @since 4.2. Question ordering number in the quiz.
```

dpalou (changes requested):

src/core/features/question/services/question.ts
NoelDeMartin: After discussing this, we've reached the conclusion that it's ok to leave this as is.

The main issue is that the question doesn't have any identification (question ID doesn't exist), so the only thing we can use to identify a question is the `questionnumber`. It is true that it can change, but in the way it's implemented now in the LMS once anyone has started an attempt the questions cannot be updated anymore. So it shouldn't be a problem to use it as identification, because downloading a question starts an attempt.

dpalou (approved):

PR #3626, opened 2023-04-20
MOBILE-4288: Filter multilang group strings

NoelDeMartin:


src/addons/filter/filter.module.ts
```diff
@@ -25,7 +25,6 @@ import { AddonFilterGlossaryModule } from './glossary/glossary.module';
 import { AddonFilterMathJaxLoaderModule } from './mathjaxloader/mathjaxloader.module';
 import { AddonFilterMediaPluginModule } from './mediaplugin/mediaplugin.module';
 import { AddonFilterMultilangModule } from './multilang/multilang.module';
-import { AddonFilterMultilang2Module } from './multilang2/multilang2.module';
```
dpalou, line 28: In the app, each folder in addons/filter correspond to a Moodle plugin. With this change, we're mixing 2 plugins in the same addon, which I think it's weird. IMO it's better to have 2 separate folders, even if there's more boilerplate, and we can create an utils function to filter some text:

```
let filtered = await AddonFilterMultilangHandler.filter(text as unknown as string);

filtered = await AddonFilterMultilang2Handler.filter(filtered);
```

src/core/components/group-selector/group-selector.ts
```diff
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-import { Component, Input, Output, EventEmitter, ChangeDetectionStrategy } from '@angular/core';
+import { AddonFilterMultilang } from '@addons/filter/multilang/services/multilang';
+import { Component, Input, Output, EventEmitter, ChangeDetectionStrategy, OnInit, OnChanges, SimpleChanges, ChangeDetectorRef } from '@angular/core';
```
dpalou, line 16: There's a warning about line length in here.

dpalou (changes requested):

PR #3629, opened 2023-04-20
Mobile 4248

alfonso-salces:


src/addons/mod/data/components/action/action.ts
```diff
@@ -139,4 +142,51 @@ export class AddonModDataActionComponent implements OnInit {
...
+                action: async () => {
+                    this.viewEntry();
+                    await PopoverController.dismiss();
+                },
+                text: 'addon.mod_data.more',
+                icon: 'fas-magnifying-glass-plus',
+            });
+        }
+
+        await CoreDomUtils.openPopover<AddonModDataActionsMenuItem>({
```
dpalou, line 184:
```suggestion
        await CoreDomUtils.openPopover({
```

The type passed to `openPopover` is the type of the value passed to `PopoverController.dismiss()`. But you aren't passing any value when dismissing the popover, so you don't need to define this type (it's void by default).

src/addons/mod/data/components/action/action.ts
```diff
@@ -139,4 +142,51 @@ export class AddonModDataActionComponent implements OnInit {
         CoreEvents.trigger(AddonModDataProvider.ENTRY_CHANGED, { dataId: dataId, entryId: entryId }, this.siteId);
     }
 
+    async actionsMenu(): Promise<void> {
+        const items = [
+            {
+                action: async () => {
+                    this.editEntry();
+                    await PopoverController.dismiss();
```
dpalou, line 150: Since all actions call `PopoverController.dismiss`, you could do this in the `AddonModDataActionsMenuComponent`. E.g. clicking an item would call a function defined in this component (instead of calling `item.action()` directly), and this function would do the dismiss and call `item.action()`.

dpalou (changes requested):

src/addons/mod/data/components/action/action.ts
```diff
@@ -139,4 +143,65 @@ export class AddonModDataActionComponent implements OnInit {
...
     }
 
+    /**
+     * Open actions menu popover.
+     */
+    async actionsMenu(): Promise<void> {
+        const access = await AddonModData.getDatabaseAccessInformation(this.database.id, {
+            cmId: this.database.coursemodule,
+            groupId: this.group,
+        });
```
dpalou, line 152: I think it's better to pass this info as an Input because we already have it both in the list page and the entry page. This is a bit tricky, because these components are rendered dynamically from some HTML text.

I think you'll need to add something like this in `displayShowFields`, in the addon-mod-data-action part:

`[access]="access"`

And then you need to add this access data to the JS data passed to the compilation: [here](https://github.com/moodlehq/moodleapp/blob/main/src/addons/mod/data/components/index/index.ts#L371) and [here](https://github.com/moodlehq/moodleapp/blob/main/src/addons/mod/data/pages/entry/entry.ts#L211).

src/addons/mod/data/components/action/action.ts
```diff
@@ -139,4 +143,65 @@ export class AddonModDataActionComponent implements OnInit {
...
+                !this.entry.approved
+                    ? {
+                        action: () => this.approveEntry(),
+                        text: 'addon.mod_data.approve',
+                        icon: 'fas-thumbs-up',
+                    }
+                    : {
+                        action: () => this.disapproveEntry(),
+                        text: 'addon.mod_data.disapprove',
+                        icon: 'fas-rotate-left',
```
dpalou, line 185: IMO this icon should be far-thumbs-down.

src/addons/mod/data/components/action/action.ts
```diff
@@ -139,4 +143,65 @@ export class AddonModDataActionComponent implements OnInit {
...
+                    },
+            );
+        }
+
+        if (this.mode === AddonModDataTemplateMode.LIST) {
+            items.unshift({
+                action: () => this.viewEntry(),
+                text: 'addon.mod_data.showmore',
+                icon: 'far-thumbs-down',
```
dpalou, line 194: IMO this icon should be fas-magnifying-glass-plus.

src/addons/mod/data/components/action/action.ts
```diff
@@ -139,4 +142,65 @@ export class AddonModDataActionComponent implements OnInit {
...
+        const items: AddonModDataActionsMenuItem[] = [];
+
+        if (this.entry.canmanageentry) {
+            items.push(
+                {
+                    action: () => this.deleteEntry(),
+                    text: 'core.delete',
+                    icon: 'fas-trash',
+                },
+            );
```
dpalou, line 163: If an entry is marked as deleted we should display a "restore" option instead of delete, like it's done [in here](https://github.com/moodlehq/moodleapp/blob/main/src/addons/mod/data/components/action/addon-mod-data-action.html#L14).

src/addons/mod/data/components/actionsmenu/actionsmenu.ts
```diff
@@ -0,0 +1,41 @@
...
+     * @param item item from which the action will be executed.
+     */
+    async onItemClick(item: AddonModDataActionsMenuItem): Promise<void> {
+        item.action();
+        await PopoverController.dismiss();
+    }
+
+}
+
+export interface AddonModDataActionsMenuItem { text: string; icon: string; action: () => void };
```
dpalou, line 41: IMO it's better to put this in several lines to improve readability.

src/addons/mod/data/services/data-helper.ts
```diff
@@ -426,6 +427,10 @@ export class AddonModDataHelperProvider {
             approvalstatus: database.approval,
             comments: database.comments,
 
+            actionsmenu: (entry.canmanageentry && !entry.deleted)
```
dpalou, line 430: IMO the `entry.deleted` part isn't needed, because if you can manage an entry you will always see an option: delete or restore.

dpalou (changes requested):

src/addons/mod/data/components/index/index.ts
```diff
@@ -95,6 +95,7 @@ export class AddonModDataIndexComponent extends CoreCourseModuleMainActivityComp
         database: AddonModDataData;
         title: string;
         group: number;
+        access: AddonModDataGetDataAccessInformationWSResponse | undefined;
```
dpalou, line 98: The "access" input in AddonModDataActionComponent has been defined as mandatory, so this field shouldn't allow undefined. If you allow undefined here, then the input should be optional as well.

The same happens in entry.ts.

dpalou (changes requested):

dpalou (approved):

PR #3632, opened 2023-04-20
Mobile 4215

dpalou:


package.json
```diff
@@ -89,29 +89,28 @@
...
     "cordova-plugin-add-swift-support": "2.0.2",
     "cordova-plugin-advanced-http": "3.3.1",
     "cordova-plugin-androidx-adapter": "1.1.3",
-    "cordova-plugin-badge": "0.8.8",
     "cordova-plugin-camera": "6.0.0",
     "cordova-plugin-chooser": "1.3.2",
     "cordova-plugin-customurlscheme": "5.0.2",
     "cordova-plugin-device": "2.1.0",
     "cordova-plugin-file": "6.0.2",
+    "cordova-plugin-file-opener2": "4.0.0",
```
NoelDeMartin, line 103: Given that we're using the external plugin again, should we update the README of `@moodlehq/cordova-plugin-file-opener` to indicate that we're using the other plugin again?

Which is a bit confusing, because in their README they say it's not maintained and incompatible with Android 11 :/. But I guess you've tested it and it works, so I guess it just means they didn't update their README.

src/core/services/local-notifications.ts
```diff
@@ -71,9 +71,8 @@ export class CoreLocalNotificationsProvider {
     async initialize(): Promise<void> {
         await CorePlatform.ready();
 
-        if (!this.isAvailable()) {
-            return;
-        }
+        // Request permission when the app starts.
+        LocalNotifications.requestPermission();
```
NoelDeMartin, line 75: Does this mean we'll request the permissions when the app is opened, even before logging in to a site? Why is this necessary now, and how did it work before?

I think it'd be better to ask when necessary instead; and showing a message that explains why we need the permission before requesting it. Also, what about sites that don't have any calendar events? We probably don't need to request the permission at all.

NoelDeMartin (changes requested):

package.json
```diff
@@ -73,13 +73,13 @@
     "@ionic-native/web-intent": "5.36.0",
     "@ionic-native/zip": "5.36.0",
     "@ionic/angular": "5.9.4",
-    "@moodlehq/cordova-plugin-file-opener": "3.0.5-moodle.1",
+    "@moodlehq/cordova-plugin-badge": "0.8.8-moodle.1",
```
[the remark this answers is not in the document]
dpalou: Nice catch, I released a new version of local-notifications plugin using the original repo instead of the fork. I also added a deprecation notice in the badge plugin repo.

package.json
dpalou: I added a deprecation notice in the README of the file-opener repo.

Yes, it seems to work fine. And our fork only included the permission removal, so if the original one doesn't work then our fork won't work either xD

src/core/services/local-notifications.ts
dpalou: Before this change, the notifications were enabled by default in all devices, there was no need to request the permission. After this patch, the notifications can be disabled by default in Android 13+ (in an emulator they were disabled, in a real Pixel device they were still enabled by default, I don't know the rationale behind this), and they will always be enabled in Android 12 or older like they used to be. If notifications are disabled by default then we need to ask for the permission.

In iOS the app already asks for the permission when you start the app for the first time. This is how it always worked, and we don't do it in JS, the native code does this (I'm not sure if the plugin does this directly or iOS displays it automatically when you enable push notifications).

I talked about this with Juan, and one of the problems here is that notifications are also used for push notifications. If the user is using the app when he receives a push then we could ask for the permission in that moment. But if the app is in background we cannot display a JS modal explaining why do we need the permission, it would need to be something native. And just requesting the permission with no explanation when the app is in background can look weird too. So we decided to do it like this.

NoelDeMartin (approved):

PR #3634, opened 2023-04-24
MOBILE-4317 calendar: Store events from getEventsList in DB

dpalou:
This schedules their default reminder if it's a new event

### DIFF
--- a/src/addons/calendar/services/calendar.ts
+++ b/src/addons/calendar/services/calendar.ts
@@ -658,8 +658,7 @@ export class AddonCalendarProvider {
             const response: AddonCalendarGetCalendarEventByIdWSResponse =
                 await site.read('core_calendar_get_calendar_event_by_id', params, preSets);
 
-            this.storeEventInLocalDb(response.event, { siteId });
-            this.updateEventsReminders([response.event], site.getId());
+            this.updateLocalEvents([response.event], { siteId });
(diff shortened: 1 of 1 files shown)


src/addons/calendar/services/calendar.ts
```diff
@@ -958,8 +958,10 @@ export class AddonCalendarProvider {
             uniqueCacheKey: true,
             updateFrequency: CoreSite.FREQUENCY_SOMETIMES,
         };
-        const response: AddonCalendarGetCalendarEventsWSResponse =
-            await site.read('core_calendar_get_calendar_events', params, preSets);
+        const response =
+            await site.read<AddonCalendarGetCalendarEventsWSResponse>('core_calendar_get_calendar_events', params, preSets);
+
+        this.storeEventsInLocalDB(response.events, { siteId });
```
NoelDeMartin, line 964: Like we've discussed internally, I think it'd be better if we call `updateEventsReminders` here as well as we're doing elsewher. Furthermore, I think we should have both of these operations in a single function (maybe called "updateLocalEvents" or something?) to avoid this situation in the future.

dpalou: Thanks Noel, done!

NoelDeMartin (approved):

PR #3638, opened 2023-04-24
MOBILE-4282: Respect indentation format

NoelDeMartin:


src/core/features/course/services/course.ts
```diff
@@ -250,6 +251,28 @@ export class CoreCourseProvider {
...
 
+    /**
+     * Check whether a course has indentation enabled.
+     *
+     * @param site Site.
+     * @param courseId Course id.
+     * @returns Whether indentation is enabled.
+     */
+    async isCourseIndentationEnabled(site: CoreSite, courseId: number): Promise<boolean> {
+        if (!site.isVersionGreaterEqualThan('4.0')) {
```
dpalou, line 262: I see that MDL-76994 (the backport of the setting) was integrated in 4.0.8 and 4.1.3, so maybe it would be better to do:

`site.isVersionGreaterEqualThan(['4.0.8', '4.1.3', '4.2'])`

What do you think? I guess it will still work without this check because we won't receive the setting, but we'll avoid an unneeded WS call.

dpalou (changes requested):

src/core/features/course/services/course.ts
dpalou: IMO the change doesn't complicate the code, but anyway let's keep it like this.

dpalou (approved):

PR #3645, opened 2023-04-26
MOBILE-4300 course-storage: Update section size when delete all

toanlamt:


src/addons/storagemanager/pages/course-storage/course-storage.ts
```diff
@@ -492,6 +492,15 @@ export class AddonStorageManagerCourseStoragePage implements OnInit, OnDestroy {
...
             CoreCourseHelper.calculateSectionsStatus(this.sections, this.courseId, false, false);
+
+            // For delete all, reset all section sizes so icons are updated.
+            if (this.totalSize == 0) {
+                this.sections.map((section) => {
+                    section.calculatingSize = true;
+                    section.totalSize = 0;
+                    section.calculatingSize = false;
+                });
+            }
```
NoelDeMartin, line 503:
```suggestion
            if (this.totalSize === 0) {
                this.sections.map(section => section.totalSize = 0);
            }
```

NoelDeMartin (changes requested):
Hey, thanks for opening the issue and looking into this!

I think the proper way to fix this would be to make it reactive, so that we don't have to remember to reset section sizes like that. But I understand it would be too much work to refactor everything, so I think your solution is fine.

I've added some suggestions to clean up the code, but other than that I think it looks good and we can merge this (remember to move the issue in jira to "waiting for integration review").

NoelDeMartin (approved):

PR #3690, opened 2023-05-24
[4.3] MOBILE-4344 cron: Check notification and messages

alfonso-salces:


dpalou (approved):

dpalou:
The behat tests failed, I relaunched them and they failed again in the same scenario so maybe it is a real failure. Can you take a look?

dpalou (approved):

PR #3698, opened 2023-06-01
MOBILE-4270 calendar: Fix all courses filter item

NoelDeMartin:


src/addons/calendar/components/filter/filter.ts
```diff
@@ -60,10 +60,20 @@ export class AddonCalendarFilterComponent implements OnInit {
      * @inheritdoc
      */
     ngOnInit(): void {
+        const allCoursesName = Translate.instant('core.fulllistofcourses');
```
dpalou, line 63: Instead of using the translated name to compare, wouldn't it be better to check if id is -1? And we can use a constant for that instead of hardcoding -1.

dpalou (changes requested):

dpalou (approved):

PR #3699, opened 2023-06-06
MOBILE-4345 quiz: Fix sequential navigation

NoelDeMartin:


src/addons/mod/quiz/tests/behat/quiz_navigation.feature
```diff
@@ -69,3 +83,57 @@ Feature: Attempt a quiz in app
     When I press "Submit all and finish" in the app
     And I press "OK" near "Once you submit" in the app
     Then I should find "Review" in the app
+    And I should find "Text of the first question" in the app
+    And I should find "Text of the second question" in the app
+
+  Scenario: Sequential navigation
```
dpalou, line 89: Have you tested if this scenario works in Moodle 3.11? It's possible that sequential quizzes displayed all the questions in the popover (but making them not clickable or something like that). But I might be wrong, I haven't tested this in a long time.

dpalou (changes requested):

src/addons/mod/quiz/tests/behat/quiz_navigation.feature
NoelDeMartin: They should indeed show all the questions, but because of MDL-75210 they only show one for the current page (it's related with the issue of the summary showing only one question, etc.).

dpalou (approved):

PR #3705, opened 2023-06-08
[4.3] MOBILE-4318 reportbuilder: Create report builder link handler

alfonso-salces:


src/core/features/reportbuilder/services/handlers/reportbuilder-link.ts
```diff
@@ -0,0 +1,62 @@
...
+import { CoreContentLinksHandlerBase } from '@features/contentlinks/classes/base-handler';
+import { CoreContentLinksAction } from '@features/contentlinks/services/contentlinks-delegate';
+import { CoreNavigator } from '@services/navigator';
+import { CoreDomUtils } from '@services/utils/dom';
+import { makeSingleton } from '@singletons';
+import { CoreReportBuilderHandler, CoreReportBuilderHandlerService } from './reportbuilder';
+
+/**
+ * Content links handler for report builder
+ * Match reportbuilder/view.php?id=6 with a valid data and optional report id.
```
dpalou, line 26: You say "optional report id" but the id seems to be mandatory in the regexp (unless there is a `d` parameter but I think that part could be wrong, see my other comment). Or the report id is another parameter that the app isn't using?

src/core/features/reportbuilder/services/handlers/reportbuilder-link.ts
```diff
@@ -0,0 +1,62 @@
...
+ */
+@Injectable({ providedIn: 'root' })
+export class CoreReportBuilderLinkHandlerService extends CoreContentLinksHandlerBase {
+
+    name = 'CoreReportBuilderLinkHandler';
+    featureName = 'CoreReportBuilderDelegate';
+    pattern = /\/reportbuilder\/view\.php.*([?&](d|id)=\d+)/;
```
dpalou, line 33: The regexp is looking for either a `d` or an `id` parameter, but the `d` parameter isn't used. Is this ok?

src/core/features/reportbuilder/services/handlers/reportbuilder-link.ts
```diff
@@ -0,0 +1,62 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    getActions(siteIds: string[], url: string, params: Params): CoreContentLinksAction[] {
+        return [{
+            action: async (siteId): Promise<void> => {
+                const modal = await CoreDomUtils.showModalLoading();
```
dpalou, line 41: We display a modal when we need to do WS calls, but in this case you're only navigating so I think it's better not to display it (the user should see the navigation animation immediately).

src/core/features/reportbuilder/services/handlers/reportbuilder-link.ts
```diff
@@ -0,0 +1,62 @@
...
+                }
+            },
+        }];
+    }
+
+    /**
+     * @inheritdoc
+     */
+    async isEnabled(): Promise<boolean> {
+        return await CoreReportBuilderHandler.isEnabled();
```
dpalou, line 58: Wouldn't it be better to use `CoreReportBuilder.isEnabled` instead? It's better to create a dependency with the main service rather than with a handler.

src/core/features/reportbuilder/services/handlers/reportbuilder-link.ts
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Content links handler for report builder
+ * Match reportbuilder/view.php?id=6 with a valid data and optional report id.
+ */
+@Injectable({ providedIn: 'root' })
+export class CoreReportBuilderLinkHandlerService extends CoreContentLinksHandlerBase {
+
+    name = 'CoreReportBuilderLinkHandler';
+    featureName = 'CoreReportBuilderDelegate';
```
dpalou, line 32: This property is used to check if this feature is disabled, to avoid having to implement the `isEnabled` function just for this. But in this handler you still need to implement the `isEnabled` function because there are other conditions, and that function is also checking if this feature is disabled so we're doing this twice. IMO we can remove this property because it's already handled in the `isEnabled` function.

dpalou (changes requested):

src/core/features/reportbuilder/services/handlers/reportbuilder-link.ts
alfonso-salces: The comment is wrong, we require a report id to open an specific report, so you're right.

src/core/features/reportbuilder/services/handlers/reportbuilder-link.ts
alfonso-salces: Is not ok, I forget to remove it when I take the regex from other handler.

dpalou (approved):

PR #3708, opened 2023-06-12
MOBILE-4261 user: Use initials instead of default avatar when possible

crazyserver:


src/core/services/sites.ts
```diff
@@ -2016,9 +2018,11 @@ export type CoreSiteBasicInfo = {
     id: string; // Site ID.
     siteUrl: string; // Site URL.
     siteUrlWithoutProtocol: string; // Site URL without protocol.
-    fullName?: string; // User's full name.
+    fullname?: string; // User's full name.
```
dpalou, line 2021: Please document these changes in upgrade.txt. In theory this change could affect plugins because they have access to the service, but plugins should only work in a single site (they shouldn't access other sites or the list of sites) so IMO it's ok to change this.

dpalou (changes requested):

PR #3718, opened 2023-06-19
MOBILE-4372 course-format: Change course index fab position if lang is rtl

alfonso-salces:


crazyserver:
Hi Alfonso!

I'm not sure about the solution, I would prefer solving that using CSS. Checking at Ionic code, this positioned functions that support RTL:

```scss
:host(.fab-horizontal-start) {
  @include position-horizontal(
    calc(#{$fab-content-margin} + var(--ion-safe-area-left, 0px)), null
  );
}

:host(.fab-horizontal-end) {
  @include position-horizontal(
    null, calc(#{$fab-content-margin} + var(--ion-safe-area-right, 0px))
  );
}
``` 

This position-horizontal functions applies changes when RTL is set. So we should investigate better why on real devices RTL is not working.

NoelDeMartin:
I'm looking at this, and having @crazyserver's comment into account, I wonder why do we even need to do anything here. Usually, "start" and "end" are semantic and already take RTL/LTR into account (in contrast with using "right" or "left"). So is that an ionic bug, or for some reason in this context the start/end keywords actually mean right/left?

Other than this, looking at the screenshot in the issue I wonder if maybe the solution should be to position the user tour from right to left instead of moving the FAB? Also, I see we're using `CoreUserToursAlignment.Start` and that may have the same problem as ionic's `horizontal` attribute. But this is one we made ourselves, so we could fix it so that Start means right with RTL and left with LTR.

alfonso-salces:
Thanks for your appreciations @NoelDeMartin and @crazyserver,
I removed the changes in the template and adjusted the user tour alignment for RTL.
Also, added some css to fix the margin of modal lateral when we are in a RTL language.

crazyserver (approved):

PR #3720, opened 2023-06-23
MOBILE-4267 about: Add idnumber, department and institution

alfonso-salces:


src/core/features/user/pages/about/about.html
```diff
@@ -59,14 +59,6 @@ <h2>{{ user.fullname }}</h2>
...
                     </ion-label>
                 </ion-item>
-                <ion-item class="ion-text-wrap" *ngIf="formattedAddress">
-                    <ion-label>
-                        <p class="item-heading">{{ 'core.user.address' | translate}}</p>
-                        <p><a class="core-anchor" [href]="encodedAddress" core-link auto-login="no" [showBrowserWarning]="false">
-                                {{ formattedAddress }}
-                            </a></p>
-                    </ion-label>
-                </ion-item>
```
dpalou, line 69: In the same file we also display "user.address" at the top. I think in this case we only display the city and country so I guess it's ok to keep it (please confirm that it doesn't display the full address).

Also, since formattedAddress and encodedAddress is no longer used you can remove the variables and the JS code that calculates them.

src/core/features/user/pages/about/about.ts
```diff
@@ -88,11 +89,12 @@ export class CoreUserAboutPage implements OnInit, OnDestroy {
     async ngOnInit(): Promise<void> {
         this.userId = CoreNavigator.getRouteNumberParam('userId') || 0;
         this.courseId = CoreNavigator.getRouteNumberParam('courseId') || 0;
+        this.canShowDepartment = !!this.courseId && this.userId == this.site.getUserId();
```
dpalou, line 92: Juan's comment says:

>  Institution, Department and ID number BUT ONLY when the current user is viewing other user profile (not for itself)

But you're doing the opposite, right? You're only displaying it when viewing your own profile.

Also, I don't know why are you checking the course id.

src/core/features/user/pages/about/about.ts
```diff
@@ -88,11 +89,12 @@ export class CoreUserAboutPage implements OnInit, OnDestroy {
...
 
         // Allow to change the profile image only in the app profile page.
         this.canChangeProfilePicture =
             !this.courseId &&
-            this.userId == this.site.getUserId() &&
+            this.userId == this.site.getUserId()&&
```
dpalou, line 97: You removed a needed space here.

dpalou (changes requested):

PR #3722, opened 2023-06-26
MOBILE-4360 tests: Fix behat tests

alfonso-salces:


src/tests/behat/customlangstrings.feature
```diff
@@ -6,12 +6,14 @@ Feature: Custom lang strings
       | username |
       | student1 |
 
+  @lms_from4.0
```
NoelDeMartin, line 9: It seems like the entire feature "custom lang strings" is only available in 4.0, so it would make sense to have this tag directly on the feature rather than individual scenarios.

NoelDeMartin (approved):

NoelDeMartin (approved):

PR #3724, opened 2023-06-28
MOBILE-3371: Global Search

NoelDeMartin:
Behat tests won't pass until [MDL-56020](https://tracker.moodle.org/browse/MDL-56020) is integrated in core.


src/core/features/search/components/global-search-filters/global-search-filters.component.ts
```diff
@@ -0,0 +1,236 @@
...
+
+        this.emitFiltersUpdated();
+    }
+
+    /**
+     * Get the status for a filter representing a group of filters.
+     *
+     * @param filters Filters in the group.
+     * @returns Group filter status
+     */
```
dpalou, line 200: This function was confusing for me because I didn't know checkboxes could have an indeterminate status. Maybe we should document what each value does (true -> enabled, false -> disabled, null -> indeterminate).

src/core/features/search/components/global-search-filters/global-search-filters.component.ts
```diff
@@ -0,0 +1,236 @@
...
+@Component({
+    selector: 'core-search-global-search-filters',
+    templateUrl: 'global-search-filters.html',
+    styleUrls: ['./global-search-filters.scss'],
+})
+export class CoreSearchGlobalSearchFiltersComponent implements OnInit {
+
+    allSearchAreaCategories: boolean | null = true;
+    searchAreaCategories: Filter<CoreSearchGlobalSearchSearchAreaCategory>[] = [];
+    allCourses: boolean | null = true;
```
dpalou, line 37: For consistency with the `searchAllAreaCategories` comment maybe this one should be named `searchAllCourses`. But then `searchAreaCategories` can be confused with these ones too... That's the problem of having the same word for the verb and the noun lol

Maybe we can have:

```
searchAllAreaCategories: boolean | null = true;
searchAllCourses: boolean | null = true;
areaCategories: Filter<CoreSearchGlobalSearchSearchAreaCategory>[] = [];
courses: Filter<CoreEnrolledCourseData>[] = [];
```

src/core/components/empty-box/empty-box.scss
```diff
@@ -28,6 +28,11 @@
     &.core-empty-box-clickable {
         z-index: 0;
     }
+
+    &.dimmed {
+        opacity: .32;
+    }
```
dpalou, line 34: Have you checked if this complies the contrast for accessibility? The dimmed empty box text will have less contrast so maybe the contrast is not enough now.

src/core/features/search/components/global-search-filters/global-search-filters.component.ts
```diff
@@ -0,0 +1,236 @@
...
+        }
+
+        const checked = this.allSearchAreaCategories;
+
+        this.searchAreaCategories.forEach(searchAreaCategory => {
+            if (searchAreaCategory.checked === checked) {
+                return;
+            }
+
+            searchAreaCategory.checked = checked;
```
dpalou, line 105: Every time a value changes it will call `onSearchAreaCategoryInputChanged`, which will check the status of all categories and call `emitFiltersUpdated`. The same happens for courses. So we'll be calling a lot of functions that aren't really needed, and we'll iterate the whole list n times in `getGroupFilterStatus` (so this is an O(n^2) problem).

I guess the lists shouldn't be long enough to make this be a problem, but maybe it would be better to treat it. I'm not sure if this can be treated while using ngModel, but I think you can control it if you use FormControl:

`control.setValue(foo, { emitEvent: false });`

src/addons/block/globalsearch/services/block-handler.ts
```diff
@@ -0,0 +1,61 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    async isEnabled(): Promise<boolean> {
+        const site = CoreSites.getRequiredCurrentSite();
+
+        return !site.isFeatureDisabled('CoreNoDelegate_GlobalSearch')
+            && site.wsAvailable('core_search_get_results')
+            && site.canUseAdvancedFeature('enableglobalsearch');
```
dpalou, line 41: It might be nice to abstract this into a function since it's used twice in the app. And if we end up implementing a like handler (I'll add another comment about that) you should probably have to use it there too.

dpalou (changes requested):
Along with the comments below, shouldn't we create a link handler to open the global search if the user clicks a link to /search/index.php?

src/core/classes/items-management/paginated-items-manager-source.ts
```diff
@@ -79,22 +81,24 @@ export abstract class CorePaginatedItemsManagerSource<Item = unknown> extends Co
      * Load more items, if any.
      */
     async load(): Promise<void> {
-        if (this.dirty) {
-            const { items, hasMoreItems } = await this.loadPageItems(0);
+        await this.sempahore.run(async () => {
```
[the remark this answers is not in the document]
NoelDeMartin: I don't remember exactly, but I think this happened sometimes when scrolling. Maybe this happens because the infinite-loader may be triggered more than once, depending on how "fast" the loader is shown. I'm not sure how easy it would be to fix something like that.

What you mention about waiting for the previous load to finish; the problem with that is that it can create similar race conditions to the ones raised in [MOBILE-4119](https://tracker.moodle.org/browse/MOBILE-4119). For example, if we update something (like the search query or filters), and then call "load" but a load was already happening, this would be ignored and result in the wrong results list. That wouldn't be a problem if we apply this logic of "waiting the previous one to finish" in the infinite loader callback instead. But we would still need to keep the semaphore in the source, so I'm not sure it's worth looking into.

All in all, this semaphore mechanism is something that should not be necessary, but if it happens at some point I don't think it's too much of a problem. Worst case scenario would be that we load 2 or 3 pages instead of one, but that's not the end of the world keeping in mind that it should happen often. Without the semaphore, we may be loading the same page multiple times and have duplicated items, which is worse.

src/core/features/search/services/global-search.ts
```diff
@@ -36,3 +43,312 @@ export type CoreSearchGlobalSearchResultModule = {
...
+        // await CoreAnalytics.logEvent({
+        //     type: CoreAnalyticsEventType.VIEW_ITEM_LIST,
+        //     ws: 'core_search_view_results',
+        //     name: 'search',
+        //     data: params, // TODO what is this data supposed to be?
+        //     url: CoreUrlUtils.addParamsToUrl('/search/index.php', {
+        //         q: query,
+        //         // TODO context?
+        //     }),
+        // });
```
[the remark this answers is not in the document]
NoelDeMartin: I completely forgot about this xD thanks for noticing.

src/core/features/search/components/global-search-filters/global-search-filters.component.ts
```diff
@@ -0,0 +1,236 @@
...
+type Filter<T=unknown> = T & { checked: boolean };
+
+@Component({
+    selector: 'core-search-global-search-filters',
+    templateUrl: 'global-search-filters.html',
+    styleUrls: ['./global-search-filters.scss'],
+})
+export class CoreSearchGlobalSearchFiltersComponent implements OnInit {
+
+    allSearchAreaCategories: boolean | null = true;
```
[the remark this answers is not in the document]
NoelDeMartin: In this case I think it's better to live it like that, because this value indicates whether all the filters from the `searchAreaCategories` property are checked or not (in the same way that `allCourses` works for `courses`). I think if we rename this, we should also rename the other attributes and it'd end up like this:

```ts
    searchAllSearchAreaCategories: boolean | null = true;
    searchSearchAreaCategories: Filter<CoreSearchGlobalSearchSearchAreaCategory>[] = [];
    searchAllCourses: boolean | null = true;
    searchCourses: Filter<CoreEnrolledCourseData>[] = [];
```

I'm not convinced that is better.

src/core/features/search/components/global-search-filters/global-search-filters.component.ts
NoelDeMartin: I replied in the other comment, let's have the conversation there.

src/core/components/empty-box/empty-box.scss
NoelDeMartin: I'm using the values provided by the design specs, but I'll double check.

src/core/components/course-image/course-image.ts
```diff
@@ -0,0 +1,81 @@
...
+import { CoreCourseListItem } from '@features/courses/services/courses';
+import { CoreCoursesHelper } from '@features/courses/services/courses-helper';
+import { CoreColors } from '@singletons/colors';
+
+@Component({
+    selector: 'core-course-image',
+    templateUrl: 'course-image.html',
+    styleUrls: ['./course-image.scss'],
+})
+export class CoreCourseImageComponent implements OnInit, OnChanges {
```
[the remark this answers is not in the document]
NoelDeMartin: Eventually yes, but I didn't want to add this refactor in this PR because it's big enough as it is. And I'm not sure it can be replaced so easily without some testing. So I decided to leave it as an improvement to do later.

src/core/features/search/services/global-search.ts
```diff
@@ -36,3 +43,312 @@ export type CoreSearchGlobalSearchResultModule = {
...
+    /**
+     * Get available search areas.
+     *
+     * @returns Search areas.
+     */
+    async getSearchAreas(): Promise<CoreSearchGlobalSearchSearchArea[]> {
+        const site = CoreSites.getRequiredCurrentSite();
+        const params: CoreSearchGetSearchAreasListWSParams = {};
+
+        const { areas } = await site.read<CoreSearchGetSearchAreasListWSResponse>('core_search_get_search_areas_list', params);
```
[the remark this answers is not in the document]
NoelDeMartin: Ok, I followed your suggestion and also added a PTR in the filters popover.

src/core/components/empty-box/empty-box.scss
NoelDeMartin: I just took a look and you're right it didn't pass the a11y bar; and what it's more it didn't match the design either so I'm not sure where I got that. In any case, it's fixed now. Thanks for paying attention to these details.

src/core/features/search/components/global-search-filters/global-search-filters.component.ts
NoelDeMartin: I've debugged this, and it seems to be working with `O(n)` instead of `O(n^2)`. I think this happens because given the single-threaded nature of javascript, the first input change will recompute all the filters and then when other listeners are called, they won't be triggered again (because there's already some guards in the listeners to prevent computing if the value is already applied).

src/core/classes/semaphore.ts
```diff
@@ -0,0 +1,89 @@
...
+            });
+        }
+    }
+
+    /**
+     * Notify listeners that the resource has been released.
+     */
+    private notifyAvailable(): void {
+        this.listeners.forEach(listener => listener());
+    }
```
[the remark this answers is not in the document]
NoelDeMartin: I've removed everything relating with the semaphore because I still need to debug it further, and I think we can add it later as an improvement.

src/core/features/search/components/global-search-filters/global-search-filters.component.ts
```diff
@@ -0,0 +1,267 @@
...
+        }
+
+        this.allSearchAreaCategories = this.getGroupFilterStatus(this.searchAreaCategories);
+    }
+
+    /**
+     * Update courses.
+     */
+    private async updateCourses(): Promise<void> {
+        const courses = await CoreCourses.getUserCourses(true);
```
dpalou, line 187: You're passing `preferCache=true`, so the `invalidateUserCourses` in the refresh function won't do anything. If we want to allow updating the course list here you should remove this param. Another option is to only refresh the search areas if the user performs a PTR, but it can be confusing for the user.

src/core/features/search/components/global-search-filters/global-search-filters.component.ts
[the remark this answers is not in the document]
dpalou: Ok, let's leave it like this then.

src/core/components/course-image/course-image.ts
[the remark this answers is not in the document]
dpalou: Ok, I opened an issue so we don't forget: https://tracker.moodle.org/browse/MOBILE-4413

dpalou (changes requested):

NoelDeMartin:
@dpalou Other than what we already discussed, I've also implemented support for 2 new result types: side blocks and course sections. This is in a new commit, but all the other changes are mostly amended in the previous ones. I also had to rebase to resolve merge conflicts, so you may have to look into the commits again.

src/core/features/sitehome/services/handlers/index-link.ts
```diff
@@ -36,14 +37,22 @@ export class CoreSiteHomeIndexLinkHandlerService extends CoreContentLinksHandler
...
+        }
+
         return [{
             action: (siteId: string): void => {
                 CoreNavigator.navigateToSitePath(
                     `/${CoreMainMenuHomeHandlerService.PAGE_NAME}/${CoreSiteHomeHomeHandlerService.PAGE_NAME}`,
                     {
                         preferCurrentTab: false,
                         siteId,
+                        params: pageParams,
```
dpalou, line 55: The unit tests are failing, I think it's related to this change.

dpalou (changes requested):

dpalou (approved):

PR #3727, opened 2023-06-29
Mobile 4009

crazyserver:


src/core/components/components.module.ts
```diff
@@ -91,6 +92,7 @@ import { CoreSheetModalComponent } from '@components/sheet-modal/sheet-modal';
         CoreModIconComponent,
         CoreNavBarButtonsComponent,
         CoreNavigationBarComponent,
+        CorePasswordModalComponent,
```
NoelDeMartin, line 95: This declaration means that this component will be included in the initial bundle of the app, but I'm not sure if it's so important to warrant that. For modals in particular, we've already started using a pattern to load them lazily so maybe this one could also use that. For example, you can see how we're doing that for the audio recorder: https://github.com/moodlehq/moodleapp/blob/main/src/core/features/fileuploader/services/fileuploader.ts#L162

src/core/features/course/pages/course-summary/course-summary.page.ts
```diff
@@ -288,11 +288,40 @@ export class CoreCourseSummaryPage implements OnInit, OnDestroy {
...
         }
 
+        const guestInstanceId = await this.guestInstanceId;
+        if (this.useGuestAccess && this.guestAccessPasswordRequired && guestInstanceId) {
+
+            let password = '';
+            let error: string | undefined;
+
+            let validated = await CoreCourse.isGuestAccessPreauthorized(this.courseId);
+            while (!validated) {
```
NoelDeMartin, line 303: Rather than doing this, I think it'd be easier to pass a validator function as an input for the modal. I think it would make a better UX as well, because the modal itself could handle the loading state or something else, rather than modals opening/closing every time (which is fine if we decide that's the better UX, but it doesn't have to be a limitation).

src/core/features/courses/services/courses.ts
```diff
@@ -1857,3 +1893,28 @@ export type CoreCourseAnyCourseDataWithOptions = CoreCourseAnyCourseData & {
...
+    password: string; // the course password
+};
+
+/**
+ * Params of enrol_guest_validate_password WS.
+ */
+export type EnrolGuestValidatePasswordWSParams = {
+    instanceid: number; // instance id of guest enrolment plugin
+    password: string; // the course password
+};
```
NoelDeMartin, line 1911: It seems like these two types should actually be only one.

src/core/features/course/tests/behat/guest_access.feature
```diff
@@ -0,0 +1,160 @@
...
+    And I set the following fields to these values:
+      | Allow guest access | Yes |
+    And I press "Save changes"
+
+    Given I entered the app as "student1"
+    When I press "Site home" in the app
+    And I press "Available courses" in the app
+    And I press "Course 1" in the app
+
+    Then I should find "Course summary" in the app
```
NoelDeMartin, line 84:
```suggestion
    And I press "Course 1" in the app
    Then I should find "Course summary" in the app
```

src/core/features/course/tests/behat/guest_access.feature
```diff
@@ -0,0 +1,160 @@
...
+    And I should find "Test feedback name" in the app
+    And I should find "Test glossary" in the app
+    And I should find "Quiz 1" in the app
+    And I should find "Test survey name" in the app
+    And I should find "Test wiki name" in the app
+    And I should find "Test lesson name" in the app
+    And I should find "Test scorm name" in the app
+    And I should find "Test workshop name" in the app
+    And I press "assignment" in the app
+    And I should not find "Add submission" in the app
```
NoelDeMartin, line 103:
```suggestion
    And I should find "Test workshop name" in the app

    When I press "assignment" in the app
    Then I should not find "Add submission" in the app
```

src/core/features/course/tests/behat/guest_access.feature
```diff
@@ -0,0 +1,160 @@
...
+      | Allow guest access | Yes |
+      | Password | moodle_rules |
+    And I press "Save changes"
+
+    Given I entered the app as "student1"
+    When I press "Site home" in the app
+    And I press "Available courses" in the app
+    And I press "Course 1" in the app
+
+    Then I should find "Course summary" in the app
```
NoelDeMartin, line 120:
```suggestion
    And I press "Course 1" in the app
    Then I should find "Course summary" in the app
```

src/core/features/course/tests/behat/guest_access.feature
```diff
@@ -0,0 +1,160 @@
...
+    And I should find "Course" in the app
+    And I should find "Guest access requires password" in the app
+
+    When I press "View course" "ion-button" in the app
+    And I set the following fields to these values in the app:
+      | Password | wrong |
+
+    And I press "Submit" "ion-button" in the app
+
+    And I should find "Incorrect access password, please try again" in the app
```
NoelDeMartin, line 130:
```suggestion
    And I set the following fields to these values in the app:
      | Password | wrong |
    And I press "Submit" "ion-button" in the app
    Then I should find "Incorrect access password, please try again" in the app
```

src/core/features/course/tests/behat/guest_access.feature
```diff
@@ -0,0 +1,160 @@
...
+    And I press "Submit" "ion-button" in the app
+
+    And I should find "That access password was incorrect, please try again" in the app
+
+    And I set the following fields to these values in the app:
+      | Password | moodle_rules |
+
+    And I press "Submit" "ion-button" in the app
+
+    Then the header should be "Course 1" in the app
```
NoelDeMartin, line 145:
```suggestion
    Given the following config values are set as admin:
      | showhint | 1 | enrol_guest |
    When I press "Submit" "ion-button" in the app
    Then I should find "That access password was incorrect, please try again" in the app

    When I set the following fields to these values in the app:
      | Password | moodle_rules |
    And I press "Submit" "ion-button" in the app
    Then the header should be "Course 1" in the app
```

src/core/features/course/tests/behat/guest_access.feature
```diff
@@ -0,0 +1,160 @@
...
+    And I should find "Test feedback name" in the app
+    And I should find "Test glossary" in the app
+    And I should find "Quiz 1" in the app
+    And I should find "Test survey name" in the app
+    And I should find "Test wiki name" in the app
+    And I should find "Test lesson name" in the app
+    And I should find "Test scorm name" in the app
+    And I should find "Test workshop name" in the app
+    And I press "assignment" in the app
+    And I should not find "Add submission" in the app
```
NoelDeMartin, line 160:
```suggestion
    And I should find "Test workshop name" in the app

    When I press "assignment" in the app
    Then I should not find "Add submission" in the app
```

NoelDeMartin (changes requested):

src/core/features/course/services/course-helper.ts
```diff
@@ -630,19 +627,18 @@ export class CoreCourseHelperProvider {
...
-            if (!info.status) {
-                // Not active, reject.
-                return false;
-            }
 
-            // Don't allow guest access if it requires a password.
-            return !info.passwordrequired;
+            // Don't allow guest access if it requires a password and it's available.
+            return {
+                guestAccess: !!info.status && (!info.passwordrequired || CoreCourses.isValidateGuestAccessPasswordAvailable()),
```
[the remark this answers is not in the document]
crazyserver: It was defined as PARAM_RAW on WS but used as a boolean. Changed! THanks!

src/core/components/password-modal/password-modal.ts
```diff
@@ -12,37 +12,42 @@
...
 })
-export class AddonModLessonPasswordModalComponent {
+export class CorePasswordModalComponent {
 
     @ViewChild('passwordForm') formElement?: ElementRef;
 
+    @Input() password = '';
+
+    @Input() title = 'core.login.password';
+    @Input() placeholder =  'core.login.password';
```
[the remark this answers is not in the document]
crazyserver: I don't think it's a problem, because it's really coming from moodle.php language.

src/core/features/course/services/course.ts
```diff
@@ -1030,6 +1030,16 @@ export class CoreCourseProvider {
         return sections.reduce((previous: CoreCourseModuleData[], section) => previous.concat(section.modules || []), []);
     }
 
+    /**
+     * Check if the user has access to the course as guest with a previous sent password.
+     *
+     * @param courseId Course Id.
+     * @returns Wether the user has access.
+     */
+    async isGuestAccessPreauthorized(courseId: number): Promise<boolean> {
```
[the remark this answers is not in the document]
crazyserver: I've inlined in the call.

crazyserver:
Thanks Noel! Can you review again, please?

src/core/features/course/services/course-helper.ts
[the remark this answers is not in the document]
NoelDeMartin: It seems like we're actually using the same type in the app for two different types in the LMS, so we should probably fix that.

One type is indeed a boolean, but the one that is typed as a PARAM_RAW seems to be `bool | string` where a string means an error message. It's difficult to see just from the webservice description, but if you search on the code you'll eventually find it here: https://github.com/moodle/moodle/blob/master/enrol/self/lib.php#L330

So I think we should treat them as two separate types in the app as well and handle them accordingly. In particular, it seems like we're not declaring the WS response types for `core_enrol_get_course_enrolment_methods` anywhere (we just return the untyped response). We should always type WS responses explicitly in case anything changes or we refactor some types. Personally I also think we should never expose WS types to the rest of the application, but that's another matter and we're already doing that throughout the app so that's only a suggestion.

src/addons/storagemanager/pages/course-storage/course-storage.ts
```diff
@@ -657,6 +664,25 @@ export class AddonStorageManagerCourseStoragePage implements OnInit, OnDestroy {
...
+        }
+
+        try {
+            // The user is not enrolled in the course. Use getCourses to see if it's an admin/manager and can see the course.
+            return await CoreCourses.getCourse(courseId);
+        } catch {
+            // Ignore errors.
+        }
+
+        return await CoreCourses.getCourseByField('id', this.courseId);
```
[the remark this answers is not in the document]
NoelDeMartin: Yeah, only reading that I still don't understand what's the difference but if we're not sure how it works we should probably look into it before calling these WSs. In any case, whatever we know about their differences, even if it's only what you just wrote in that comment, should be documented. Otherwise, these two functions are equivalent from the outside and that could cause some problems later on (I have used methods to retrieve courses myself and I wouldn't have been aware of the difference).

src/core/features/course/services/course-helper.ts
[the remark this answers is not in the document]
crazyserver: I've chenged it a bit but checking the code, the WS returns of the guest method are wrong and should be bool|string as well (they are using the same function).

NoelDeMartin (approved):

phyr0s:
@NoelDeMartin  @crazyserver  this commit MOBILE-4009 courses: Pass isGuest param on navigation -->breaks the animation --> e8e2b94b1. --> this line now take time the await to process so when you go from main to one course the animation is weird and broken.

this.isGuest = CoreNavigator.getRouteBooleanParam('isGuest') ??
            (!!this.course && (await CoreCourseHelper.courseUsesGuestAccessInfo(this.course.id)).guestAccess);

crazyserver:
I'll try to amend that! Thanks for noticing!

PR #3729, opened 2023-07-05
MOBILE-4331: Implement native plugin

NoelDeMartin:
This PR only implements the plugin, but it's not actually used anywhere in the app (this will be implemented in a follow-up PR). It can be used like this:

```ts
CoreNative.plugin('systemUI')?.setNavigationBarColor('#ff00ff');
```

Also, keep in mind that I just copied the code from [fagundes/cordova-plugin-navigationbar](https://github.com/fagundes/cordova-plugin-navigationbar/blob/master/src/android/NavigationBar.java), but presumably that has to be updated in some way (that's the whole reason why we're doing our own plugin to begin with).

Finally, I wasn't able to test this in an emulator because I didn't manage to display the navigation bar :/. But I tried with a real device and it worked.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -13,7 +13,9 @@ jobs:
       with:
         node-version-file: '.nvmrc'
     - name: Install npm packages
-      run: npm ci --no-audit
+      run: |
+        npm ci --no-audit
+        npm ci --no-audit --prefix cordova-plugin-moodleapp
     - name: Check langindex
       run: |
         result=$(cat scripts/langindex.json | grep \"TBD\" | wc -l); test $result -eq 0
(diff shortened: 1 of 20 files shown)


crazyserver (approved):

phyr0s:
Hello @NoelDeMartin, Hope you are well in Japan :) 
Since this merge ionic serve stop working in master --> MoodleAppPlugins is not found.

NoelDeMartin:
Hey @phyr0s thanks :D.

About serving the app, that's probably happening because you haven't run `npm ci` or `npm install` after updating the code, can you double-check?

Other than that, I realized you may have some problems trying to build for android if you haven't run `npm ci` or `npm install` inside of the plugin folder (`/cordova-plugin-moodleapp`), but I'm trying to work on a fix so that it isn't necessary. In this PR I also modified the scripts you run with `npm run dev:android` and `npm run prod:android` so that the plugin is rebuilt when necessary, so that shouldn't be a problem.

phyr0s:
Runned npm install and same :( @NoelDeMartin

phyr0s:
Runned npm install and same :( @NoelDeMartin

NoelDeMartin:
@phyr0s Hm can you try running `npm ci` instead? And also make sure you're using the proper npm version (we recommend using nvm and doing `nvm install`).

In any case, to double check that everything is working properly, you should have a `cordova-plugin-moodleapp` folder within your `node_modules`, and that folder should be a symlink to the plugin folder, which has the `MoodleAppPlugins` type defined so it should be fine. We actually know it should work because that's what [the ci pipeline](https://github.com/moodlehq/moodleapp/actions/runs/5483738834) is doing and tests pass, so it should be working for you as well.

phyr0s:
with npm ci is working @NoelDeMartin, maybe is better to add post install to execute same as npm ci when npm install.

PR #3731, opened 2023-07-06
MOBILE-4283 quiz: Allow students to hide timer

crazyserver:


src/core/components/timer/core-timer.html
```diff
@@ -1,21 +1,32 @@
...
+                    {{ timeLeft | coreSecondsToHMS }}
+                </span>
+            </div>
 
-<div *ngIf="mode === modeBasic" class="core-timer ion-padding" role="timer"
-    [ngClass]="{'ion-text-center': align == 'center', 'ion-text-end': align == 'right'}">
-    <ng-container *ngTemplateOutlet="timerTemplate"></ng-container>
-</div>
+            <div *ngIf="hiddable" class="core-timer-visibility">
+                ( <a *ngIf="showTimeLeft" (click)="toggleTimeLeftVisibility()">{{ 'core.hide' | translate }}</a>
```
NoelDeMartin, line 22: Even though this looks like an anchor, it should be a button for accessibility. Even if it isn't, I'm not sure how an anchor without `href` will be interpreted by screen readers... In any case, I'd double check the accessibility of this component (additionally, I suppose the timer should have an `aria-live` attribute to indicate that it changes, but that was already unsupported before so I guess we can leave it like that).

NoelDeMartin (changes requested):

src/core/components/timer/core-timer.html
crazyserver: I agree on the button part, I'll change it.

About the aria live, we're using role timer that is has aria live implicit on it. https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/timer_role

NoelDeMartin (approved):

PR #3735, opened 2023-07-11
MOBILE-4283 timer: Make timer unhiddable on 100 secs or less

crazyserver:


src/core/components/timer/timer.ts
```diff
@@ -73,6 +73,11 @@ export class CoreTimerComponent implements OnInit, OnDestroy {
             container = container || this.elementRef.nativeElement;
             this.timeLeft = Math.max(endTime - CoreTimeUtils.timestamp(), 0);
 
+            if (this.timeLeft <= 100) {
```
NoelDeMartin, line 76: Couldn't be cancel the interval at this point? What's the point to keep recalculating the same thing (presumably, time left won't increase).

NoelDeMartin (changes requested):

src/core/components/timer/timer.ts
NoelDeMartin: I see, sorry I didn't realize this interval is also used to update the time left.

NoelDeMartin (approved):

PR #3737, opened 2023-07-12
Mobile 4331

crazyserver:


src/app/app.component.ts
```diff
@@ -206,9 +206,30 @@ export class AppComponent implements OnInit, AfterViewInit {
 
             this.logger.debug('Hide splash screen');
             SplashScreen.hide();
+            this.setSystemUIColorsAfterSplash();
         });
     }
 
+    /**
```
NoelDeMartin, line 213: I think this comment should go inside the method as inline comments, given that these are implementation details. If anything, we could mention that setting the color won't be completed instantly even though this method is synchronous (or maybe we could even return a promise to make that clear, no need to await on the result though).

I'm also not a fan of the name of the function, given that in reality it has nothing to do with the splash; it's just setting a color constantly for a second. Not sure about what other name to use though, so I guess it's fine to leave it like this if we can't come up with anything better.

dpalou: If we don't like the name of this function maybe we can create a function named "hideSplashScreen" that hides the splash screen and then does this code without needing an extra function.

package.json
```diff
@@ -84,6 +84,7 @@
     "@moodlehq/cordova-plugin-ionic-webview": "5.0.0-moodle.2",
     "@moodlehq/cordova-plugin-local-notification": "0.9.0-moodle.11",
     "@moodlehq/cordova-plugin-qrscanner": "3.0.1-moodle.5",
+    "@moodlehq/cordova-plugin-statusbar": "git+https://github.com/moodlemobile/cordova-plugin-statusbar.git",
```
dpalou, line 87: I think you also need to change the name in the cordova->plugins section, line 232.

Also, we shouldn't integrate this until the plugin is released and we should use the npm version instead of github.

cordova-plugin-moodleapp/src/android/SystemUI.java
```diff
@@ -32,11 +32,7 @@ public class SystemUI extends CordovaPlugin {
     public boolean execute(String action, JSONArray args, CallbackContext callbackContext) {
```
dpalou, line 32: Shouldn't we remove the whole class and the equivalent TS class? If you want I can do it later in the issue to store the token in a secure storage, since in that issue I'm adding a new class to the plugin.

dpalou (changes requested):

src/app/app.component.ts
dpalou: If we don't like the name of this function maybe we can create a function named "hideSplashScreen" that hides the splash screen and then does this code without needing an extra function.

cordova-plugin-moodleapp/src/android/SystemUI.java
crazyserver: Yes please, if you can do it it would be great!

src/app/app.component.ts
crazyserver: Solved!

package.json
dpalou: The plugin version is ok now, but you haven't changed the other line in package.json:

https://github.com/moodlehq/moodleapp/blob/10b0cfab9fb4efee4ffcfe8454bc9260aa9a1eb3/package.json#L232

dpalou (approved):

PR #3738, opened 2023-07-15
fix(quiz select): added cancel text into select quiz selection

phyr0s:
Added translation for Cancel string

### DIFF
--- a/src/addons/qtype/match/component/addon-qtype-match.html
+++ b/src/addons/qtype/match/component/addon-qtype-match.html
@@ -15,7 +15,7 @@
                 {{ row.accessibilityLabel }}
             </label>
         </ion-label>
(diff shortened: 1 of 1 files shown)


NoelDeMartin:
Hey, thanks for your contribution. You're right that this select's cancel is not translated, but I've checked and none of the selects in the app are :scream:. So I think if we fix it, we should fix it everywhere rather than only in this file. I'll ask the rest of the team to see what they think, but if nobody sees anything wrong with that I'll open a new issue to do this, it shouldn't be too much work.

NoelDeMartin:
Thanks for opening this, I've discussed it with the rest of the team and we've decided to open an issue to fix it across the app. I'm closing this in favour of that, and I'll be sure to credit you in the commits when that is fixed :).

You can follow that issue here: [MOBILE-4387](https://tracker.moodle.org/browse/MOBILE-4387)

PR #3742, opened 2023-07-17
Mobile 4009

crazyserver:


src/core/components/password-modal/password-modal.ts
```diff
@@ -108,4 +108,5 @@ export type CorePasswordModalResponse = {
     password: string;
     validated?: boolean;
     error?: string;
+    cancel?: boolean;
```
NoelDeMartin, line 111: I don't think we should declare this property here, given that the password modal is not returning this anywhere. Instead, if you want to extend this type in the course summary usage, you can declare it there like `CorePasswordModalResponse & { cancel?: boolean }` or something.

NoelDeMartin (changes requested):
Acceptance tests were failing, but I'm not sure if the failures were related with these changes so I triggered them again. But if it keeps failing maybe something has introduced a regression (or most likely, the tests need to be updated for the new workflow).

src/core/components/password-modal/password-modal.ts
crazyserver: Thanks Done!

crazyserver:
I've added a new function userHasAccessToCourse to let the user enter the course when offline

NoelDeMartin (approved):

phyr0s:
@NoelDeMartin  @crazyserver  animations still broken in go back action (enter works fine no glitch) due await getting if user is guest  as --> https://github.com/moodlehq/moodleapp/pull/3727

NoelDeMartin:
Thanks for letting us know @phyr0s. We're on low capacity at the moment because many members of the team are on holidays, but we'll take a look after the holiday period.

PR #3754, opened 2023-07-25
MOBILE-4323 enrol: Create enrol delegates

crazyserver:


src/core/features/course/pages/course-summary/course-summary.page.ts
```diff
@@ -302,6 +260,33 @@ export class CoreCourseSummaryPage implements OnInit, OnDestroy {
             await CoreCourseOptionsDelegate.getMenuHandlersToDisplay(this.course, refresh, this.useGuestAccess);
     }
 
+    /**
+     * Self enrol in a course.
```
dpalou, line 264: This JSDoc seems wrong, now this function seems more generic (and if I understood it right this function is for "view" enrols, not for self enrol, right?).

src/core/features/course/pages/course-summary/course-summary.page.ts
```diff
@@ -61,21 +59,21 @@ export class CoreCourseSummaryPage implements OnInit, OnDestroy {
     @ViewChild('courseThumb') courseThumb?: ElementRef;
 
     isEnrolled = false;
+
     canAccessCourse = true;
+    useGuestAccess = false;
```
dpalou, line 64: Now we can have other "view" enrol methods besides guest, so I guess we should rename this variable (and also other params of functions where this variable is passed).

src/core/features/course/services/course-helper.ts
```diff
@@ -622,19 +624,16 @@ export class CoreCourseHelperProvider {
             }
 
             // Check if guest access is enabled.
-            const enrolmentMethods = await CoreCourses.getCourseEnrolmentMethods(courseId, siteId);
-
-            const method = enrolmentMethods.find((method) => method.type === 'guest');
-
-            if (!method) {
+            const enrolmentMethods = await CoreEnrolHelper.getCourseEnrolmentMethods(courseId, 'guest', siteId);
```
dpalou, line 627: Until now the only way to access a course without being enrolled was guest access, but now there can be other methods (not in the app's core, but in custom apps or when we support site plugins). So maybe all the code related to guest should be refactored to check all "view" enrols and also use some new naming that doesn't include the guest word.

src/core/features/course/services/course-helper.ts
```diff
@@ -622,19 +624,16 @@ export class CoreCourseHelperProvider {
...
+            if (!enrolmentMethods) {
                 return { guestAccess: false };
             }
 
-            const info = await CoreCourses.getCourseGuestEnrolmentInfo(method.id);
+            const info = await AddonEnrolGuest.getGuestEnrolmentInfo(enrolmentMethods[0].id);
```
dpalou, line 632: We should try to avoid importing addons in core, it'd be better if we could obtain this info through the delegate.

src/addons/enrol/guest/services/guest.ts
```diff
@@ -0,0 +1,159 @@
...
+     */
+    async getGuestEnrolmentInfo(instanceId: number, siteId?: string): Promise<AddonEnrolGuestInfo> {
+        const site = await CoreSites.getSite(siteId);
+
+        const params: EnrolGuestGetInstanceInfoWSParams = {
+            instanceid: instanceId,
+        };
+
+        const preSets: CoreSiteWSPreSets = {
+            cacheKey: this.getGuestEnrolmentInfoCacheKey(instanceId),
```
dpalou, line 45: This functions was previously in courses and used the courses cache key. Please see my other comment about cache key to see a way to handle this.

src/core/features/courses/services/courses.ts
```diff
@@ -344,70 +345,10 @@ export class CoreCoursesProvider {
...
-    protected getCourseGuestEnrolmentInfoCacheKey(instanceId: number): string {
-        return ROOT_CACHE_KEY + 'guestinfo:' + instanceId;
-    }
-
-    /**
-     * Check if guest password validation WS is available on the current site.
-     *
-     * @returns Whether guest password validation WSget courses by field is available.
-     */
-    isValidateGuestAccessPasswordAvailable(): boolean {
```
dpalou, line 377: You should keep the functions isValidateGuestAccessPasswordAvailable and validateGuestAccessPassword as deprecated.

dpalou (changes requested):

src/core/features/courses/services/courses.ts
```diff
@@ -1323,42 +1253,10 @@ export class CoreCoursesProvider {
      * @param siteId Site ID. If not defined, use current site.
      * @returns Promise resolved if the user is enrolled. If the password is invalid, the promise is rejected
      *         with an object with errorcode = CoreCoursesProvider.ENROL_INVALID_KEY.
+     * @deprecated since 4.3, use AddonEnrolSelf.selfEnrol instead.
```
dpalou, line 1256: Shouldn't we recommend using the delegate method instead of using the addon directly?

src/core/features/enrol/services/enrol-delegate.ts
```diff
@@ -0,0 +1,221 @@
...
+import { CoreDelegate, CoreDelegateHandler } from '@classes/delegate';
+import { makeSingleton } from '@singletons';
+import { CoreCourseEnrolmentMethod } from '@features/courses/services/courses';
+
+/**
+ * Enrolment actions.
+ */
+export enum CoreEnrolAction {
+    BROWSER = 'browser', // User should use the browser to enrol. Ie. paypal
+    SELF = 'self', // User can enrol itself.
```
dpalou, line 25: "itself" shouldn't be used for people, I'd say "himself or herself".

src/core/features/enrol/services/enrol-delegate.ts
```diff
@@ -0,0 +1,221 @@
...
+    enrol?(method: CoreCourseEnrolmentMethod): Promise<boolean>;
+
+    /**
+     * Check if the user can access to the course.
+     * This method should be implemented in VIEW enrolment actions.
+     *
+     * @param method Course enrolment method.
+     * @returns Whether the user can access.
+     */
+    canAccess?(method: CoreCourseEnrolmentMethod): Promise<boolean>;
```
dpalou, line 68: The typing could be improved so only certain types of plugins need to define the functions. E.g. you can have:

```
export interface CoreEnrolViewHandler extends CoreEnrolHandler {
    enrolmentAction: CoreEnrolAction.VIEW;
    
    canAccess(method: CoreCourseEnrolmentMethod): Promise<boolean>;
    ...
}
```

This is pseudo code, I haven't tried it. But this way you can define specific functions for each action.

src/addons/enrol/guest/services/enrol-handler.ts
```diff
@@ -0,0 +1,156 @@
...
+            // Don't allow guest access if it requires a password if not supported.
+            if (!info.passwordrequired) {
+                noPasswordRequired = true;
+            } else {
+                passwordRequired = true;
+            }
+            if (noPasswordRequired && passwordRequired) {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
crazyserver: Nope, sorry!

src/core/features/enrol/services/enrol-helper.ts
```diff
@@ -0,0 +1,170 @@
...
+
+                return;
+            }
+
+            const action = CoreEnrolDelegate.getEnrolmentAction(type);
+
+            if (action === CoreEnrolAction.BROWSER) {
+                // Only add the browser action once.
+                addBrowserOption = true;
+            }
```
dpalou, line 74: This can be changed to a single line:

`addBrowserOption = addBrowserOption || action === CoreEnrolAction.BROWSER;`

src/core/features/enrol/services/enrol-helper.ts
```diff
@@ -0,0 +1,170 @@
...
+
+            switch (action) {
+                case CoreEnrolAction.BROWSER:
+                    hasBrowser = true;
+                    break;
+                case CoreEnrolAction.NOT_SUPPORTED:
+                    hasNotSupported = true;
+                    break;
+            }
+        });
```
dpalou, line 121: Why are you doing this before calling getCourseEnrolmentMethods? After calling it you're also checking for browser or not supported. And in here you don't know if the methods are enabled or not, right? You don't have `method.status`.

src/addons/enrol/self/services/enrol-handler.ts
```diff
@@ -0,0 +1,177 @@
...
+            // Don't allow self access if it requires a password if not supported.
+            if (!info.enrolpassword) {
+                noPasswordRequired = true;
+            } else {
+                passwordRequired = true;
+            }
+            if (noPasswordRequired && passwordRequired) {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
crazyserver: Yep it can happen, it will show two icons in the course list (as LMS does)

src/core/features/enrol/services/enrol-helper.ts
```diff
@@ -0,0 +1,170 @@
...
+        enrolmentMethods.forEach((method) => {
+            if (!CoreUtils.isTrueOrOne(method.status)) {
+                return;
+            }
+
+            const action = CoreEnrolDelegate.getEnrolmentAction(method.type);
+
+            switch (action) {
+                case CoreEnrolAction.SELF:
+                    // Only add the browser action once.
```
dpalou, line 137: This comment (and the one in VIEW) is wrong.

src/core/features/siteplugins/classes/handlers/enrol-handler.ts
```diff
@@ -0,0 +1,39 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { CoreLogger } from '@singletons/logger';
+import { CoreSitePluginsBaseHandler } from './base-handler';
+import { CoreEnrolAction, CoreEnrolHandler } from '@features/enrol/services/enrol-delegate';
+import { CoreSitePluginsContent, CoreSitePluginsEnrolHandlerData } from '@features/siteplugins/services/siteplugins';
+
+/**
+ * Handler to support a enrol using a site plugin.
```
dpalou, line 21: I'd put a `todo` or similar specifying that right now only browser methods can be supported.

src/core/features/siteplugins/services/siteplugins-helper.ts
```diff
@@ -800,6 +807,32 @@ export class CoreSitePluginsHelperProvider {
...
+     */
+    protected registerEnrolHandler(
+        plugin: CoreSitePluginsPlugin,
+        handlerName: string,
+        handlerSchema: CoreSitePluginsEnrolHandlerData,
+        initResult: CoreSitePluginsContent | null,
+    ): string | undefined {
+        const uniqueName = CoreSitePlugins.getHandlerUniqueName(plugin, handlerName);
+        const type = (handlerSchema.moodlecomponent || plugin.component).replace('enrol_', '');
+        const action = handlerSchema.enrolmentAction ?? CoreEnrolAction.BROWSER;
```
dpalou, line 827: IMO it doesn't make much sense to allow specifying an enrolmentAction because plugins cannot implement the required functions like "enrol", I'd always set this to BROWSER for now and I'd add a `todo` to support the rest of actions.

src/core/features/courses/services/courses.ts
crazyserver: isValidateGuestAccessPasswordAvailable has been recently created and not released.

dpalou (changes requested):

src/core/features/course/services/course-helper.ts
dpalou: After talking to Pau we decided it was best to use the "guest" word instead of "view". This way we don't need to refactor as much code, and in LMS they also use the guest word in generic classes like enrol_plugin (e.g. try_guestaccess).

dpalou (approved):

PR #3756, opened 2023-07-26
MOBILE-4272: Decouple workshop from initial bundle

NoelDeMartin:


src/addons/mod/workshop/services/handlers/prefetch.ts
```diff
@@ -12,401 +12,33 @@
...
-     */
-    async sync(module: CoreCourseAnyModuleData, courseId: number, siteId?: string): Promise<AddonModDataSyncResult> {
-        return AddonModWorkshopSync.syncWorkshop(module.instance, siteId);
-    }
+    Object.assign(lazyHandler, {
+        name: ADDON_MOD_WORKSHOP_PREFETCH_NAME,
+        modName: ADDON_MOD_WORKSHOP_PREFETCH_MODNAME,
+        component: ADDON_MOD_WORKSHOP_PREFETCH_COMPONENT,
+        updatesNames: ADDON_MOD_WORKSHOP_PREFETCH_UPDATE_NAMES,
+    });
```
dpalou, line 41: Why is this assign needed? Is it to be able to access these properties without having to create the instance?

I don't like the fact that we need to replicate the values, but maybe there's no other way.

src/addons/mod/workshop/services/handlers/prefetch-lazy.ts
```diff
@@ -0,0 +1,416 @@
...
+    ADDON_MOD_WORKSHOP_PREFETCH_MODNAME,
+    ADDON_MOD_WORKSHOP_PREFETCH_NAME,
+    ADDON_MOD_WORKSHOP_PREFETCH_UPDATE_NAMES,
+} from '@addons/mod/workshop/constants';
+
+/**
+ * Handler to prefetch workshops.
+ */
+@Injectable({ providedIn: 'root' })
+export class AddonModWorkshopPrefetchHandlerService extends CoreCourseActivityPrefetchHandlerBase {
```
dpalou, line 47: At first it made more sense to me to name this file just `prefetch`, and the file that loads this file lazily name it `prefetch-lazy`, but I think in modules we do it like you did here, the module that is loaded lazily we call it "-lazy", so for consistency I guess we should leave it like this.

dpalou (changes requested):

src/addons/mod/workshop/services/handlers/prefetch.ts
NoelDeMartin: Yes, these are the only values that are needed before initializing the async instance.

I also don't like the idea of replicating the values. I think we could find a way to avoid it, but it'll probably result in even more indirection and more complex code :/. Although maybe it's worth it, I'll explore it.

src/addons/mod/workshop/services/handlers/prefetch-lazy.ts
NoelDeMartin: Ok we can discuss the naming before having the final version, I'm also not 100% sure about it. I'll bring it up in some of our team meetings.

dpalou (approved):

PR #3765, opened 2023-08-01
MOBILE-4191 scorm: Check scormstandard to change CMI_STRING_256

alfonso-salces:


src/addons/mod/scorm/classes/data-model-12.ts
```diff
@@ -255,6 +255,11 @@ export class AddonModScormDataModel12 {
      * @param userData The user default data.
      */
     protected init(userData: AddonModScormUserDataMap): void {
+        if (this.scorm.scormStandard !== 0) {
```
dpalou, line 258:
```suggestion
        if (this.scorm.scormStandard === 0 || this.scorm.scormStandard === undefined) {
```
The value of the variables should be changed if the setting is disabled (=== 0). Also, if we don't have the setting because the Moodle site doesn't support returning it I think we decided to always apply the 64000 value so affected SCORMs can work in the app when using old sites.

src/addons/mod/scorm/services/scorm.ts
```diff
@@ -1124,6 +1124,12 @@ export class AddonModScormProvider {
             currentScorm.warningMessage = warning?.message;
         }
 
+        const scormOptions: Record<string, string> = CoreUtils.objectToKeyValueMap(response.options, 'name', 'value');
```
dpalou, line 1127: Is it really needed to specify `Record<string, string>`? I think the function should calculate it automatically. If it doesn't then I think you can specify it like this: `CoreUtils.objectToKeyValueMap<string>`.

dpalou (changes requested):

src/addons/mod/scorm/classes/data-model-12.ts
alfonso-salces: What do you think about use `!this.scorm.scormStandard` directly?
Maybe this is cleaner and cover nullish too.

dpalou: Sure, sounds good to me.

src/addons/mod/scorm/services/scorm.ts
alfonso-salces: Yeah, this is not really required so I will remove this typing. Thanks

dpalou (approved):

PR #3766, opened 2023-08-01
MOBILE-4362 behat: Improve BBB wait steps

NoelDeMartin:


local_moodleappbehat/tests/behat/behat_app.php
```diff
@@ -646,24 +666,46 @@ public function i_press_in_the_app(string $locator) {
...
-        $this->spin(function() {
-            $result = $this->runtime_js('pullToRefresh()');
+    public function i_pull_to_refresh_in_the_app(?string $locator = null) {
+        $timeout = 0;
+        $startTime = time();
 
-            if ($result !== 'OK') {
-                throw new DriverException('Error pulling to refresh - ' . $result);
+        if (!is_null($locator)) {
+            $timeout = 6000;
```
dpalou, line 677: I think `time()` returns time in seconds, so this 6000 means it will wait 1h and 40 minutes?

dpalou (changes requested):

local_moodleappbehat/tests/behat/behat_app.php
NoelDeMartin: You're right, actually I should have noticed because if I wanted to wait 60 seconds in milliseconds it should have been 60000 xD. Anyhow, it's fixed now.

dpalou (approved):

PR #3767, opened 2023-08-01
MOBILE-4401 wiki: Fix selected subwiki when wiki is empty

dpalou:
This commit avoids a redundant error and also allows the create button to appear if the user has permissions to create the first page

### DIFF
--- a/src/addons/mod/wiki/components/index/index.ts
+++ b/src/addons/mod/wiki/components/index/index.ts
@@ -343,41 +343,18 @@ export class AddonModWikiIndexComponent extends CoreCourseModuleMainActivityComp
      * @param subwiki Subwiki.
      */
     protected async fetchSubwikiPages(subwiki: AddonModWikiSubwiki): Promise<void> {
-        const subwikiPages = await AddonModWiki.getSubwikiPages(subwiki.wikiid, {
-            groupId: subwiki.groupid,
-            userId: subwiki.userid,
-            cmId: this.module.id,
-        });
+        const subwikiPages = subwiki.id <= 0 ?
+            [] :
+            await AddonModWiki.getSubwikiPages(subwiki.wikiid, {
+                groupId: subwiki.groupid,
+                userId: subwiki.userid,
+                cmId: this.module.id,
+            });
 
-        if (!this.currentPage) {
-            if (!this.pageTitle) {
-                // No page specified, search first page.
-                const firstPage = subwikiPages.find((page) => page.firstpage );
-                if (firstPage) {
-                    this.currentPage = firstPage.id;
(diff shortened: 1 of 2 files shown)


src/addons/mod/wiki/components/index/index.ts
```diff
@@ -356,6 +358,8 @@ export class AddonModWikiIndexComponent extends CoreCourseModuleMainActivityComp
                 if (firstPage) {
                     this.currentPage = firstPage.id;
                     this.pageTitle = firstPage.title;
+                } else {
```
alfonso-salces, line 361: Maybe this is a good moment to improve the cognitive complexity using guard clauses.

I think that we can create a new method to improve the readability, something like that is fine.

    /**
     * Set current page.
     *
     * @param subwikiPages list of subwikis to find in.
     */
    setCurrentPage(subwikiPages: AddonModWikiSubwikiPage[]): void {
        if (!this.currentPage) {
            return;
        }

        if (this.pageTitle) {
            const page = subwikiPages.find((page) => page.title === this.pageTitle );

            if (!page) {
                return;
            }

            this.currentPage = page.id;
        }

        const firstPage = subwikiPages.find((page) => page.firstpage );

        if (!firstPage) {
            return;
        }

        this.currentPage = firstPage.id;
        this.pageTitle = firstPage.title;
    }


With that, we can remove the `if (!this.currentPage)` completly

alfonso-salces (changes requested):

src/addons/mod/wiki/components/index/index.ts
```diff
@@ -343,11 +343,13 @@ export class AddonModWikiIndexComponent extends CoreCourseModuleMainActivityComp
...
-            userId: subwiki.userid,
-            cmId: this.module.id,
-        });
+        const subwikiPages = subwiki.id <= 0 ?
+            [] :
+            await AddonModWiki.getSubwikiPages(subwiki.wikiid, {
+                groupId: subwiki.groupid,
+                userId: subwiki.userid,
+                cmId: this.module.id,
+            });
```
alfonso-salces, line 353:
```suggestion
setCurrentPage(subwikiPages);
```

src/addons/mod/wiki/components/index/index.ts
dpalou: Thanks Alfonso, I applied your suggestion. Also, I realised that, with my new code, the block of code after "`const dbPages`" wasn't needed anymore because now we're always setting `pageTitle` to `this.wiki.firstpagetitle` if needed so I removed a few lines there too.

alfonso-salces (approved):

PR #3768, opened 2023-08-02
MOBILE-4191 scorm: Make options field optional

alfonso-salces:


dpalou (approved):

src/addons/mod/scorm/services/scorm.ts
```diff
@@ -1871,7 +1873,7 @@ export type AddonModScormGetScormsByCoursesWSParams = {
  * Data returned by mod_scorm_get_scorms_by_courses WS.
  */
 export type AddonModScormGetScormsByCoursesWSResponse = {
-    options: AddonModScormOptions[];
+    options?: AddonModScormOptions[];
```
NoelDeMartin, line 1876: I think we should add a comment here saying that this is available from 4.3+

dpalou: Yes, agree. @alfonso-salces you can follow this format:

https://github.com/moodlehq/moodleapp/blob/main/src/addons/badges/services/badges.ts#L162

dpalou (approved):

PR #3772, opened 2023-08-21
fix(core-block): refresh component 

phyr0s:
Component is not refreshing the variable this.block in the core-block component so when enter when loadAllCourses to get this.allCourses --> in here this.block is missing --> const showCategories = this.block.configsRecord?.displaycategories?.value === '1';

So the variable showCategories is false and i lot of cases should be true but by default just put false.

If you enter in one course and then go to main page variables like categoryname are lost forever.


dpalou:
Hi phyros,

all Pull Requests to this project need to be linked to an issue in https://tracker.moodle.org. Please create an issue in there using the project "Moodle app (MOBILE)", and then include the issue name in your commit. All commits need to follow the format "ISSUE component: Description", for example:

MOBILE-4370 course: Make section availability clickable

On the other hand, I don't understand the problem you have. Looking at the solution I guess you're setting _loadedBlock_ to undefined to force re-creating the myoverview block, but I don't understand why do you need to do this and in which case. Can you please provide some detailed steps to reproduce the problem, explaining what is happening and what do you expect to happen?

Cheers,
Dani

phyr0s:
Will reopen with the tracker id + more changes and detail

PR #3776, opened 2023-08-31
Mobile 4201

crazyserver:


src/core/features/login/login.scss
```diff
@@ -112,6 +112,13 @@
             display: none;
         }
     }
+
+    ion-page.can-go-back {
+        ion-button.ion-back-button {
+            display: none;
+        }
+    }
+
```
dpalou, line 156: This style seems a bit contradictory. If the page allows going back (can-go-back) you hide the back button?

crazyserver: This is hiding the automatically added back button. I couldn't find a better way to manage or change the behaviour of this button, so I added a fake one, and remove the original one to avoid duplicates.

crazyserver: Sorry, this hides our custom button when the original is shown. Since the original has been removed, there's no need of those styles either.

dpalou (approved):

PR #3777, opened 2023-09-01
Mobile 4405 - Implement auto logout

alfonso-salces:


src/core/services/storage.ts
```diff
@@ -0,0 +1,205 @@
...
+        return new CoreStorageService(async () => {
+            const siteId = await CoreSites.getStoredCurrentSiteId();
+            const site = await CoreSites.getSite(siteId);
+            const table = await getStorageTable(site.getDb());
+
+            return table;
+        });
+    }
+
+    static forSite(site: CoreSite): AsyncInstance<CoreStorageService> {
```
dpalou, line 148: forCurrentSite and forSite don't have JSDoc.

src/core/services/storage.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    static forCurrentSite(): CoreStorageService {
+        return new CoreStorageService(async () => {
+            const siteId = await CoreSites.getStoredCurrentSiteId();
+            const site = await CoreSites.getSite(siteId);
+            const table = await getStorageTable(site.getDb());
+
+            return table;
+        });
+    }
```
dpalou, line 146: This function will create a new instance every time is executed, it's not using SERVICE_INSTANCES. Shouldn't it use SERVICE_INSTANCES to avoid creating the same instance several times?

src/core/services/sites.ts
```diff
@@ -1421,6 +1422,25 @@ export class CoreSitesProvider {
...
+                return;
+            }
+
+            if (autoLogoutType === CoreAutoLogoutType.CUSTOM) {
+                await CoreAutoLogout.handleSessionClosed(autoLogoutTime, site);
+
+                return;
+            }
+
+            await CoreAutoLogout.handleAppClosed(site);
```
dpalou, line 1441: IMO it'd be better to move this code to a new function so the restoreSession function is more readable.

src/core/classes/site.ts
```diff
@@ -835,7 +835,7 @@ export class CoreSite {
 
             if (CoreUtils.isExpiredTokenError(error)) {
                 // Session expired, trigger event.
-                CoreEvents.trigger(CoreEvents.SESSION_EXPIRED, {}, this.id);
+                CoreEvents.trigger(CoreEvents.SESSION_EXPIRED, { redirectPath: Router.url }, this.id);
```
dpalou, line 838: This commit can give us problems. I see 2 right now:

1. You need to pass the query parameters, not just the path. Because some parameters aren't included in the URL.
2. When the current page is "restored", the user will be sent to the "final" page, he won't have the whole history. And the only way to go back to a "previous" page will be to click the bottom tab to go back to the root page. IMO this can be confusing, some users might not know that clicking the tab will redirect to the root. Also, loading some pages directly can cause problems as Noel explained in the DevJam webinar (e.g. site plugin pages might not have some data loaded if they're opened directly).

IMO this change deserves its own issue and a better analysis.

dpalou (changes requested):

dpalou:
Hi Alfonso, I decided to pass the review to Noel because he had some concerns about the storage implementation and he knows this part better than me. I sent you some comments about things I saw while reviewing the patch.

src/core/classes/site.ts
alfonso-salces: Hi Dani, well in that case im gonna drop this commit and I will open an issue soon.
Thanks for the review!

NoelDeMartin (approved):

dpalou (approved):

PR #3778, opened 2023-09-04
Mobile 4406 - Hide download icon in resources type label

alfonso-salces:


NoelDeMartin:
@alfonso-salces I see that the development of [this issue](https://tracker.moodle.org/browse/MOBILE-4406) is still in progress, should this be a draft instead?

PR #3780, opened 2023-09-05
Mobile 3839

dpalou:


cordova-plugin-moodleapp/src/ts/plugins/SecureStorage.ts
```diff
@@ -0,0 +1,76 @@
...
+export class SecureStorage {
+
+    /**
+     * Get one or more values.
+     *
+     * @param names Names of the values to get.
+     * @param collection The collection where the values are stored.
+     * @returns Object with name -> value.
+     */
+    async get(names: string | string[], collection: string): Promise<Record<string, string | null>> {
```
NoelDeMartin, line 27:
```suggestion
    async get(names: string | string[], collection: string): Promise<Record<string, string>> {
```

It seems like both Android and iOS already filter out null values, so it should be fine to type it without null.

cordova-plugin-moodleapp/src/ts/plugins/SecureStorage.ts
```diff
@@ -0,0 +1,76 @@
...
+        });
+    }
+
+    /**
+     * Delete one or more values.
+     *
+     * @param names Names to delete.
+     * @param collection The collection where to delete the values.
+     */
+    async delete(names: string[], collection: string): Promise<void> {
```
NoelDeMartin, line 55:
```suggestion
    async delete(names: string | string[], collection: string): Promise<void> {
```

cordova-plugin-moodleapp/src/ios/SecureStorage.m
```diff
@@ -0,0 +1,165 @@
...
+            });
+        } else {
+            query[(id)kSecValueData] = newValue;
+            status = SecItemAdd((CFDictionaryRef)query, nil);
+        }
+
+        if (status != errSecSuccess) {
+            NSLog(@"Error storing value for %@ in collection %@. Status: %d", name, collection, (int) status);
+            hasErrors = true;
+        }
```
NoelDeMartin, line 82: It seems like errors can produce partial writes (for example, if I'm writing 5 strings and only one fails, the other 4 will be written).

I think we should either return which values were written or roll back the partial writes. I prefer the second one, but I'm not sure how easy it'd be to implement correctly. In case we follow the first approach, we should always be aware of this quirk when calling this method, and make sure that the stored values don't have any inconsistencies (this one may be a lot more work in the long run, that's why I prefer the second approach).

It seems like this is not a problem in Android, but it's worth double-checking. What are the possible causes for this failing, anyways?

src/core/features/native/services/native.ts
```diff
@@ -32,10 +32,6 @@ export class CoreNativeService {
      * @returns Plugin instance.
      */
     plugin<Plugin extends keyof MoodleAppPlugins>(plugin: Plugin): AsyncInstance<MoodleAppPlugins[Plugin]> | null {
```
NoelDeMartin, line 34: We have removed the `CorePlatform.isAndroid()` check here, but we should actually add `CorePlatform.isMobile()` instead of removing it.

Additionally, if we are ever going to have some partial plugins again (working only in one OS), we should probably do something about it here. But I think it's ok for now, because the only native plugin (SecureStorage) is available in both OSes.

src/core/services/sites.ts
```diff
@@ -830,7 +833,18 @@ export class CoreSitesProvider {
...
             oauthId,
-        });
+        };
+
+        if (CorePlatform.isMobile()) {
+            promises.push(this.storeTokensInSecureStorage(id, token, privateToken));
+
+            site.token = '';
+            site.privateToken = '';
+        }
```
NoelDeMartin, line 843: I understand why real "secure storage" can only be implemented in Android/iOS, but the code resulting from this is more complicated than it should. I think we should be able to have a "secure storage" interface for a browser environment as well (even if it's not actually secure), and use that one to simplify the code everywhere. Under the hood, it could use localStorage or something like that, and it'd make it easier to debug some potential errors because the data is stored the same way in any environment,

cordova-plugin-moodleapp/src/ts/plugins/SecureStorage.ts
```diff
@@ -0,0 +1,76 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/**
+ * Allows retrieving and storing items in a secure storage.
+ */
+export class SecureStorage {
```
NoelDeMartin, line 18: Looking at the PR I have a suggestion about this, but it may be too much work to refactor now (although it may be better in the long run, it depends how much we're going to use this secure storage moving forward).

Instead of having this collections with name/value pairs, I think we should be able to just have name/value pairs. With the difference that values could be anything, not just strings. This can be implemented by serializing values to JSON in the JS layer, and so the native layers only need to store plain strings as we're doing now. Among other things, I think this would be a better approach because it follows the same approach we are already using for config storage. Eventually, we may even have everything under a single interface, making it possible to write code compatible with any storage backend.

NoelDeMartin (changes requested):

cordova-plugin-moodleapp/src/ios/SecureStorage.m
dpalou: You're right, I implemented the rollback code for iOS (I squashed it into the iOS commit). I don't know when can it fail, I wasn't able to make it fail. The only case where the app crashed was if I tried to insert a null value (the whole app crashed, it didn't return an error), that won't be possible anymore.

src/core/features/native/services/native.ts
dpalou: Now we have a mock implementation for browser so the isMobile check isn't needed either.

src/core/services/sites.ts
dpalou: I added a new commit to support SecureStorage in browser too (using localStorage, so it's not secure). I changed the code in sites provider to be the same in all platforms.

cordova-plugin-moodleapp/src/ts/plugins/SecureStorage.ts
dpalou: The data we're storing right now belong to sites, it's similar to the "Site database", so I think it makes sense to divide them in "sites" (I named them collections to make it more generic).  This makes it easier to delete all the data from one site. CoreConfig is used to store data for the whole app, not for sites, so I think the use case here is a bit different.

If we want to store data for the whole app (not for a site) we could use an empty collection name ("") or a specific name (e.g. MoodleApp). So if we need it we could implement a class with the CoreConfig interface, and behind the scenes it can easily use the current SecureStorage implementation.

dpalou:
Summary of the changes done so it's easier to review:

- I implemented rollback for iOS (it's in the iOS commit). I had to abstract some code in new functions so I recommend reviewing the whole file again.
- If you try to store undefined or null values they will be ignored (it's controlled in the TS code). At first I thought about throwing an error, but the privateToken is optional (some sites might not have it) so I decided not store it in that case. If you try to retrieve it then it won't be found so the returned object will not contain the property (it will be undefined).
- In sites provider now the code is the same for all platforms.
- Added a new commit to copy the plugin JS code to iOS platform and plugins folder.
- Added a new commit to support SecureStorage in browser.

NoelDeMartin (approved):

PR #3784, opened 2023-09-12
Mobile 4410

dpalou:


src/core/services/app.ts
```diff
@@ -671,6 +675,36 @@ export class CoreAppProvider {
...
     }
 
+    /**
+     * Get install referrer URL (Android only).
+     * This function will try to retrieve the siteurl supplied as referrer to Google Play.
+     *
+     * @returns Referrer URL, undefined if not supported, already consumed or no referrer URL.
+     */
+    async getInstallReferrerUrl(): Promise<string | undefined> {
+        if (!CorePlatform.isAndroid()) {
```
NoelDeMartin, line 685: I reckon we're doing this because the installReferrer native implementation is only available in Android. But I think that knowing this should be responsibility of the `CoreNative` class, rather than all code consuming native code.

Before this PR, the `CoreNative.plugin` method would always work, because secureStorage is available in all the platforms (including non-native with mocks). But now that we're back to having partial functionality depending on the platform, I think we should refactor the method to return an optional response again. Otherwise, it would return an asyncInstance where calling any method returns an error.

I guess we could also have a different method, like `isPluginAvailable` or something, to check before using the native plugin. But I don't like it as much because it also leaves the responsibility to the consumers of the API to call it. I prefer designing the API in a way that is not possible to do the wrong thing.

src/core/services/app.ts
```diff
@@ -49,6 +51,8 @@ import { CoreMainMenuProvider } from '@features/mainmenu/services/mainmenu';
 @Injectable({ providedIn: 'root' })
 export class CoreAppProvider {
 
+    protected static readonly INSTALL_REFERRER_CONSUMED = 'install_referrer_consumed';
```
NoelDeMartin, line 54: I think we should move all this code elsewhere, maybe to `CoreReferrer` or something? This file is already too big, and it's imported everywhere throughout the app (it's actually one of the main culprits for the lack of lazy loading). So I think it should be as lean as possible from now on, and ideally we'll eventually even make it leaner.

In this case, I don't think we'd want to load the referrer code every time the app starts. Ideally, that would only need to happen if users open the login screen, which should already be on a lazy chunk. So users who are already logged in would never load this code. Maybe you can use this opportunity to play around with the [bundle analyzer](https://github.com/NoelDeMartin/moodleapp-bundle-analysis) to make sure that this is actually happening.

src/core/services/app.ts
```diff
@@ -671,6 +675,36 @@ export class CoreAppProvider {
...
+     * This function will try to retrieve the siteurl supplied as referrer to Google Play.
+     *
+     * @returns Referrer URL, undefined if not supported, already consumed or no referrer URL.
+     */
+    async getInstallReferrerUrl(): Promise<string | undefined> {
+        if (!CorePlatform.isAndroid()) {
+            return;
+        }
+
+        // Only get the referrer URL once, it's only needed when the app is installed.
```
NoelDeMartin, line 689: I'm not sure about this, it seems like this is a concern for the login screen, since we only want to prepopulate it with the referrer url on the first launch. But the referrer url itself doesn't go away. I would add that restriction (of only using it on the first launch) in the login page or related code.

src/core/features/login/pages/site/site.ts
```diff
@@ -98,8 +98,21 @@ export class CoreLoginSitePage implements OnInit {
...
             url = await this.initSiteSelector();
-        } else if (CoreConstants.CONFIG.enableonboarding && !CorePlatform.isIOS()) {
-            this.initOnboarding();
+        } else {
+            url = await this.getInstallReferrerUrl() ?? '';
+
+            const showOnboarding = CoreConstants.CONFIG.enableonboarding && !CorePlatform.isIOS();
+
+            if (url) {
+                this.connect(new Event('click'), url);
```
NoelDeMartin, line 107: It seems like we're passing this `new Event` just because the `connect` method is stopping it when it comes from the UI, but in this case it is not really necessary. Maybe we should refactor the method to remove this parameter, and stop the propagation in a different method that is called from the UI (onConnect or something).

NoelDeMartin (changes requested):

src/core/services/app.ts
dpalou: I cannot think of any use case where we want to use the URL more than once, so I think it's ok to encapsulate this logic here. That way the page code is simpler.

src/core/features/login/pages/site/site.ts
dpalou: Instead of creating a new function, I made the event parameter to be optional.

cordova-plugin-moodleapp/src/ts/plugins/InstallReferrer.ts
```diff
@@ -0,0 +1,39 @@
...
+        });
+    }
+
+}
+
+export type InstallReferrerResult = {
+    referrer: string;
+    clickTime: number;
+    appInstallTime: number;
+    instantExperienceLaunched: boolean;
```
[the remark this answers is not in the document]
dpalou: We aren't using them, and I'm not sure if we'll ever do, but since the Android API returned them I decided to included them too since it was easy to do and IMO it doesn't hurt to have them.

src/core/services/app.ts
NoelDeMartin: Ok, I think it's fine to leave it like this now that we are calling it "consume". That should tell anyone calling this API in the future that this is not a normal getter.

NoelDeMartin (approved):

PR #3807, opened 2023-10-05
Mobile 4362

crazyserver:


src/core/services/utils/dom.ts
```diff
@@ -1805,7 +1805,11 @@ export class CoreDomUtilsProvider {
      */
     async openPopover<T = void>(options: OpenPopoverOptions): Promise<T | undefined> {
 
+        options = Object.assign({
+            showBackdrop: false,
+        }, options);
```
dpalou, line 1810:
```suggestion
        options = {
            showBackdrop: false,
            ...options,
        };
```
I prefer this because `Object.assign` returns type `any`, while the spread operator returns the right type.

dpalou, line 1810: Also related to this, I see in the CSS that you always set the backdrop to transparent. Do we want to show the backdrop at all then? If we don't want to show the backdrop we can just change the order to force `showBackdrop` being false.

src/theme/theme.base.scss
```diff
@@ -1869,8 +1869,10 @@ video::-webkit-media-text-track-display {
 }
 
 ion-modal {
-    .modal-wrapper {
-        --border-radius: var(--modal-radius);
+    @media only screen and (min-width: 768px) and (min-height: 600px) {
+        .modal-wrapper {
+            --border-radius: var(--modal-radius);
+        }
```
dpalou, line 1878: If I understand this right, the `border-radius` will only be applied to modals in tablets because in phones the modals are fullscreen by default. But what about `.core-modal-lateral` and `.core-modal-fullscreen`? These modals are bigger in tablet too. I see they have `--modal-radius: 0px;` but I'm not sure if this is the same as `--border-radius`.

src/core/features/user/pages/about/about.ts
```diff
@@ -279,4 +299,72 @@ export class CoreUserAboutPage implements OnInit, OnDestroy {
...
     }
 
+    /**
+     * Translates legacy timezone names.
+     *
+     * @param tz Timezone name.
+     * @returns Readable timezone name.
+     */
+    protected getLegacyTimezone(tz: string): string {
+        const legacyTzs = {
```
dpalou, line 305: Since this is a constant IMO it should be declared in the class using `protected static readonly`.

src/core/features/user/pages/about/about.ts
```diff
@@ -279,4 +299,72 @@ export class CoreUserAboutPage implements OnInit, OnDestroy {
...
+            '': 'Australia/Perth', // Fallback on default timezone.
+            '-13.0': 'Australia/Perth', // Fallback on default timezone.
```
dpalou, line 307: Is this comment about default timezone right? It looks ok for '' but '-13' doesn't seem a default timezone.

dpalou (changes requested):

src/core/features/user/pages/about/about.ts
crazyserver: Code has been copied from moodle LMS..

src/core/services/utils/dom.ts
crazyserver: Code removed, it will be solved using css

crazyserver: Code removed, it will be solved using css

src/core/features/user/pages/about/about.ts
crazyserver: Code moved to CoreUserHelper and changes applied

src/theme/theme.base.scss
crazyserver: That was working but I improved it a bit using --border-radius directly.

dpalou (approved):

PR #3809, opened 2023-10-09
Mobile 4362

dpalou:


NoelDeMartin (dismissed):
I'm a bit wary of the SCORM screenshots, I'm not sure if they'll be the same across runs. Also, it seems like two of them are the same. If we're testing something like the iframe width we could test it with a javascript assertion instead.

But I'm ok with merging this for now, if they turn out to be troublesome we can remove them.

dpalou:
IMO the screenshots shouldn't change unless we do some deliberate change in the app, the package itself hasn't changed in 5 years. But I guess we'll see ^^'

About 2 screenshots being the same, that's because there are 2 different configurations that have the same result in the app. Since we don't support percentages in width/height, when that happens we display the iframe in full size like when you use the "Current window" configuration, that's why the screenshots are exactly the same.

NoelDeMartin (approved):

PR #3813, opened 2023-10-10
MOBILE-4407 url: Use Url parsing to determine last file

crazyserver:


NoelDeMartin (approved):
I'm not a fan of the method `getLastFileWithoutParams`, in name nor description. But that was already there, so I guess it isn't related to this PR so I'm going ahead with the merge.

PR #3822, opened 2023-10-16
MOBILE-4428 quiz: Fix adaptive behaviour

NoelDeMartin:


dpalou:
Thanks Noel,

the code looks good, but I looked for all the usages of `addBehaviourButton` and I saw it's also used in `searchBehaviourButton`, which is used in `extractQbehaviourRedoButton`. In this function we're also looking for `input[type="submit"]`. Maybe we should change it in there too even if it's not needed right now to prevent this feature to be broken if they change that input to a button like they did in MDL-78874. What do you think? I'm happy to integrate the patch if you prefer to leave it like this.

NoelDeMartin:
Thanks @dpalou for taking a look. I've followed your suggestion and I've also added a couple more tests to cover other scenarios, please take another look.

PR #3825, opened 2023-10-17
Mobile 4362

crazyserver:


src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -95,11 +106,16 @@ export class CoreCourseListModTypePage implements OnInit {
...
-                    if (!CoreCourseHelper.canUserViewModule(mod, section) || !CoreCourse.moduleHasView(mod)) {
+                    if (!CoreCourseHelper.canUserViewModule(mod, section) ||
+                        !CoreCourse.moduleHasView(mod) ||
+                        mod.visibleoncoursepage === 0) {
                         // Ignore this module.
                         return false;
                     }
 
+                    // Do not show indentation on this page.
+                    mod.indent = 0;
```
NoelDeMartin, line 117: Is there a way to avoid this mutation? For example, passing a `showIndent="false"` or something.

I don't like the idea of "lying" about this module. It does have an indent, we just don't want to show it. Doing this may create some unintended consequences in the future, for example if we rely on the `indent` property for something else. I'm aware of YAGNI, but it's also important to keep a proper data model rather than patching it for our convenience.

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -34,10 +34,14 @@ import { CoreAnalytics, CoreAnalyticsEventType } from '@services/analytics';
...
 
+    static readonly LOAD_MORE_ACTIVITIES = 20; // How many activities should load each time showMoreActivities is called.
+
     sections: CoreCourseSection[] = [];
     title = '';
     loaded = false;
-    courseId?: number;
+    courseId = 0;
+    canLoadMore = false;
+    showSectionId = 0;
```
NoelDeMartin, line 44: This should be called `showSectionIndex`, or even better `lastShownSectionIndex`, `shownSectionsLength`, etc.

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -34,10 +34,14 @@ import { CoreAnalytics, CoreAnalyticsEventType } from '@services/analytics';
 })
 export class CoreCourseListModTypePage implements OnInit {
 
+    static readonly LOAD_MORE_ACTIVITIES = 20; // How many activities should load each time showMoreActivities is called.
```
NoelDeMartin, line 37:
```suggestion
    private static readonly PAGE_LENGTH = 20; // How many activities should load each time showMoreActivities is called.
```

Besides the naming, this could be a constant declared in the function where it is used, given that it's only used in one place at the moment.

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -125,19 +141,64 @@ export class CoreCourseListModTypePage implements OnInit {
             const result = await CoreCourseHelper.addHandlerDataForModules(sections, this.courseId);
 
             this.sections = result.sections;
+
+            this.canLoadMore = false;
```
dpalou, line 145: This line isn't needed because you do it again in showMoreActivities.

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -125,19 +141,64 @@ export class CoreCourseListModTypePage implements OnInit {
...
+            this.showSectionId = 0;
+            this.showMoreActivities();
         } catch (error) {
             CoreDomUtils.showErrorModalDefault(error, 'Error getting data');
         }
     }
 
+    /**
+     * Show more activities (only used when showing all the sections at the same time).
```
dpalou, line 154: This comment says "only used when showing all the sections at the same time". But looking at the code, can we see only 1 section?

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -125,19 +141,64 @@ export class CoreCourseListModTypePage implements OnInit {
...
+     * @param infiniteComplete Infinite scroll complete function. Only used from core-infinite-loading.
+     */
+    showMoreActivities(infiniteComplete?: () => void): void {
+        this.canLoadMore = false;
+
+        const sections = this.sections || [];
+        let modulesLoaded = 0;
+        let i: number;
+        for (i = this.showSectionId + 1; i < sections.length; i++) {
+            if (!sections[i].hasContent || !sections[i].modules) {
```
NoelDeMartin, line 165: I see a couple of problems of following this approach.

First, it makes the code beyond more complicated (rather than just checking the length of the sections to calculate `canLoadMore`, we have to do yet another loop).

Also, it creates the possibility to load 0 items. Imagine the case that we have 20 modules in a row without content. It's very unlikely, but better to make it impossible.

Rather than doing this, I think it'd be better to filter these out when we assign the sections array in the `fetchData` method. Actually, it seems like we are already filtering them so maybe we could just move this additional check in there.

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -125,19 +141,64 @@ export class CoreCourseListModTypePage implements OnInit {
...
+     *
+     * @param infiniteComplete Infinite scroll complete function. Only used from core-infinite-loading.
+     */
+    showMoreActivities(infiniteComplete?: () => void): void {
+        this.canLoadMore = false;
+
+        const sections = this.sections || [];
+        let modulesLoaded = 0;
+        let i: number;
+        for (i = this.showSectionId + 1; i < sections.length; i++) {
```
dpalou, line 164: In the first execution, `i` will be 1 so you will always ignore the section 0. Do we really want this?

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -125,19 +141,64 @@ export class CoreCourseListModTypePage implements OnInit {
...
+            // Check if any of the following sections have any content.
+            let thereAreMore = false;
+            for (i++; i < sections.length; i++) {
+                if (sections[i].hasContent && sections[i].modules && sections[i].modules?.length > 0) {
+                    thereAreMore = true;
+                    break;
+                }
+            }
+            this.canLoadMore = thereAreMore;
+        }
```
NoelDeMartin, line 189:
```suggestion
        this.canLoadMore = sections.slice(i).some(section => sections[i].hasContent && sections[i].modules && sections[i].modules.length > 0);
```

If you follow my suggestion above, this will be removed. But if you don't, that's a better approach.

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -125,19 +141,64 @@ export class CoreCourseListModTypePage implements OnInit {
...
+
+            modulesLoaded += sections[i].modules.length;
+
+            if (modulesLoaded >= CoreCourseListModTypePage.LOAD_MORE_ACTIVITIES) {
+                break;
+            }
+        }
+
+        this.showSectionId = i;
+        this.canLoadMore = i < sections.length;
```
dpalou, line 177:
```suggestion
        this.canLoadMore = i < sections.length - 1;
```
Shouldn't this code be like this? If I'm not wrong, if `i` is `sections.length - 1` it means that the last section was already treated so there are no more to treat. Your code handles it properly in the next if, but IMO there's no need to in this case.

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -125,19 +141,64 @@ export class CoreCourseListModTypePage implements OnInit {
...
+
+        if (this.canLoadMore) {
+            // Check if any of the following sections have any content.
+            let thereAreMore = false;
+            for (i++; i < sections.length; i++) {
+                if (sections[i].hasContent && sections[i].modules && sections[i].modules?.length > 0) {
```
dpalou, line 183:
```suggestion
                if (sections[i].hasContent && sections[i].modules?.length > 0) {
```

src/core/features/course/pages/list-mod-type/list-mod-type.ts
dpalou, line 44:
```suggestion
    showSectionIndex = 0;
```
I think this name is better since this is not the section ID, it's the index in the array of sections.

dpalou (changes requested):

src/core/features/course/pages/list-mod-type/list-mod-type.ts
```diff
@@ -125,19 +138,46 @@ export class CoreCourseListModTypePage implements OnInit {
...
+        for (i = this.lastShownSectionIndex + 1; i < this.sections.length; i++) {
+            modulesLoaded += this.sections[i].modules.length;
+
+            if (modulesLoaded >= CoreCourseListModTypePage.PAGE_LENGTH) {
+                break;
+            }
+        }
+
+        this.lastShownSectionIndex = i;
+        this.canLoadMore = i < this.sections.length;
```
NoelDeMartin, line 168:
```suggestion
        let modulesLoaded = 0;

        while (this.lastShownSectionIndex < this.sections.length - 1 && modulesLoaded >= CoreCourseListModTypePage.PAGE_LENGTH) {
            this.lastShownSectionIndex++;

            modulesLoaded += this.sections[this.lastShownSectionIndex].modules.length;
        }

        this.canLoadMore = this.lastShownSectionIndex < this.sections.length - 1;
```

This simplifies the code and takes care of @dpalou's open discussion.

dpalou (dismissed):

NoelDeMartin (approved):

PR #3831, opened 2023-10-20
MOBILE-4405 sites: Fix sites older than 4.3 autologout

alfonso-salces:


NoelDeMartin (approved):

dpalou:
Hi guys, I'm a bit late to the party but I was just thinking that maybe it would be better to treat this in the `refreshListeners` function. With the current code, the app will set the intervals for sites that don't support auto-logout, when they should be treated as if auto logout was `CoreAutoLogoutType.NEVER`. And maybe we can detect it just by checking that the setting is undefined instead of checking the site version. What do you think?

If it's treated there then maybe there's no need to treat it in `handleAutoLogout`.

PR #3835, opened 2023-10-23
MOBILE-4362 quiz: Fix wrong renderization of ddwtos in some cases

dpalou:


NoelDeMartin (approved):
I'm not a fan of the `nextTicks(5)`. But considering how messy all this file is, I think it's fine to leave it just in case :sweat_smile:.

PR #3844, opened 2023-11-07
Mobile 4459

dpalou:


src/core/services/sites.ts
```diff
@@ -2195,6 +2209,7 @@ export type CoreSiteBasicInfo = {
     badge?: number; // Badge to display in the site.
     siteHomeId?: number; // Site home ID.
     loggedOut: boolean; // If Site is logged out.
+    isDemoModeSite: boolean;
```
NoelDeMartin, line 2212:
```suggestion
    demoMode: boolean;
```

I think it's better to use this nomenclature to distinguish between booleans and methods. Whilst reviewing the code, I saw this in a couple of templates and initially I though it was a typo missing the parentheses. Angular won't complain if a method is used without a call, given that it's always truthy. And we are using this same name for methods elsewhere in the codebase, so I think it's better to keep them distinct to avoid confusion.

src/core/components/site-list/site-list.ts
```diff
@@ -0,0 +1,80 @@
...
+ *         <!-- Content to be placed in the item.
+ *     </ng-template>
+ * </core-site-list>
+ */
+@Component({
+    selector: 'core-site-list',
+    templateUrl: 'site-list.html',
+    styleUrls: ['site-list.scss'],
+})
+export class CoreSiteListComponent<T extends CoreSiteBasicInfo> {
```
NoelDeMartin, line 42:
```suggestion
export class CoreSitesListComponent<T extends CoreSiteBasicInfo> {
```

I'm aware that this change requires modifying a lot of code and filenames, sorry :sweat_smile:.

src/assets/storybook/sites/schoolbarbara.ts
```diff
@@ -0,0 +1,32 @@
...
+        version: '2022041900',
+        sitename: 'School',
+        username: 'barbara',
+        firstname: 'Barbara',
+        lastname: 'Gardner',
+        fullname: 'Barbara Gardner',
+        lang: 'en',
+        userid: 1,
+        siteurl: 'https://campus.example.edu',
+        userpictureurl: 'assets/storybook/avatars/barbara.png',
```
NoelDeMartin, line 29: Would it be possible to use a url here instead of a real asset? I think it'd be useful to reduce the number of fixtures, although I don't think that's too important anyways. But we could use urls such as `https://placekitten.com/g/300/300` for this type of thing.

NoelDeMartin (changes requested):

src/core/services/sites.ts
dpalou: This morning I started working on the next issue, MOBILE-4460, and this property will be removed in that issue xD Now there are more conditions to determine if a URL should be displayed, it's not just if the site is demo, so I decided to remove this property and always use a method instead.

So if you don't mind we can leave this name as it is since it will be removed soon.

src/core/components/site-list/site-list.ts
dpalou: I renamed all the files and CSS classes, I think I didn't miss any.

src/assets/storybook/sites/schoolbarbara.ts
dpalou: Ok, I used https://pravatar.cc/ because it supports returning different images by adding a parameter to the URL.

NoelDeMartin (approved):

PR #3846, opened 2023-11-09
MOBILE-4464 siteplugins: Add class to identify site plugins

alfonso-salces:


src/core/features/siteplugins/components/module-index/module-index.ts
```diff
@@ -54,7 +54,7 @@ export class CoreSitePluginsModuleIndexComponent implements OnInit, OnDestroy, C
 
     @ViewChild(CoreSitePluginsPluginContentComponent) content?: CoreSitePluginsPluginContentComponent;
 
-    component?: string;
+    @HostBinding('class') component?: string;
```
dpalou, line 57: The change looks good to me, but for consistency I think we should apply it to `CoreSitePluginsBlockComponent` and `CoreSitePluginsCourseFormatComponent` too, for consistency.

dpalou (changes requested):

dpalou (approved):

PR #3849, opened 2023-11-10
MOBILE-4451 settings: Create error log page

alfonso-salces:


src/core/services/ws.ts
```diff
@@ -120,7 +121,7 @@ export class CoreWSProvider {
 
             return this.addToRetryQueue<T>(method, siteUrl, dataToSend, preSets);
         } else {
-            return this.performPost<T>(method, siteUrl, dataToSend, preSets);
+            return this.performPost<T>(method + 'jkshdkajs', siteUrl, dataToSend, preSets);
```
crazyserver, line 124: I think this change was made for testing and you should remove it.

src/core/features/settings/lang.json
```diff
@@ -60,6 +60,7 @@
     "navigatorlanguage": "Navigator language",
     "navigatoruseragent": "Navigator userAgent",
     "networkstatus": "Internet connection status",
+    "nologsavailable": "No logs available",
```
crazyserver, line 63: I would like to discuss if this is needed to be translated. Now we're only translation 1 over 6 translatable strings. I think you should delete the current translatable string.

crazyserver, line 63: I would like to discuss if this is needed to be translated. Now we're only translation 1 over 6 translatable strings. I think you should delete the current translatable string.

src/core/services/ws.ts
crazyserver, line 124: I think this change was made for testing and you should remove it.

crazyserver (changes requested):

crazyserver (approved):

PR #3851, opened 2023-11-13
MOBILE-4463: Implement cache invalidation

NoelDeMartin:


dpalou (changes requested):

upgrade.txt
```diff
@@ -1,13 +1,17 @@
 This files describes API changes in the Moodle Mobile app,
 information provided here is intended especially for developers.
 
+=== 4.4.0 ===
+
+- CoreCache has been deprecated, use plain object as in-memory stores instead.
```
dpalou, line 6:
```suggestion
- CoreCache has been deprecated for site plugins, plugins should use plain object as in-memory stores instead. CoreCache can still be used in the app's code.
```
IMO it's important to specify that it has been deprecated for site plugins, it hasn't been deprecated in the app itself.

dpalou (changes requested):

dpalou (approved):

PR #3852, opened 2023-11-13
Mobile 3947 Pre ionic 7 fixes

crazyserver:


src/addons/badges/pages/user-badges/user-badges.html
```diff
@@ -19,7 +19,7 @@ <h1>{{ 'addon.badges.badges' | translate }}</h1>
 
             <ion-list *ngIf="!badges.empty" class="ion-no-margin">
                 <ion-item button class="ion-text-wrap" *ngFor="let badge of badges.items" [attr.aria-label]="badge.name"
-                    (click)="badges.select(badge)" [attr.aria-current]="badges.getItemAriaCurrent(badge)" detail="true">
+                    (click)="badges.select(badge)" [attr.aria-current]="badges.getItemAriaCurrent(badge)" [detail]="true">
```
dpalou, line 22: About these changes, have you checked what happens if we compile a template from a site plugin and it doesn't comply this? I guess this check only affects our code, not plugins, but it would be nice to confirm that if a plugin uses `detail="true"` the compilation works fine too.

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -72,6 +74,16 @@ export class CoreUserAvatarComponent implements OnInit, OnChanges, OnDestroy {
      * @inheritdoc
      */
     ngOnInit(): void {
+        if (this.site && !this.user) {
+            this.user = {
+                id: 0,
```
dpalou, line 79: If you have `CoreSiteInfo`, you can use this.site.userid. In CoreSiteBasicInfo we don't have it, but we could calculate it easily in `getSites` and add it to the type.

src/addons/badges/pages/user-badges/user-badges.html
crazyserver: This is a change imposed by strictTemplates on tsconfig. So I think it would not affect external code. I'll check that anyway

src/core/directives/format-text.ts
```diff
@@ -584,7 +586,7 @@ export class CoreFormatTextDirective implements OnChanges, OnDestroy, AsyncDirec
         });
 
         // Handle all kind of frames.
-        const frameControllers = frames.map((frame: FrameElement) => {
+        const frameControllers = frames.map<FrameElementController>((frame) => {
```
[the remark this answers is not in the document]
crazyserver: The return is a new FrameElementController and type is for the returned value

src/core/components/user-avatar/user-avatar.ts
crazyserver: Done!

src/addons/mod/workshop/components/index/addon-mod-workshop-index.html
```diff
@@ -22,13 +22,13 @@ <h2 class="ion-text-wrap">{{ phases[workshop.phase].title }}</h2>
...
                     [attr.aria-label]="'addon.mod_workshop.tasktodo' | translate"></ion-icon>
-                <ion-icon slot="start" name="fas-circle-xmark" color="danger" *ngIf="task.completed == ''"
+                <ion-icon slot="start" name="fas-circle-xmark" color="danger" *ngIf="task.completed === ''"
                     [attr.aria-label]="'addon.mod_workshop.taskfail' | translate">
                 </ion-icon>
-                <ion-icon slot="start" name="fas-circle-info" color="info" *ngIf="task.completed == 'info'"
+                <ion-icon slot="start" name="fas-circle-info" color="info" *ngIf="task.completed === 'info'"
                     [attr.aria-label]="'addon.mod_workshop.taskinfo' | translate">
                 </ion-icon>
                 <ion-icon slot="start" name="fas-circle-check" color="success" *ngIf="task.completed == '1'"
```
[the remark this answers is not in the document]
crazyserver: I won't change that just in case it adds a bug, same with the other === '1'

dpalou (approved):

PR #3855, opened 2023-11-17
Mobile 3947: Pre Ionic 7 changes

crazyserver:


crazyserver:
Hi @dpalou !

I've added here last 6 commits. But rebasing I've got conflicts on commits:

MOBILE-3947 chore: Fix some HTML strict types
MOBILE-3947 user: Fix user types on user-avatar

src/core/features/settings/pages/dev/dev.ts
```diff
@@ -196,6 +197,16 @@ export class CoreSettingsDevPage implements OnInit {
         await CoreDomUtils.showToast('Caches invalidated', true, ToastDuration.LONG);
     }
 
+    /**
+     * Delete all data from the app.
+     */
+    async clearFileStorage(): Promise<void> {
+        await CoreFile.clearSitesFolder();
```
dpalou, line 204: I'm not a fan of this change, it will leave the app in an inconsistent state. Our filepool service will still say that files are downloaded, but then the files won't be found in the disk, and the app doesn't check this.

One possible solution would be to use something similar to `CoreSettingsHelperProvider.deleteSiteStorage`, but that's what space usage already does.

AFAIK you added this option to allow deleting the files of sites that no longer exist in the app (e.g. because you clear the browser cache in webapp). IMO it would be better to just check all the folders inside the `sites` folder and, and if any of the sites no longer exists in the DB, then we delete the folder. So instead of deleting all files we only delete the files of sites that no longer exist in the app. If we do this then we should change the text of this feature.

Maybe this is something that we should do periodically instead of having a manual option here. But I'm not sure if this situation is possible in a real device so we might be adding a cron task that does nothing in a real device.

src/core/features/settings/pages/general/general.ts
```diff
@@ -64,10 +64,13 @@ export class CoreSettingsGeneralPage {
         // Get the supported languages.
         const languages = CoreConstants.CONFIG.languages;
         for (const code in languages) {
-            this.languages.push({
-                code: code,
-                name: languages[code],
-            });
+            const exists = await CoreLang.checkLanguageExists(code);
```
dpalou, line 67: I don't like this change much, it's adding a lot of processing (we currently have more than 50 languages in the app). This is something that won't change in runtime, IMO it would be better to check this when the app is compiled. When compiling we can check the languages in the fiilesystem and remove the missing ones from config.json, that way this check doesn't need to be done at runtime.

src/core/services/lang.ts
```diff
@@ -173,10 +173,14 @@ export class CoreLangProvider {
...
-                CoreConfig.set('current_language', language),
-            ]);
+            await this.reloadLanguageStrings();
+            await CoreConfig.set('current_language', language);
+        } catch {
+            const defaultLang = this.getDefaultLanguage();
+            if (language !== defaultLang) {
+                return this.changeCurrentLanguage(defaultLang);
+            }
+            // Ignore errors.
```
dpalou, line 183: If we remove missing langs when compiling as I said in my previous comment I guess this code won't be needed either, right?

If you want to keep it, then maybe we should keep to the previous language instead of using the default one. And notify the user that something has failed.

dpalou (changes requested):

crazyserver:
Thanks Dani! I've removed the lang commit and applied the changes to the clear storage

dpalou (approved):

PR #3858, opened 2023-11-21
Mobile 4469

dpalou:


src/core/classes/sites/candidate-site.ts
```diff
@@ -12,84 +12,43 @@
...
-/**
- * Class that represents a site (combination of site + user).
- * It will have all the site data and provide utility functions regarding a site.
- * To add tables to the site's database, please use registerSiteSchema exported in @services/sites.ts. This will make sure that
- * the tables are created in all the sites, not just the current one.
- *
- * @todo Refactor this class to improve "temporary" sites support (not fully authenticated).
- */
-export class CoreSite {
+export class CoreCandidateSite extends CoreUnauthenticatedSite {
```
NoelDeMartin, line 51: To be consistent with the naming, why not call this `CoreAuthenticatedSite`? That way we'd have `CoreUnauthenticatedSite` > `CoreAuthenticatedSite` > `CoreSite`.

I don't like "unauthenticated" too much either, instead of using negatives I'd rather using something descriptive like `CoreGuestSite` or `CoreAnonymousSite`. But if we're going to have a subclass called `CoreAuthenticatedSite`, I think it's a worthy trade-off to have a consisent naming in the hierarchy.

It's also not clear what `CoreSite` has that `CoreAuthenticated` doesn't, but I guess it's fine if the final site we're going to use has the more concise name. The documentation in the classes seems clear enough, so it's fine to leave it like that.

Anyways, these are just ramblings about naming, feel free to ignore if you don't like the suggestion.

src/core/classes/sites/site.ts
```diff
@@ -0,0 +1,957 @@
...
+import { firstValueFrom } from '../../utils/rxjs';
+import { CoreFilepool } from '@services/filepool';
+import { CoreSiteInfo } from './unauthenticated-site';
+import { CoreCandidateSite, CoreCandidateSiteOptionalData, CoreSiteWSPreSets, WSObservable } from './candidate-site';
+
+/**
+ * Class that represents a site (combination of site + user).
+ * It will have all the site data and provide utility functions regarding a site.
+ * To add tables to the site's database, please use registerSiteSchema exported in @services/sites.ts. This will make sure that
+ * the tables are created in all the sites, not just the current one.¡
```
NoelDeMartin, line 60:
```suggestion
 * the tables are created in all the sites, not just the current one.
```

src/core/classes/sites/site.ts
```diff
@@ -0,0 +1,957 @@
...
+import { map } from 'rxjs/operators';
+import { firstValueFrom } from '../../utils/rxjs';
+import { CoreFilepool } from '@services/filepool';
+import { CoreSiteInfo } from './unauthenticated-site';
+import { CoreCandidateSite, CoreCandidateSiteOptionalData, CoreSiteWSPreSets, WSObservable } from './candidate-site';
+
+/**
+ * Class that represents a site (combination of site + user).
+ * It will have all the site data and provide utility functions regarding a site.
+ * To add tables to the site's database, please use registerSiteSchema exported in @services/sites.ts. This will make sure that
```
NoelDeMartin, line 59: I realize this was here before, but looking at the code I'm not sure why we need to mention it here. In fact, it seems like the `registerSiteSchema` method is not used anywhere and we're using `CORE_SITE_SCHEMAS` instead. So maybe we could deprecate or remove that function altogether?

In any case, I don't think we should mention anything about databases here. Or if we do, why here and not in all the other classes doing similar things?

src/core/classes/sites/site.ts
```diff
@@ -0,0 +1,957 @@
...
+ * the tables are created in all the sites, not just the current one.¡
+ */
+export class CoreSite extends CoreCandidateSite {
+
+    id: string;
+    infos?: CoreSiteInfo;
```
NoelDeMartin, line 65:
```suggestion
```

This is already declared in the parent class.

src/core/classes/sites/candidate-site.ts
```diff
@@ -117,108 +76,50 @@ export class CoreSite {
...
         CoreConstants.CONFIG.cache_update_frequency_rarely || 43200000,
     ];
 
-    // Rest of variables.
+    // WS that we allow to call even if the site is logged out.
+    protected static readonly ALLOWED_LOGGEDOUT_WS = [
+        'core_user_remove_user_device',
+    ];
+
+    id?: string | undefined; // For candidate sites it will always be undefined.
```
NoelDeMartin, line 91: I'm not sure we need to mention this here, or even if we do maybe we should say something like "for subclasses to implement" or something. After all, because of polymorphism, subclasses are also "candidate sites". So it's not true that candidate sites will always have an undefined id.

Even like that, though, it's a bit weird to have this declared just for subclasses... And looking at the code, I think this is only used to trigger events. So instead, what do you think of not declaring an id at all, and implementing a protected method like `triggerSiteEvent` which calls the even without an id and subclasses can override it using the id.

If that isn't straightforward, I think it's ok with just updating the comment.

src/core/classes/sites/candidate-site.ts
```diff
@@ -1714,12 +1330,15 @@ export class CoreSite {
                 try {
                     const config = await this.requestPublicConfig();
 
-                    if (cachePreSets.saveToCache) {
-                        this.saveToCache(method, {}, config, cachePreSets);
-                    }
+                    this.saveToCache(method, {}, config, cachePreSets);
```
NoelDeMartin, line 1333: This was within an if checking for `cachePreSets.saveToCache`, why was it removed?

src/core/classes/sites/site.ts
```diff
@@ -0,0 +1,957 @@
...
+        } else {
+            return this.config;
+        }
+    }
+
+    /**
+     * @inheritdoc
+     */
+    protected getDisabledFeatures(): string | undefined {
+        return this.getStoredConfig('tool_mobile_disabledfeatures');
```
NoelDeMartin, line 681: It seems like the parent class is using `publicConfig`, so I wonder if we should look into that if `this.config` is undefined.

Also, I think I've already mentioned this before but it's confusing to understand the difference between `config` and `publicConfig`. Is `publicConfig` a subset of `config`, or are they completely different things? If it is indeed a subset, we should declare that relation with types. If they are not the same, the naming makes it very confusing.

src/core/features/login/services/login-helper.ts
```diff
@@ -56,6 +64,10 @@ export class CoreLoginHelperProvider {
     static readonly FAQ_QRCODE_INFO_DONE = 'qrcode_info_done';
     static readonly FAQ_URL_IMAGE_HTML = '<img src="assets/img/login/faq_url.png" role="presentation" alt="">';
     static readonly FAQ_QRCODE_IMAGE_HTML = '<img src="assets/img/login/faq_qrcode.png" role="presentation" alt="">';
+    static readonly EMAIL_SIGNUP_FEATURE_NAME = 'CoreLoginEmailSignup';
+    static readonly FORGOTTEN_PASSWORD_FEATURE_NAME = 'NoDelegate_ForgottenPassword';
+    static readonly IDENTITY_PROVIDERS_FEATURE_NAME = 'NoDelegate_IdentityProviders';
+    static readonly IDENTITY_PROVIDER_FEATURE_NAME_PREFIX = 'NoDelegate_IdentityProvider_';
```
NoelDeMartin, line 70: Public constants should be declared in a different file, as we documented [in the guidelines](https://moodledev.io/general/development/policies/codingstyle-moodleapp#using-constants). If it's too much work to refactor the existing ones, we should do it at least for new code.

src/core/features/login/services/login-helper.ts
```diff
@@ -397,14 +411,56 @@ export class CoreLoginHelperProvider {
      * Get the valid identity providers from a site config.
      *
      * @param siteConfig Site's public config.
-     * @param disabledFeatures List of disabled features already treated. If not provided it will be calculated.
      * @returns Valid identity providers.
+     * @deprecated since 4.4. Please use getValidIdentityProvidersForSite instead.
      */
-    getValidIdentityProviders(siteConfig?: CoreSitePublicConfigResponse, disabledFeatures?: string): CoreSiteIdentityProvider[] {
+    getValidIdentityProviders(siteConfig?: CoreSitePublicConfigResponse): CoreSiteIdentityProvider[] {
```
NoelDeMartin, line 417: It seems like the implementation of this function has a lot of overlap with `getValidIdentityProvidersForSite`. I understand that it's deprecated, so eventually it should disappear, but would it be possible to rewrite it such that the code is not duplicated?

I think it's ok to continue calling the deprecated function within the new function, that way when we actually remove this function, we just move the code to the new function. But until that happens, the code is not duplicated.

src/core/classes/sites/candidate-site.ts
dpalou: Renamed to CoreAuthenticatedSite. I did it in a new commit because squashing the changes gave a bunch of conflicts.

src/core/classes/sites/site.ts
dpalou: I removed the comment because, as you said, this probably wasn't the best place to have it.

The function `registerSiteSchema` is for site plugins, the JSDoc of the function says "This function is meant for site plugins to create DB tables in current site". So we cannot deprecate it.

src/core/classes/sites/candidate-site.ts
dpalou: I created the triggerSiteEvent function in a new commit.

src/core/classes/sites/candidate-site.ts
dpalou: I think I removed this initially because part of the logic wasn't here and wasn't needed, but then I ended up changing the logic and I didn't add it back. Nice catch!

I added it back, and I also removed the setPublicConfig call in here because it's already done in the next .then.

src/core/classes/sites/site.ts
dpalou: I changed it to check `config` as you suggested.

Unfortunately, `publicConfig` and `config` have different data and they even have different signatures. Public config is retrieved using the WS `tool_mobile_get_public_config`, which returns some defined data (wwwroot, sitename, etc.). Config is retrieved using the WS `tool_mobile_get_config` and it returns an array with name and value for each config, so we don't know the exact configs being returned.

Public config has some data that config doesn't (e.g. the logo URL) and config has data that public config doesn't (like a lot of settings for the site). I agree it's confusing, but to fix this we would need to change the WebServices (or maybe even deprecate at least one of them and create a new one, because I'm not sure we can fix this without breaking the existing API).

src/core/features/login/services/login-helper.ts
dpalou: Right, I added a new commit to move the constants.

src/core/features/login/services/login-helper.ts
dpalou: I wanted to reuse the implementation, but there are 2 problems here:

1. The deprecated function is synchronous but the new one is async. This could be fixed by moving the common synchronous logic into a new function.
2. This would require using CoreSitesFactory in login-helper, and this creates a new circular dependency. IMO it's better to avoid that if it's just for a deprecated function.

NoelDeMartin (approved):

PR #3860, opened 2023-11-24
MOBILE-3947 qr-scanner: Create qr sscanner service

alfonso-salces:


src/core/features/native/native.module.ts
```diff
@@ -31,7 +31,7 @@ import { Keyboard } from '@awesome-cordova-plugins/keyboard/ngx';
 import { LocalNotifications } from '@awesome-cordova-plugins/local-notifications/ngx';
 import { MediaCapture } from '@awesome-cordova-plugins/media-capture/ngx';
 import { Push } from '@moodlehq/ionic-native-push/ngx';
-import { QRScanner } from '@ionic-native/qr-scanner/ngx';
+import { QRScanner } from './plugins';
```
NoelDeMartin, line 34: The value we pass to the compiler should be the service class, no the singleton instance like you're doing. It's a bit confusing with the plugins because we've change the nomenclature we normally use, but you should be able to look at the Chooser plugin to see how it's done.

src/core/features/native/plugins/qrscanner.ts
```diff
@@ -0,0 +1,95 @@
...
+import { filter } from 'rxjs/operators';
+
+/**
+ * QR Scanner plugin wrapper
+ */
+@Injectable({ providedIn: 'root' })
+export class QRScannerService {
+
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    private window: any;
```
NoelDeMartin, line 25: I don't like using any :(. Could we instead add the types for the plugin to window? Similar to what we did for the chooser plugin.

NoelDeMartin (changes requested):

src/core/features/native/plugins/qrscanner.ts
```diff
@@ -0,0 +1,182 @@
...
+        });
+    }
+
+    /**
+     * Disable the device's light.
+     *
+     * @param camera Provide `0` for back camera, and `1` for front camera.
+     * @returns QR scanner status.
+     */
+    useCamera(camera: number): Promise<QRScannerStatus> {
```
NoelDeMartin, line 140: I'm not even sure if we're using this, but maybe it'd be better to declare this as an enum instead to restrict the value to be 0 or 1 (and at the same time, make it easier to understand without reading the JSDocs).

NoelDeMartin (changes requested):

PR #3861, opened 2023-11-24
Mobile 3947

crazyserver:


src/core/components/swipe-slides/swipe-slides.html
```diff
@@ -1,6 +1,7 @@
-<ion-slides *ngIf="loaded" (ionSlideWillChange)="slideWillChange()" (ionSlideDidChange)="slideDidChange()" [options]="options">
-    <ion-slide *ngFor="let item of items; index as index" [attr.aria-hidden]="!isActive(index)">
+<swiper-container #swiperRef *ngIf="loaded" (slidechangetransitionstart)="slideWillChange()" (slidechangetransitionend)="slideDidChange()"
```
dpalou, line 1: Adding it here because there is no "right" place to add this comment: IMO we should document in upgrade.txt that ion-slides is no longer supported and now you need to use swiper-container and swiper-slide.

dpalou (changes requested):

crazyserver:
Thanks Dani! Can you recheck again?

PR #3863, opened 2023-11-28
MOBILE-3947 native: Fix chooser plugin

NoelDeMartin:


dpalou:
Thanks Noel, the code looks good to me but I'm not a fan of the SitePluginsProvider part, IMO it makes the code harder to understand. If I understood it right, the only reason for having this code is to allow renaming a service for site plugins, right? Why don't we just name our service `Chooser` instead of `CoreNativeChooserService`? I know it doesn't follow our naming conventions, but we can have some justified exceptions like this one.

NoelDeMartin:
@dpalou Ok, I've undone all that and just named the service like we had it before.

dpalou (approved):

PR #3865, opened 2023-11-28
Mobile 3947

dpalou:


webpack.config.js
```diff
@@ -18,6 +18,7 @@ module.exports = config => {
     config.optimization.minimizer.push(
         new TerserPlugin({
             terserOptions: {
+                ngDevMode: true,
```
NoelDeMartin, line 21: Why is this necessary? As far as I know, this file is not even being used yet because the custom builder to customize webpack was removed during the upgrade. Or is this now being used out of the box with the new builders?

There is a card open in trello to look into webpack, so if it is in fact not being used we can look into this at that point (maybe add a comment with some info in case it's relevant).

NoelDeMartin (changes requested):

webpack.config.js
crazyserver: Nah, was just for testig purposes...

NoelDeMartin (approved):

PR #3868, opened 2023-11-29
MOBILE-3947: Fix build

NoelDeMartin:


src/core/features/emulator/services/zip.ts
```diff
@@ -50,10 +50,11 @@ export class ZipMock extends Zip {
      *
      * @param source Path to the source ZIP file.
      * @param destination Destination folder.
-     * @param onProgress Optional callback to be called on progress update
+     * @param onProgressFunction Optional callback to be called on progress update
      * @returns Promise that resolves with a number. 0 is success, -1 is error.
      */
-    async unzip(source: string, destination: string, onProgress?: (ev: {loaded: number; total: number}) => void): Promise<number> {
+    async unzip(source: string, destination: string, onProgressFunction?: Function): Promise<number> {
+        const onProgress = onProgressFunction as (ev: {loaded: number; total: number}) => void;
```
dpalou, line 57: This threw an error because the Zip API has changed in awesome-cordova-plugins, they added an onSuccess callback before the onProgress. So this change is wrong IMO.

I don't know why they added this new onSuccess callback, the function returns a Promise so the "success" should be when the promise is resolved, I don't know why we need a callback for that. And this change in the API is a breaking change for plugins, although I think it won't affect any of them.

Maybe we should remove @awesome-cordova-plugins/zip and implement it ourselves as we did with other plugins, keeping the old API. What do you think?

dpalou (changes requested):

src/core/features/emulator/services/zip.ts
NoelDeMartin: Ok, I looked into it and it seems like the API is quite straightforward so I removed the native plugin and implemented the wrapper from scratch.

dpalou (approved):

PR #3877, opened 2023-12-05
MOBILE-3947: Fix application startup and behat gestures

NoelDeMartin:


src/core/constants.ts
```diff
@@ -14,8 +14,15 @@
...
 import envJson from '@/assets/env.json';
 import { EnvironmentConfig } from '@/types/config';
+import { InjectionToken } from '@angular/core';
 import { CoreBrowser } from '@singletons/browser';
 
+/**
+ * Injection token used for dependencies marked as optional that will never
+ * be resolved by Angular injectors.
+ */
+export const NULL = new InjectionToken('null');
```
dpalou, line 24: Shouldn't we call this variable something like `INJECT_NULL` or `INJECTION_TOKEN_NULL` instead of just `NULL`? I feel like NULL is too generic.

dpalou (changes requested):

src/core/constants.ts
NoelDeMartin: That's how I did it initially, but then I realized that all the injection tokens from Angular don't follow that convention, they are just called whatever they provide: `ROUTES`, `APP_INITIALIZER`, `INJECTOR`, etc.

If you still don't like it, we can rename it to something like `NULL_INJECTION_TOKEN`, I wouldn't mind changing it.

dpalou: I prefer `NULL_INJECTION_TOKEN`. I understand what you say about Angular not following that convention, but those words cannot be confused with a primitive type, while `NULL` can. I wouldn't mind naming it `NULL` if it was declared in a file dedicated to injection, but since this constant is declared in core/constants I think it's better to name it `NULL_INJECTION_TOKEN`.

dpalou (changes requested):

dpalou (approved):

PR #3878, opened 2023-12-12
Mobile 3947

crazyserver:


src/addons/mod/assign/components/submission/submission.ts
```diff
@@ -1124,9 +1124,15 @@ export class AddonModAssignSubmissionComponent implements OnInit, OnDestroy, Can
             return [];
         }
 
+        // Receved submission statement should not be undefined. It would mean that the WS is not returning the value.
```
dpalou, line 1127: I looked at this and this code was added because in Moodle 3.1 the submissionstatement field wasn't returned, it was added to 3.2. So I guess we can remove the check to see if it's undefined, right?

src/addons/calendar/components/filter/filter.html
```diff
@@ -11,17 +11,17 @@
...
+            <ion-toggle [(ngModel)]="filter[type]" (ionChange)="onChange()">
+                {{ 'addon.calendar.' + type + 'events' | translate}}
+            </ion-toggle>
         </ion-item>
         <core-spacer *ngIf="filter.course || filter.category || filter.group" />
         <ng-container *ngIf="filter.course || filter.category || filter.group">
             <ion-radio-group [(ngModel)]="courseId" (ionChange)="onChange()">
                 <ion-item class="ion-text-wrap" *ngFor="let course of sortedCourses">
-                    <ion-label>
+                    <ion-radio [value]="course.id">
```
dpalou, line 22: I saw that in some ion-radio you put labelPlacement="start", but not in this one or in some others (e.g. in edit-event.html). According to the Ionic docs, the default value is "start". Is there any reason why it's explicit in some places but not in others?

src/core/components/show-password/show-password.scss
```diff
@@ -3,30 +3,29 @@
 :host {
     display: contents;
 
-    ion-button {
+    // Only applies to old way (surrounding).
```
dpalou, line 6: Maybe it would be better to say "deprecated" way instead of "old"? In the component's JSDoc you say deprecated.

dpalou (changes requested):

src/addons/calendar/components/filter/filter.html
crazyserver: Revised and removed. I think I did it becausa it was previously specified.

src/core/components/show-password/show-password.scss
crazyserver: OK!

src/addons/mod/assign/components/submission/submission.ts
crazyserver: I've added a new commit, is that what you meant?

crazyserver:
Thanks Dani! I've added a new commit about core-input-errors. Please review it. Tomorrow I'll check your answers and test the assignment part becore fixing up the commits.

src/addons/mod/assign/components/submission/submission.ts
dpalou: Sorry, yesterday I looked at the original commit where this field was added to the WebService and then this field was always returned, but I see that now there are some conditions when it could happen that the field is undefined (maybe it's not possible to have that combination, but just in case I guess it's better to assume it can happen). 

So ignore my comment, I think the code is ok.

src/core/components/input-errors/input-errors.ts
```diff
@@ -59,31 +57,61 @@ export class CoreInputErrorsComponent implements OnInit, OnChanges {
     constructor(
         element: ElementRef,
     ) {
-        this.element = element.nativeElement;
+        this.hostElement = element.nativeElement;
+        //  this.hostElement.setAttribute('slot', 'error');
```
dpalou, line 61: This line can be removed.

src/core/components/input-errors/input-errors.ts
```diff
@@ -59,31 +57,61 @@ export class CoreInputErrorsComponent implements OnInit, OnChanges {
...
+            time: this.errorMessages.time || 'core.login.invalidtime',
+            url: this.errorMessages.url || 'core.login.invalidurl',
+            // Set empty values by default, the default error messages will be built in the template when needed.
+            max: this.errorMessages.max || '',
+            min: this.errorMessages.min || '',
+        };
+
+        this.errorMessages.requiredTrue = this.errorMessages.required;
+
+        this.errorKeys = this.errorMessages ? Object.keys(this.errorMessages) : [];
```
dpalou, line 83: Now `errorMessages` will always be defined, right?

src/core/features/login/services/login-helper.ts
```diff
@@ -252,60 +252,6 @@ export class CoreLoginHelperProvider {
...
-     * @param emailMsg Code of the string for invalid email error.
-     * @param patternMsg Code of the string for pattern not match error.
-     * @param urlMsg Code of the string for invalid url error.
-     * @param minlengthMsg Code of the string for "too short" error.
-     * @param maxlengthMsg Code of the string for "too long" error.
-     * @param minMsg Code of the string for min value error.
-     * @param maxMsg Code of the string for max value error.
-     * @returns Object with the errors.
-     */
-    getErrorMessages(
```
dpalou, line 269: If we don't want to deprecate this, then we should document in upgrade.txt that this function has been removed because it could impact plugins (I don't think it will, so I'm fine with removing it as long as we document it).

dpalou (changes requested):

NoelDeMartin (changes requested):
I just noticed that some Behat tests are failing in this PR that weren't failing in the `ionic7` branch, for example the forum "Create new discussion" test. It fails when trying to set the value of a RTE field, so it's probably related to changes made in this PR.

package.json
```diff
@@ -75,7 +75,7 @@
     "@awesome-cordova-plugins/sqlite": "^6.3.0",
     "@awesome-cordova-plugins/status-bar": "^6.3.0",
     "@awesome-cordova-plugins/web-intent": "^6.3.0",
-    "@ionic/angular": "^7.0.0",
+    "@ionic/angular": "^7.6.0",
```
phyr0s, line 78: As recomendation avoid using ^  in all package.json dependencies.

For example in the past i had lot's of problems with this --> "@awesome-cordova-plugins because sometimes they add new methods / name change and suddenly your app stop working with no reason 🗡️ ☠️

phyr0s: Also version 7.6.1 fixes several issues with the datetime component, menu, input and textarea

dpalou: Hi phyr0s, we usually use "^" while we do development. Right before starting our pre-release manual testing, we remove the "^" and set the exact versions that were installed using "^". If we detect any problem during the testing we try to fix it or we revert to a previous version.

dpalou (approved):

package.json
dpalou: For the record, Pau changed the Ionic version to 7.6.1 too.

PR #3881, opened 2023-12-14
MOBILE-4481 ws.ts: Get mimetype from file transfer headers

alfonso-salces:


src/core/services/ws.ts
```diff
@@ -271,8 +271,10 @@ export class CoreWSProvider {
 
                 // Google Drive extensions will be considered invalid since Moodle usually converts them.
                 if (!extension || ['gdoc', 'gsheet', 'gslides', 'gdraw', 'php'].includes(extension)) {
+                    const requestContentType = fileDownloaded.headers?.['Content-Type']?.match(/\[(.*?)\]/)?.[1];
```
dpalou, line 274: IMO this line should be after the comment "Not valid, get the file's mimetype."

src/core/services/ws.ts
```diff
@@ -258,7 +258,7 @@ export class CoreWSProvider {
             onProgress && transfer.onProgress(onProgress);
 
             // Download the file in the tmp file.
-            await transfer.download(url, fileEntry.toURL(), true, {
+            const fileDownloaded = await transfer.download(url, fileEntry.toURL(), true, {
```
dpalou, line 261: I noticed that this `fileDownloaded` is typed as `any` in ionic-native. I also noticed that the types modified in your PR in the file-transfer fork are wrong, here:

https://github.com/moodlemobile/cordova-plugin-file-transfer/commit/a8ee06df0fc368f2d7a2af261f872e86c190f34b#diff-093ad82a25aee498b11febf1cdcb6546e4d223ffcb49ed69cc275ac27ce0ccceR54

It should be `headers` instead of `header`, and it should be typed as optional because in iOS the headers won't be returned.

Taking into account that the Cordova plugin defines types and that ionic-native is wrongly typed (IMO the result shouldn't be type `any`), I propose that we stop using ionic-native for file-transfer as we did with other Cordova plugins.

FileTransfer is only used in CoreWS, and I don't think any plugin will use it directly because files should be downloaded through CoreFilepool, so in this case we might just use the Cordova plugin directly (using window.FileTransfer), without creating a wrapper. Site plugins will lose access to the FileTransfer service because it will no longer exist.

Let's discuss this internally with the rest of the team.

dpalou (changes requested):

package.json
```diff
@@ -79,7 +79,7 @@
     "@ionic/cordova-builders": "^10.0.0",
     "@moodlehq/cordova-plugin-advanced-http": "3.3.1-moodle.1",
     "@moodlehq/cordova-plugin-camera": "6.0.0-moodle.2",
-    "@moodlehq/cordova-plugin-file-transfer": "1.7.1-moodle.5",
+    "@moodlehq/cordova-plugin-file-transfer": "^1.7.1-moodle.8",
```
dpalou, line 82: We don't use "^" in our plugins because we control their releases. Sometimes we introduce breaking changes in minor releases and we don't want our app to install those minors by mistake.

src/core/services/ws.ts
```diff
@@ -254,14 +254,21 @@ export class CoreWSProvider {
             // Create the tmp file as an empty file.
             const fileEntry = await CoreFile.createFile(tmpPath);
 
-            const transfer = FileTransfer.create();
-            onProgress && transfer.onProgress(onProgress);
+            const transfer = new window.FileTransfer();
+
+            onProgress && transfer.onprogress(onProgress[1]);
```
dpalou, line 259:
```suggestion
            if (onProgress) {
                transfer.onprogress = onProgress;
            }
```
This line is wrong, using onProgress[1] in a callback will return undefined IMO. The reason TS was complaining is because onprogress needs to be assigned like this.

src/core/services/ws.ts
```diff
@@ -254,14 +254,21 @@ export class CoreWSProvider {
...
-                    'User-Agent': navigator.userAgent,
-                },
+            const fileDownloaded = await new Promise<{
+                entry: globalThis.FileEntry;
+                headers: Record<string, string> | undefined;
+            }>((resolve, reject) => {
+                transfer.download(
+                    url,
+                    fileEntry.toURL(),
+                    (params) => resolve(params),
```
dpalou, line 269: Can we rename this "params" to "result" or something like this? The name confused me a bit because I thought they were "input" parameters.

src/core/services/ws.ts
```diff
@@ -254,14 +254,21 @@ export class CoreWSProvider {
...
             // Download the file in the tmp file.
-            await transfer.download(url, fileEntry.toURL(), true, {
-                headers: {
-                    'User-Agent': navigator.userAgent,
-                },
+            const fileDownloaded = await new Promise<{
+                entry: globalThis.FileEntry;
+                headers: Record<string, string> | undefined;
+            }>((resolve, reject) => {
+                transfer.download(
```
dpalou, line 266: With your new code you lost 2 parameters: trustAllHosts and options (the headers we send). You need to pass them after the error callback.

src/core/services/ws.ts
```diff
@@ -271,8 +278,10 @@ export class CoreWSProvider {
 
                 // Google Drive extensions will be considered invalid since Moodle usually converts them.
                 if (!extension || ['gdoc', 'gsheet', 'gslides', 'gdraw', 'php'].includes(extension)) {
+
                     // Not valid, get the file's mimetype.
-                    const mimetype = await this.getRemoteFileMimeType(url);
+                    const requestContentType = fileDownloaded.headers?.['Content-Type']?.match(/\[(.*?)\]/)?.[1];
```
dpalou, line 283: When getting the mimetype using a HEAD request we found that in some cases the mimetype could contain "parameters" like charset (see [this code](https://github.com/moodlehq/moodleapp/blob/main/src/core/services/ws.ts#L343)). I guess it can happen here too, so I would add the treatment too just in case

dpalou, line 283: In iOS, the headers are returned without "[]" in their values so this regex won't work properly. IMO it would be better if the Android's native code also removed those characters so the values returned are the same in both platforms. Example of what is returned in iOS:

Accept-Ranges: "bytes"
Cache-Control: "private, max-age=3600, no-transform"
Connection: "keep-alive"
Content-Disposition: "inline; filename=\"typesofsport.jpg\""
Content-Length: "96027"
Content-Type: "image/jpeg"
Date: "Tue, 23 Jan 2024 07:40:58 GMT"
Etag: "\"3b50e1ce6b7326e3cfdb847478215a49f6153c8b\""
Expires: "Tue, 23 Jan 2024 08:40:58 GMT"
Last-Modified: "Tue, 19 Jun 2018 11:37:21 GMT"
Pragma: ""
Server: "nginx"

src/core/services/ws.ts
```diff
@@ -254,14 +254,21 @@ export class CoreWSProvider {
             // Create the tmp file as an empty file.
             const fileEntry = await CoreFile.createFile(tmpPath);
 
-            const transfer = FileTransfer.create();
-            onProgress && transfer.onProgress(onProgress);
+            const transfer = new window.FileTransfer();
```
dpalou, line 257: IMO we should stop using `@awesome-cordova-plugins/file-transfer` everywhere, not just in this function. I think it's only used in the WS provider so it shouldn't be a lot of work

Also, right now in browser we override the `FileTransfer` service to make this work in our webapp. If we stop using `@awesome-cordova-plugins/file-transfer` then we need to change our browser code to implement the Cordova API instead of this service (basically, we need to implement `window.FileTransfer` instead of an Angular service). You can search for `FileTransferMock` to see our implementation.

dpalou (changes requested):

alfonso-salces:
Reminder: Update "@moodlehq/cordova-plugin-file-transfer" dependency to the last version before merge.

src/core/features/emulator/services/file-transfer.ts
```diff
@@ -101,7 +90,6 @@ export class FileTransferObjectMock extends FileTransferObject {
             this.xhr = xhr;
             this.source = source;
             this.target = target;
-            this.reject = reject;
```
dpalou, line 104: This was done to be able to reject the promise when calling `abort`. Now you should probably do the same but using `errorCallback` instead. This means you can replace the `reject` property with an `errorCallback` or similar, and in `abort` you should call the `errorCallback` instead.

src/core/features/emulator/services/file-transfer.ts
```diff
@@ -87,12 +67,21 @@ export class FileTransferObjectMock extends FileTransferObject {
...
-    download(source: string, target: string, trustAllHosts?: boolean, options?: { [s: string]: any }): Promise<unknown> {
+    download(
+        source: string,
+        target: string,
+        successCallback: (params: CoreFileTransferDownloadResponse) => void,
+        errorCallback: (error: FileTransferError) => void,
+        trustAllHosts?: boolean,
+        // eslint-disable-next-line @typescript-eslint/no-explicit-any
+        options?: Record<string, any>,
+    ): Promise<unknown> {
```
dpalou, line 84: The original function from the Cordova plugin doesn't return a Promise, it returns void because it uses the callbacks to return the data. So IMO it would be better to be consistent with the original API and remove all the Promises usage in here, just use the callbacks.

src/core/features/emulator/services/file-transfer.ts
```diff
@@ -141,24 +131,33 @@ export class FileTransferObjectMock extends FileTransferObject {
...
                 }
 
                 const basePath = CoreFile.getBasePathInstant();
                 target = target.replace(basePath, ''); // Remove basePath from the target.
                 target = target.replace(/%20/g, ' '); // Replace all %20 with spaces.
-
-                // eslint-disable-next-line promise/catch-or-return
-                CoreFile.writeFile(target, response).then(resolve, reject);
+                CoreFile.writeFile(target, response)
+                    .then((entry: unknown) => {
```
dpalou, line 153: Why are you using `unknown` here? Wouldn't it be better to use the inferred FileEntry type instead? If it causes conflicts with the type defined in `CoreFileTransferDownloadResponse`, you can try importing:

`import { FileEntry } from '@awesome-cordova-plugins/file/ngx';`

and use this type in `CoreFileTransferDownloadResponse`.

src/core/features/emulator/services/file-transfer.ts
```diff
@@ -230,10 +229,10 @@ export class FileTransferObjectMock extends FileTransferObject {
     /**
      * Registers a listener that gets called whenever a new chunk of data is transferred.
      *
-     * @param listener Listener that takes a progress event.
+     * @param event Listener that takes a progress event.
      */
-    onProgress(listener: (event: ProgressEvent) => void): void {
-        this.progressListener = listener;
+    onprogress(event: ProgressEvent<EventTarget>): void {
```
dpalou, line 234: `onprogress` now needs to be a property, not a function, since it's used like this:

`transfer.onprogress = onProgress;`

src/core/features/emulator/services/file-transfer.ts
```diff
@@ -285,10 +284,18 @@ export class FileTransferObjectMock extends FileTransferObject {
...
      * @returns Promise that resolves to a FileUploadResult and rejects with FileTransferError.
      */
-    upload(fileUrl: string, url: string, options?: FileUploadOptions): Promise<FileUploadResult> {
+    upload(
+        fileUrl: string,
+        url: string,
+        successCallback: (result: FileUploadResult) => void,
+        errorCallback: (error: FileTransferError) => void,
+        options?: FileUploadOptions,
+    ): Promise<FileUploadResult> {
```
dpalou, line 298: As with download, I think it's better to remove all Promises from here and just use the callbacks.

src/core/features/native/native.module.ts
```diff
@@ -73,7 +71,6 @@ export const CORE_NATIVE_SERVICES = [
         Diagnostic,
         File,
         FileOpener,
-        FileTransfer,
```
dpalou, line 76: Please add a note in upgrade.txt mentioning that the `FileTransfer` service is no longer available.

src/core/services/ws.ts
```diff
@@ -254,14 +253,25 @@ export class CoreWSProvider {
...
+                transfer.onprogress = onProgress;
+            }
 
             // Download the file in the tmp file.
-            await transfer.download(url, fileEntry.toURL(), true, {
-                headers: {
-                    'User-Agent': navigator.userAgent,
-                },
+            const fileDownloaded = await new Promise<{
+                entry: globalThis.FileEntry;
```
dpalou, line 264: If you try to use the type `FileEntry` from `@awesome-cordova-plugins/file/ngx`, do you get a conflict? In general in the app we use that type, so it would be great if we could use the same type everywhere. I'm not sure if they're compatible or not.

dpalou (changes requested):

src/core/features/emulator/services/file-transfer.ts
```diff
@@ -43,40 +40,24 @@ export class FileTransferErrorMock implements FileTransferError {
...
-
-}
-
-/**
- * Emulates the FileTransferObject class in desktop apps and in browser.
- */
-export class FileTransferObjectMock extends FileTransferObject {
+export class FileTransferMock {
 
     progressListener?: (event: ProgressEvent) => void;
```
dpalou, line 45: This property is no longer used and should be removed (it's never assigned), now you need to use `onprogress` instead.

src/core/features/emulator/services/file-transfer.ts
```diff
@@ -43,40 +40,24 @@ export class FileTransferErrorMock implements FileTransferError {
...
     source?: string;
     target?: string;
     xhr?: XMLHttpRequest;
 
-    protected reject?: (reason?: unknown) => void;
+    onprogress!: (event: ProgressEvent) => void;
```
dpalou, line 50:
```suggestion
    // eslint-disable-next-line @typescript-eslint/no-empty-function
    onprogress: (event: ProgressEvent) => void = () => {};
```

Instead of using "!" I think it's better to initialize this with a function that does nothing. This way you can always call onprogress in the implementation without having to check if it exists or not. E.g. instead of doing this:

```
if (this.progressListener) {
    this.progressListener(ev);
}
```

Now you will be able to just do:

`this.onprogress(ev);`

dpalou (changes requested):

dpalou (approved):

PR #3887, opened 2024-01-09
Mobile 3947

crazyserver:


.github/workflows/acceptance.yml
```diff
@@ -37,7 +37,7 @@ jobs:
     - name: Additional checkouts
       run: |
         git clone --branch $MOODLE_BRANCH --depth 1 $MOODLE_REPOSITORY $GITHUB_WORKSPACE/moodle
-        git clone --branch master --depth 1 https://github.com/moodlehq/moodle-docker $GITHUB_WORKSPACE/moodle-docker
+        git clone --branch main --depth 1 https://github.com/moodlehq/moodle-docker $GITHUB_WORKSPACE/moodle-docker
```
NoelDeMartin, line 40: Since you're doing this here, I think we should also update it in `performance.yml`.

Honestly though, we haven't used that in ages so I'd be fine with removing the entire workflow file and performance tests as well xD. But if we decide to keep the files, they should be up to date.

src/addons/calendar/tests/behat/create_events.feature
```diff
@@ -20,13 +20,13 @@ Feature: Test creation of calendar events in app
       | teacher1 | C1 | editingteacher |
       | student1 | C1 | student |
 
-  @ionic7_failure
+  @flakytest_by_timestamp
```
NoelDeMartin, line 23: What's the use of this tag? I think what it marks is self-explanatory, but if it's only to give information maybe a comment would do as well... Or are we using this tag in some CI? We have the `@ci_jenkins_skip` tag to skip tests in jenkins for example, maybe just adding that one with comments would do? Also, is there any way to make it consistent rather than flaky? Or would that be too much work?

src/core/components/swipe-slides/swipe-slides.ts
```diff
@@ -46,22 +48,20 @@ export class CoreSwipeSlidesComponent<Item = unknown> implements OnChanges, OnDe
          * This setTimeout waits for Ionic's async initialization to complete.
          * Otherwise, an outdated swiper reference will be used.
          */
-        setTimeout(() => {
+        setTimeout(async () => {
             if (swiperRef?.nativeElement?.swiper) {
                 this.swiper = swiperRef.nativeElement.swiper as Swiper;
 
+                await this.initialize();
+                this.updateOptions();
```
NoelDeMartin, line 56: It seems like `updateOptions` is always called after `initialize`, so I think we could just move it inside. Also, the `onReadyPromise` seems to be resolved before the options are updated, and that could cause some race conditions. So I think that'd also be another reason to move `updateOptions()` inside `initialize`.

src/core/components/context-menu/core-context-menu-popover.html
```diff
@@ -7,17 +7,22 @@
             [href]="item.href" (click)="itemClicked($event, item)" [attr.aria-label]="item.ariaAction" [hidden]="item.hidden"
             [detail]="!!(item.href && !item.iconAction)" role="menuitem" [button]="!!(item.href && !item.iconAction)"
             [showBrowserWarning]="item.showBrowserWarning">
-            <ion-label>
+            <ion-toggle *ngIf="item.iconAction === 'toggle'" [(ngModel)]="item.toggle" (ionChange)="item.toggleChanged($event)">
```
NoelDeMartin, line 10: It seems like fixing ion-toggle required some changes in the markup, besides updating the Behat tests. Is that something that only affects the tests, or was it necessary for ion-toggle to work properly? If it's the latter, I guess we'll have to review all the ion-toggle in the app (maybe create a card in Trello if you haven't already done so).

src/testing/services/behat-dom.ts
```diff
@@ -417,7 +448,14 @@ export class TestingBehatDomUtilsService {
...
 
-            return (inputId && document.getElementById(inputId)) || undefined;
+            if (inputId) {
+                return document.getElementById(inputId) || undefined;
+            }
+
+            input = this.getShadowDOMHost(label) || undefined;
+            if (input?.tagName.startsWith('ION-')) {
+                return input;
+            }
```
NoelDeMartin, line 458: Looking at this code, and the commit that introduced it, it seems like we're finding an input by a sibling label that is not using the `for` attribute. If that's the case, I guess that means it's using the `aria-labelledby` instead? If it is so, we should also make sure about that before selecting the input. Otherwise, I don't think this is correct and screen readers would have issues with this.

NoelDeMartin (changes requested):

src/core/components/context-menu/core-context-menu-popover.html
crazyserver: It's the latter, but luckily this was the only one left to be changed, so the other have already been updated.

src/testing/services/behat-dom.ts
crazyserver: This is due to the change on ion- "input" fields changed on the ionic update. I think it's correct for a11y but some changes may be done. I'll change the condition  to get only ion-select and add a comment to clarify that if other fields fail, tag name shoud be added here.

crazyserver:
Thanks Noel, can you review again?

crazyserver:
Hi Noel! As you can see I've added 3 additional commits

NoelDeMartin (approved):

PR #3902, opened 2024-01-19
MOBILE-4266 site: Add site theme class to html tags

crazyserver:


src/app/app.component.ts
```diff
@@ -36,10 +36,12 @@ import { CoreUrl } from '@singletons/url';
 import { CoreLogger } from '@singletons/logger';
 import { CorePromisedValue } from '@classes/promised-value';
 import { register } from 'swiper/element/bundle';
+import { CoreSiteInfo, CoreSiteInfoResponse } from '@classes/sites/unauthenticated-site';
```
NoelDeMartin, line 39: I may have suggested this before, but I think we should keep the app component as lean as possible. Adding new imports to this file increases the chances to have circular dependencies, something that eventually we'd like to avoid (it's also better for code splitting, etc.).

Maybe this is an opportunity to move all this logic elsewhere? maybe a dedicated initializer?

src/app/app.component.test.ts
```diff
@@ -46,4 +50,66 @@ describe('AppComponent', () => {
...
+        expect(document.documentElement.classList.contains('theme-site-'+themeName2)).toBe(false);
+
+        if (site.infos) {
+            site.infos.theme = themeName2;
+        }
+
+        CoreEvents.trigger(CoreEvents.SITE_UPDATED, site.infos , '42');
+
+        // Wait the event to be processed.
+        await wait(50);
```
NoelDeMartin, line 96: Rather than waiting some fixed time, would it work waiting one tick instead? Or maybe more than one, but I think it's better waiting for ticks than time.

crazyserver:
Thanks Noel I think I've fixed most of your requests. I haven't moved all the code since it's quite difficult to decide where to put it.

NoelDeMartin (approved):

PR #3917, opened 2024-01-31
MOBILE-4243 course: Add communication room button to course page

crazyserver:


src/core/features/course/services/course-helper.ts
```diff
@@ -2031,6 +2031,29 @@ export class CoreCourseHelperProvider {
         );
     }
 
+    /**
+     * Get course communication room URL.
+     *
+     * @param course Course.
+     * @returns Promise resolved with the URL.
+     */
+    async getCourseCommunicationroom(course: CoreCourseAnyCourseData): Promise<string | undefined> {
```
dpalou, line 2040: This function doesn't follow camel case, IMO it should be `getCourseCommunicationRoom`.

src/core/features/course/services/course-helper.ts
```diff
@@ -2031,6 +2031,29 @@ export class CoreCourseHelperProvider {
...
+
+        const site = CoreSites.getRequiredCurrentSite();
+        if (!site.isVersionGreaterEqualThan('4.4')) {
```
dpalou, line 2043: Right now we don't have the 4.4 version specified in CoreAuthenticatedSite. Please add it with the expected release date, including a `todo` to review it once the version is released.

src/core/features/course/tests/behat/communication.feature
```diff
@@ -0,0 +1,30 @@
+@core @core_course @app @javascript @communication @lms_from4.4
```
dpalou, line 1: With Noel's latest changes the tags have changed, we no longer use "core", just "core_course". And I'm not sure if we want to include the "communication" tag, in general we no longer link the app's tests with LMS tests.

You can see the changes in this commit: https://github.com/moodlehq/moodleapp/commit/bb6ed65c88ca51fd40ae3b7911113faa05eb2b12

dpalou (changes requested):

dpalou (approved):

PR #3924, opened 2024-02-06
MOBILE-4501 cordova-plugin-moodleapp: Add diagnostic plugin

alfonso-salces:


cordova-plugin-moodleapp/plugin.xml
```diff
@@ -27,5 +49,68 @@
...
+        <source-file src="src/ios/Diagnostic_Location.m" />
+
+        <framework src="CoreLocation.framework" />
+
+        <config-file target="*-Info.plist" parent="NSLocationWhenInUseUsageDescription" comment="Default usage descriptions: override as necessary in .plist">
+            <string>This app requires access to your location when the screen is on and the app is displayed.</string>
+        </config-file>
+
+        <config-file target="*-Info.plist" parent="NSLocationAlwaysUsageDescription" comment="iOS 10">
+            <string>This app requires constant access to your location in order to track your position, even when the screen is off or the app is in the background.</string>
```
NoelDeMartin, line 89: This description is not correct for our app; we don't require constant position tracking. We can use this opportunity to update these descriptions to something that makes sense for our app.

cordova-plugin-moodleapp/src/android/Diagnostic.java
NoelDeMartin, line 1: I guess this file, and others, have been copied straight from the plugin source, so I'm not doing any reviews here.

Keeping that in mind, I think we should have a separate commit for the files we've copied from the plugin, and if possible doing the proper attribution (using [github's Co-authored-by feature](https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors)).

cordova-plugin-moodleapp/scripts/apply-diagnostic-modules.js
NoelDeMartin, line 1: I don't think this file is needed. The original plugin seems to use it to toggle the declarations within `plugin.xml`, but we aren't doing any of that.

cordova-plugin-moodleapp/src/ts/plugins/Diagnostic.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Checks whether device hardware features are enabled or available to the app, e.g. camera, GPS, wifi
+ */
+export class Diagnostic {
+
+    /**
+     * Constants for requesting and reporting the various permission states.
+     */
+    declare permissionStatus: {
```
NoelDeMartin, line 49: I'm surprised that this works at all. Using `declare` in TypeScript will not produce any JavaScript code, so this will be missing at runtime. What you should do instead is declare the class field here (using `declare`), but initialize it in the constructor (using `this.permissionStatus = ...` which will included in the JavaScript transpilation.

cordova-plugin-moodleapp/src/ts/plugins/Diagnostic.ts
```diff
@@ -0,0 +1,123 @@
...
+        // Android only
+        ['DENIED_ONCE']: 'DENIED_ONCE';
```
NoelDeMartin, line 51:
```suggestion
        DENIED_ONCE: 'DENIED_ONCE';
```

Same suggestion for all other values.

cordova-plugin-moodleapp/src/ts/plugins/Diagnostic.ts
```diff
@@ -0,0 +1,123 @@
...
+    getLocationAuthorizationStatus(): Promise<unknown> {
+        return new Promise<unknown>((resolve, reject) =>
+            cordova.exec(resolve, reject, 'Diagnostic', 'getLocationAuthorizationStatus'));
+    }
+
+    requestLocationAuthorization(): Promise<void> {
+        return new Promise<void>((resolve, reject) => cordova.exec(resolve, reject, 'Diagnostic', 'requestLocationAuthorization'));
+    }
+
+    async requestMicrophoneAuthorization(): Promise<string> {
```
NoelDeMartin, line 117: Why is this the only method declared using async syntax? I think it should be consistent, at least within the same file.

cordova-plugin-moodleapp/types/index.d.ts
```diff
@@ -29,3 +31,5 @@ declare global {
 }
 
 export type SecureStorage = InstanceType<typeof SecureStorageImpl>;
+
+export type Diagnostic = InstanceType<typeof DiagnosticImpl>;
```
NoelDeMartin, line 35: I don't think we need to export this. In the case of SecureStorage it was necessary because we're implementing a mock for browser environments, but that isn't necessary for Diagnostic.

NoelDeMartin (changes requested):

src/core/services/geolocation.ts
```diff
@@ -105,25 +112,32 @@ export class CoreGeolocationProvider {
...
                 await this.requestLocationAuthorization();
                 this.logger.log('Location authorization granted.');
                 await CoreApp.waitForResume(500);
                 await this.doAuthorizeLocation(true);
 
                 return;
-            case Diagnostic.permissionStatus.GRANTED:
-            case Diagnostic.permissionStatus.GRANTED_WHEN_IN_USE:
+
+            case diagnostic?.permissionStatus.granted:
```
NoelDeMartin, line 139:
```suggestion
            case diagnostic.permissionStatus.granted:
```

src/core/services/geolocation.ts
```diff
@@ -105,25 +112,32 @@ export class CoreGeolocationProvider {
...
+        if (!diagnostic) {
+            return;
+        }
+
+        const authorizationStatus = await diagnostic.getLocationAuthorizationStatus();
         this.logger.log(`Authorize location: status ${authorizationStatus}`);
 
         switch (authorizationStatus) {
-            case Diagnostic.permissionStatus.DENIED_ONCE:
+            case diagnostic.permission.acceptHandover:
```
NoelDeMartin, line 125:
```suggestion
            case diagnostic.permissionStatus.deniedOnce:
```

cordova-plugin-moodleapp/plugin.xml
```diff
@@ -27,5 +48,68 @@
...
+        <source-file src="src/ios/Diagnostic_Location.m" />
+
+        <framework src="CoreLocation.framework" />
+
+        <config-file target="*-Info.plist" parent="NSLocationWhenInUseUsageDescription" comment="Default usage descriptions: override as necessary in .plist">
+            <string>This app requires access to your location when request location in database activities.</string>
+        </config-file>
+
+        <config-file target="*-Info.plist" parent="NSLocationAlwaysUsageDescription" comment="iOS 10">
+            <string>This app only require access to your location when request location in database activities, we don't use your location anywhere else.</string>
```
NoelDeMartin, line 88: I don't think this is gramatically correct, maybe you can ask some of the UX Writters for help with these descriptions? We could also use the opportunity to update the one about the microphone, given that the access is not really needed for the app to function properly (it's just needed if you want to attach audios).

cordova-plugin-moodleapp/src/android/Diagnostic.java
NoelDeMartin, line 1: Maybe we can also add a link with a note to the original repository of the plugin? Rather than just adding the maintainer as a co-author, this will make it clear that we copied it from the repository, rather than we actually worked with them to move it here.

cordova-plugin-moodleapp/src/ts/plugins/Diagnostic.ts
```diff
@@ -0,0 +1,128 @@
...
+
+/**
+ * Checks whether device hardware features are enabled or available to the app, e.g. camera, GPS, wifi
+ */
+export class Diagnostic {
+
+    /**
+     * Constants for requesting and reporting the various permission states.
+     */
+    declare permissionStatus;
```
NoelDeMartin, line 49: Since this is declared without a type, how does that behave in the app? Is it inferred as any? Maybe we could extract the values as a constant and reuse it both inside of the constructor and for the declaration. Something like this:

```ts
const status = {
    // ...
} as const;

class MyClass {

    declare status: typeof status;
    
    constructor() {
        this.status = status;
    }

}
```

That's just an idea, though (and honestly I don't like it too much, but nothing else comes to mind). What I think is important is that these aren't declared as any.

NoelDeMartin (changes requested):

NoelDeMartin (approved):

PR #3933, opened 2024-02-19
MOBILE-4313 notifications: Show warnings if permissions missing

dpalou:


src/addons/calendar/pages/event/event.ts
```diff
@@ -61,6 +64,7 @@ export class AddonCalendarEventPage implements OnInit, OnDestroy {
     protected defaultTimeChangedObserver: CoreEventObserver;
     protected currentSiteId: string;
     protected updateCurrentTime?: number;
+    protected appResumeSubscription: Subscription;
```
crazyserver, line 67: Just a personal thing: Add subscriptions array and observers array to then turn them off altogether on destroy.

src/addons/notifications/pages/list/list.ts
```diff
@@ -47,12 +51,15 @@ export class AddonNotificationsListPage implements AfterViewInit, OnDestroy {
...
     loadingMarkAllNotificationsAsRead = false;
+    hasNotificationsPermission = true;
+    permissionWarningHidden = false;
 
     protected isCurrentView?: boolean;
     protected cronObserver?: CoreEventObserver;
     protected readObserver?: CoreEventObserver;
     protected pushObserver?: Subscription;
     protected pendingRefresh = false;
+    protected appResumeSubscription?: Subscription;
```
crazyserver, line 62: Just a personal thing: Add subscriptions array and observers array to then turn them off altogether on destroy.

crazyserver (changes requested):

src/addons/calendar/pages/event/event.ts
dpalou: I don't have a strong opinion about this, but having them like they are now makes it easier to identify which observers and subscriptions are in the page.

If we want to gather them together in an array maybe we should do it in the whole app. We can open a follow up issue for this if you think it's important.

crazyserver (approved):

PR #3934, opened 2024-02-20
Mobile 4499

dpalou:


src/addons/calendar/pages/event/event.ts
```diff
@@ -61,6 +64,7 @@ export class AddonCalendarEventPage implements OnInit, OnDestroy {
     protected defaultTimeChangedObserver: CoreEventObserver;
     protected currentSiteId: string;
     protected updateCurrentTime?: number;
+    protected appResumeSubscription: Subscription;
```
crazyserver, line 67: Just a personal thing: Add subscriptions array and observers array to then turn them off altogether on destroy.

src/addons/notifications/pages/list/list.ts
```diff
@@ -47,12 +51,15 @@ export class AddonNotificationsListPage implements AfterViewInit, OnDestroy {
...
     loadingMarkAllNotificationsAsRead = false;
+    hasNotificationsPermission = true;
+    permissionWarningHidden = false;
 
     protected isCurrentView?: boolean;
     protected cronObserver?: CoreEventObserver;
     protected readObserver?: CoreEventObserver;
     protected pushObserver?: Subscription;
     protected pendingRefresh = false;
+    protected appResumeSubscription?: Subscription;
```
crazyserver, line 62: Same (array thing)

src/theme/theme.base.scss
```diff
@@ -931,6 +931,25 @@ ion-card {
...
+
+    &.core-card-with-buttons .item ion-label {
+        margin-bottom: 0;
+    }
+
+    .core-card-buttons {
+        display: flex;
+        flex-direction: row;
+        justify-content: flex-end;
+        @include margin(0, 8px, 8px, 8px);
```
crazyserver, line 943: Since start and end (left and right) is the same, use margin: directly without the include.

crazyserver (changes requested):

src/theme/theme.base.scss
dpalou: Changed (the CSS was included in the previous issue though, but I fixed it here).

crazyserver (approved):

PR #3938, opened 2024-02-21
Mobile 4329

crazyserver:


src/core/features/siteplugins/classes/handlers/user-handler.ts
```diff
@@ -51,9 +51,9 @@ export class CoreSitePluginsUserProfileHandler extends CoreSitePluginsBaseHandle
 
         this.priority = handlerSchema.priority || 0;
 
-        // Only support TYPE_COMMUNICATION and TYPE_NEW_PAGE.
-        this.type = handlerSchema.type != CoreUserDelegateService.TYPE_COMMUNICATION ?
-            CoreUserDelegateService.TYPE_NEW_PAGE : CoreUserDelegateService.TYPE_COMMUNICATION;
+        // Only support LIST_ITEM and BUTTON.
+        this.type = handlerSchema.type !== CoreUserProfileHandlerType.BUTTON ?
+        CoreUserProfileHandlerType.LIST_ITEM : CoreUserProfileHandlerType.BUTTON;
```
dpalou, line 56: You should still check TYPE_COMMUNICATION here, otherwise plugins using it will start using LIST_ITEM instead of BUTTON.

dpalou (changes requested):

src/core/features/dataprivacy/services/dataprivacy.ts
```diff
@@ -0,0 +1,364 @@
...
+        return CoreDataPrivacyService.ROOT_CACHE_KEY + 'datarequests';
+    }
+
+    /**
+     * Fetch the details of a user's data request.
+     *
+     * @param siteId Site ID. If not defined, current site.
+     * @returns Promise resolved with the data requests.
+     * @since 4.4
+     * @todo Implement types and pagination.
```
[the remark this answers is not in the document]
crazyserver: No need :-) thanks!

src/theme/theme.base.scss
```diff
@@ -2015,3 +2021,83 @@ ion-item.item-label-stacked ion-datetime-button {
...
+        --cell-hover: var(--core-table-odd-cell-hover);
+    }
+
+    .even {
+        --cell-background: var(--core-table-even-cell-background);
+        --cell-hover: var(--core-table-even-cell-hover);
+    }
+
+    .odd, .even {
+        td, th, th[aria-current="page"] {
```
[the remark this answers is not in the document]
crazyserver: IT was defined this way in grades so I left it as is.

src/core/features/siteplugins/classes/handlers/user-handler.ts
dpalou: I see you now handle `TYPE_COMMUNICATION` in the `registerHandler` function of the delegate, but that's not enough. If a plugin defines "type" => "communication" in mobile.php, this check will see that it's different than `BUTTON` and will convert it to `LIST_ITEM`. But it should be converted to `BUTTON` instead.

dpalou (changes requested):

src/core/features/siteplugins/classes/handlers/user-handler.ts
crazyserver: Done :-)

PR #3956, opened 2024-03-05
MOBILE-4219 blog: Add and edit blog entries

alfonso-salces:


NoelDeMartin:
@alfonso-salces I rerun the tests just in case, but it seems like this PR has caused a regression in the glossary tests. There are also some merge conflicts you'll have to resolve.

src/addons/blog/components/entry-options-menu.html
```diff
@@ -0,0 +1,16 @@
...
+    <ion-list>
+        <ion-item button class="ion-text-wrap" (click)="action('edit')" [detail]="false">
+            <ion-icon name="fas-pen" slot="start" aria-hidden="true" />
+            <ion-label>
+                <p class="item-heading">{{ 'core.edit' | translate }}</p>
+            </ion-label>
+        </ion-item>
+        <ion-item button class="ion-text-wrap" (click)="action('delete')" [detail]="false">
+            <ion-icon name="fas-trash" slot="start" aria-hidden="true" color="danger" />
+            <ion-label color="danger">
```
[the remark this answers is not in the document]
alfonso-salces: This wont work because it will set the background-color as danger too.

src/addons/blog/pages/edit-entry/edit-entry.ts
```diff
@@ -0,0 +1,214 @@
...
+        if (selectedEntry.courseid !== 0) {
+            this.courseId = selectedEntry.courseid;
+            this.associatedCourse = await CoreCourses.getCourse(this.courseId);
+        }
+
+        this.courseId = selectedEntry.courseid;
+        this.modId = selectedEntry.coursemoduleid ? selectedEntry.coursemoduleid : this.route.snapshot.queryParams.cmId;
+
+        if (this.courseId) {
+            this.form.controls.associateWithCourse.patchValue(true);
```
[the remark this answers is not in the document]
alfonso-salces: According to the [official Angular docs](https://angular.io/guide/reactive-forms#updating-parts-of-the-data-model) are slightly different, `patchValue` is quitter unsafe than `setValue` because in complex nested forms `setValue` catch errors and `patchValue` silence them, so maybe in this case it doesn't care so much because we are not using nested controls but maybe we can take into account for other use cases, or directly use always `setValue`.

src/addons/blog/services/handlers/index-link.ts
```diff
@@ -45,7 +45,7 @@ export class AddonBlogIndexLinkHandlerService extends CoreContentLinksHandlerBas
 
         return [{
             action: async (siteId: string): Promise<void> => {
-                await CoreNavigator.navigateToSitePath('/blog', { params: pageParams, siteId });
+                await CoreNavigator.navigateToSitePath('/blog/index', { params: pageParams, siteId });
```
dpalou, line 48: This /index probably isn't needed either because the default root has a redirect, and the same happens in the user handler.

src/addons/blog/tests/behat/entries.feature
```diff
@@ -0,0 +1,30 @@
...
+    And I press "Blog entries" in the app
+    Then I should find "Blog post one" in the app
+    And I should find "Blog post two" in the app
+
+  Scenario: Delete blog entry
+    Given I entered the app as "testuser"
+    When I press the user menu button in the app
+    And I press "Blog entries" in the app
+    Then I should find "Blog post one" in the app
+    And I press "Display options" in the app
```
dpalou, line 27: I think this step is flaky because there will be 2 "Display options" in the app, one for each post. You should specify that you want to click the "Display options" of "Blog post one".

src/core/lang.json
```diff
@@ -12,6 +12,7 @@
     "areyousure": "Are you sure?",
     "back": "Back",
     "browser": "Browser",
+    "bynameondate": "By <strong> {{$a.name}} </strong> - {{$a.date}}",
```
dpalou, line 15: Are the spaces next to the name intended? The ones inside the strong tags. Just to be sure it's not adding some extra spaces.

src/addons/blog/services/blog.ts
```diff
@@ -62,27 +62,89 @@ export class AddonBlogProvider {
...
-     * @param page Page of the blog entries to fetch.
+     * @param options WS Options.
      * @param siteId Site ID. If not defined, current site.
      * @returns Promise to be resolved when the entries are retrieved.
      */
-    async getEntries(filter: AddonBlogFilter = {}, page: number = 0, siteId?: string): Promise<CoreBlogGetEntriesWSResponse> {
+    async getEntries(
+        filter: AddonBlogFilter = {},
+        options?: AddonBlogGetEntriesOptions,
+        siteId?: string,
```
dpalou, line 72: The options parameter you defined already accepts a siteId so there's no need to put it in a separate parameter.

src/addons/blog/services/blog.ts
```diff
@@ -96,6 +158,32 @@ export class AddonBlogProvider {
...
+    }
+
+    /**
+     * Is removing blog entry enabled.
+     *
+     * @param siteId Site ID.
+     * @returns is enabled or not.
+     * @since 4.4
+     */
+    async isRemovingEnabled(siteId?: string): Promise<boolean> {
```
dpalou, line 181: This function is no longer used, you can remove it.

src/addons/blog/pages/index/index.ts
```diff
@@ -82,6 +92,12 @@ export class AddonBlogEntriesPage implements OnInit {
                 }),
             });
         });
+
+        this.entryUpdateObserver = CoreEvents.on(ADDON_BLOG_ENTRY_UPDATED, async () => {
+            const loading = await CoreDomUtils.showModalLoading();
```
dpalou, line 97: When using `deleteEntry` I told you to use `showModalLoading` instead of `this.loaded = false`, but this is an event so I think this case is different because the event can be triggered by another page (anything in the app could trigger this event), and then the user would see a modal for something he isn't seeing in the screen.

src/addons/blog/pages/index/index.ts
```diff
@@ -287,13 +312,70 @@ export class AddonBlogEntriesPage implements OnInit {
...
+    /**
+     * Delete entry by id.
+     *
+     * @param id Entry id.
+     */
+    async deleteEntry(id: number): Promise<void> {
+        const loading = await CoreDomUtils.showModalLoading();
+        try {
+            await AddonBlog.deleteEntry({ entryid: id });
+            this.entries = this.entries.filter(entry => entry.id !== id);
```
dpalou, line 336: Filtering the entries locally is faster, but if the user leaves the page and opens it again he will see the deleted entry again because the list is cached in the app. You should probably refresh the list in this case too.

src/addons/blog/pages/index/index.scss
```diff
@@ -0,0 +1,75 @@
...
+:host {
+
+    ion-card {
+        padding: .5rem 1rem;
+    }
+
+    .entry {
+        border-top: 1px solid var(--stroke);
+
+        &-visibility-permission {
```
dpalou, line 12: Is this a valid selector? It's the first time I see using "&" to concatenate something that's not a valid CSS selector. IMO it would be better to put the whole ".entry-visibility-permission" here, it's easier to identify what you're doing. 

And the same applies to other things like "&-draft", "&-subject", etc.

dpalou (changes requested):

src/addons/blog/pages/index/index.scss
alfonso-salces: We have decided to avoid this change request.

src/addons/blog/pages/index/index.ts
```diff
@@ -59,6 +64,10 @@ export class AddonBlogEntriesPage implements OnInit {
     tagsEnabled = false;
     contextLevel: ContextLevel = ContextLevel.SYSTEM;
     contextInstanceId = 0;
+    entryUpdateObserver: CoreEventObserver;
+    optionsAvailable = false;
+
+    route = inject(ActivatedRoute);
```
dpalou, line 70: I think this `route` is no longer used and can be removed.

dpalou (changes requested):

dpalou (approved):

PR #3957, opened 2024-03-06
Mobile 3947

crazyserver:


src/core/features/compile/services/compile.ts
```diff
@@ -405,19 +356,119 @@ export class CoreCompileProvider {
      * @returns Lazy libraries.
      */
     async getLazyLibraries(): Promise<Type<unknown>[]> {
-        const ADDON_MOD_WORKSHOP_SERVICES = await getWorkshopServices();
-        const ADDON_COMPETENCY_SERVICES = await getCompetencyServices();
-        const ADDON_COURSECOMPLETION_SERVICES = await getCourseCompletionServices();
-
+        const CORE_SERVICES = await getCoreServices();
```
dpalou, line 359: Maybe we should use Promise.all to avoid serializing all these calls.

src/core/features/native/native.module.ts
```diff
@@ -36,30 +36,38 @@ import { StatusBar } from '@awesome-cordova-plugins/status-bar/ngx';
...
+        SplashScreen,
+        StatusBar,
+        SQLite,
+        WebIntent,
+        WebView,
+        Zip,
+    ];
+}
 
+// @TODO is this really needed here?
```
dpalou, line 70: I guess some of them are missing because they're declared in emulator.module. I think it's ok te declare them all here too if this doesn't cause any problem.

dpalou (changes requested):

src/core/features/native/native.module.ts
crazyserver: Ok, so finally we've decided to leave them here and add the ones that were missing.

src/core/features/compile/services/compile.ts
crazyserver: Yep Done! Additionally I've deprecated CoreArray.flatten since it's not needed anymore.

dpalou (approved):

dpalou (approved):

PR #3959, opened 2024-03-07
Mobile 4539

dpalou:


src/core/services/file-helper.ts
```diff
@@ -63,7 +63,7 @@ export class CoreFileHelperProvider {
         file: CoreWSFile,
         component?: string,
         componentId?: string | number,
-        state?: string,
+        state?: TDownloadStatus,
```
crazyserver, line 66: Based on changes on jsdocs of this function... did you want this to be renamed as status?

src/core/services/file-helper.ts
```diff
@@ -150,7 +150,7 @@ export class CoreFileHelperProvider {
         component?: string,
         componentId?: string | number,
         timemodified?: number,
-        state?: string,
+        state?: TDownloadStatus,
```
crazyserver, line 153: Same rename thing? There are lots, so probably you dont' want to make it...

crazyserver (changes requested):

src/core/services/file-helper.ts
dpalou: In some places we call it "state", in some others we call it "status". One of the definitions I found is:

State: the particular condition that someone or something is in at a specific time. Status: the situation at a particular time during a process. **They can be used in similar ways**.

Based on this definition, IMO it's ok to have both words.

Some people said that there's a small difference (e.g. one of them being "final" and the other being the condition during a process), but some pages said that "status" is the final condition while others said that state was the final condition ^^'

dpalou: Sorry, I didn't understand you. I reverted the JSDoc change.

crazyserver (approved):

PR #3960, opened 2024-03-07
MOBILE-4529: Decouple Survey

NoelDeMartin:


src/addons/mod/survey/services/handlers/prefetch.ts
```diff
@@ -12,103 +12,47 @@
...
-
-    /**
-     * @inheritdoc
-     */
-    async getIntroFiles(module: CoreCourseAnyModuleData, courseId: number): Promise<CoreWSFile[]> {
-        const survey = await CoreUtils.ignoreErrors(AddonModSurvey.getSurvey(courseId, module.id));
+let prefetchHandlerInstance: AsyncInstance<
+    AddonModSurveyPrefetchHandlerLazyService,
+    AddonModSurveyPrefetchHandlerService
+> | null = null;
```
dpalou, line 29: I see you added this code to make sure the instance is created only once, because `getPrefetchHandlerInstance` is called from several places. I was just wondering if we should always do this, otherwise if a dev adds a call to `getPrefetchHandlerInstance` without taking this into account we would end up with several instances. What do you think?

In any case, the idea is to review this in 4.5 so we can leave it like this and re-evaluate this later.

dpalou (changes requested):

src/addons/mod/survey/services/handlers/prefetch.ts
NoelDeMartin: In general I agree with standardizing practices, but in this case we're already doing all handlers very differently (some are lazy, some are not, etc.). So I think it's fine to leave it like this to avoid adding even more boilerplate to this pattern.

I was also surprised about this, because I thought handlers were only used through the delegates. So this may be a sign that the code should be elsewhere and that would be a better solution. But given that we're thinking on this as a temporary measure, I thought it'd be fine to leave it like this for now.

dpalou (approved):

PR #3963, opened 2024-03-08
Mobile 4526

dpalou:


src/core/features/filter/services/filter.ts
```diff
@@ -162,6 +191,29 @@ export class CoreFilterProvider {
...
 
+    /**
+     * Given a context level and instance ID, return the proper context to use.
+     *
+     * @param contextLevel The context level.
+     * @param instanceId Instance ID related to the context.
+     * @param options Options.
+     * @returns Context to use.
+     */
+    convertContext(
```
NoelDeMartin, line 202: Can we come up with a better naming for this? It wasn't clear at all what it was doing until I looked at the implementation. Maybe `getEffectiveContext` or something like that? or `getUsedContext`, etc. I'm not sure we can come up with something super intuitive, but maybe something that gives a better idea.

src/core/features/filter/services/filter.ts
```diff
@@ -470,20 +620,20 @@ export type CoreFiltersGetAvailableInContextWSParams = {
...
 export type CoreFilterFilter = {
-    contextlevel: string; // The context level where the filters are: (coursecat, course, module).
+    contextlevel: ContextLevel; // The context level where the filters are: (coursecat, course, module).
     instanceid: number; // The instance id of item associated with the context.
-    contextid: number; // The context id.
+    contextid?: number; // The context id. It will be undefined in cases where it cannot be calculated in the app.
     filter: string; // Filter plugin name.
-    localstate: number; // Filter state: 1 for on, -1 for off, 0 if inherit.
-    inheritedstate: number; // 1 or 0 to use when localstate is set to inherit.
+    localstate: CoreFilterStateValue; // Filter state: 1 for on, -1 for off, 0 if inherit.
```
NoelDeMartin, line 635: The comment says it can only be 1, -1, or 0; whilst the type says it can also be -9999. Which one is right? There are also other properties with the same problem.

src/core/features/filter/services/filter.ts
```diff
@@ -49,6 +50,8 @@ export class CoreFilterProvider {
         };
     } = {};
 
+    // @todo: Check if an "all states" memory cache is needed or not.
```
NoelDeMartin, line 53: What's keeping us from solving this TODO before merging? Is it too complicated? And how important is it? If it is important, I think we should open an issue and reference it here. Otherwise, leaving this TODO here is as good as not doing it all (and I think I'd just remove it). Also the TODO gives the impression that it may not be necessary at all. So I'd either remove it or assert that his is an actual TODO.

src/core/features/filter/services/filter.ts
dpalou: I used convertContext because I couldn't think of any better name, I changed it to getEffectiveContext.

src/core/features/filter/services/filter.ts
dpalou: In LMS, filters can have the state -9999, but this WebService only returns the ones that aren't -9999, that's why the comment didn't mention it (this comment is copied from the WS documentation).

I could use an `Omit` here to make the type more specific, but then the output of both WebServices isn't compatible and it makes it harder to handle the responses in the app. Since not returning filters with -9999 is an implementation detail of the WS I think it's better to put the generic type, specially because it makes it easier to convert the output of one WS to the other.

I removed the values from the comments since the type already specifies the possible values.

src/core/features/filter/services/filter.ts
dpalou: This was a note for myself and I forgot to remove it (my tests worked fine without this memory cache, so I didn't add it in the end). I removed the TODO.

NoelDeMartin (approved):

PR #3964, opened 2024-03-08
Mobile 4538

crazyserver:


src/addons/mod/assign/services/feedback-delegate.ts
```diff
@@ -188,6 +189,13 @@ export class AddonModAssignFeedbackDelegateService extends CoreDelegate<AddonMod
         super('AddonModAssignFeedbackDelegate');
     }
 
+    /**
+     * @inheritdoc
+     */
+    async isEnabled(): Promise<boolean> {
+        return !(await CoreSites.isFeatureDisabled('CoreCourseModuleDelegate_AddonModAssign'));
```
dpalou, line 196: All these awaits aren't needed, the isFeatureDisabled function is synchronous.

src/core/features/question/services/behaviour-delegate.ts
```diff
@@ -71,6 +72,13 @@ export class CoreQuestionBehaviourDelegateService extends CoreDelegate<CoreQuest
         super('CoreQuestionBehaviourDelegate');
     }
 
+    /**
+     * @inheritdoc
+     */
+    async isEnabled(): Promise<boolean> {
+        return !(await CoreSites.isFeatureDisabled('CoreCourseModuleDelegate_AddonModQuiz'));
+    }
```
dpalou, line 80: IMO this is wrong, question should be independent from quiz. Right now it's only used in quiz, but it was designed as a separate component so there should be no reference to quiz in question code. The same happens in question-delegate.

src/addons/mod/assign/feedback/comments/services/handler.ts
```diff
@@ -21,20 +21,20 @@ import {
...
 import { makeSingleton } from '@singletons';
 import { AddonModAssignFeedbackCommentsComponent } from '../component/comments';
+import { AddonModAssignDefaultFeedbackHandler } from '@addons/mod/assign/services/handlers/default-feedback';
 
 /**
  * Handler for comments feedback plugin.
  */
 @Injectable( { providedIn: 'root' })
-export class AddonModAssignFeedbackCommentsHandlerService implements AddonModAssignFeedbackHandler {
+export class AddonModAssignFeedbackCommentsHandlerService extends AddonModAssignDefaultFeedbackHandler {
```
dpalou, line 37: This change doesn't convince me. If I understood it right, the only "benefit" is that we don't need to specify an `isEnabled` function, but this can cause problems in the future like multiple inheritance. IMO the benefit isn't as important to do this.

dpalou (changes requested):

src/addons/mod/assign/feedback/comments/services/handler.ts
crazyserver: Ok thanks!

src/addons/mod/assign/services/feedback-delegate.ts
dpalou: Oh right, sorry.

dpalou (approved):

PR #3965, opened 2024-03-11
Mobile 4329

crazyserver:


src/core/features/dataprivacy/services/dataprivacy.ts
```diff
@@ -41,6 +42,10 @@ export class CoreDataPrivacyService {
             return false;
         }
 
+        if (site?.isFeatureDisabled(CORE_DATAPRIVACY_FEATURE_NAME)) {
+            return false;
+        }
```
dpalou, line 47: This check isn't needed here. The links delegate automatically checks it using featureName, and the user delegate checks it automatically too using the handler name and the delegate name.

src/core/features/dataprivacy/tests/behat/create_data_request.feature
```diff
@@ -59,4 +59,4 @@ Feature: Data export and delete from the privacy API
     When I run all adhoc tasks
     And I pull to refresh in the app
     And I should find "Download ready" near "Export all of my personal data" in the app
-    # TODO: Add download link and test it.
+    And I press "Download" in the app
```
dpalou, line 62: Shouldn't you test that the browser has been opened pointing to the site? You can use the exact URL or just that the site is opened, something like:

Then the app should have opened a browser tab with url "$WWWROOTPATTERN"

dpalou (changes requested):

src/core/features/dataprivacy/services/dataprivacy.ts
crazyserver: Thanks, removed

src/core/features/dataprivacy/tests/behat/create_data_request.feature
crazyserver: Done! waiting for tests

crazyserver:
Hi Dani! I've added one more commit.

dpalou (approved):

PR #3982, opened 2024-03-18
MOBILE-4268: Apply design system to Error Accordion

NoelDeMartin:


src/core/components/error-info/error-info.ts
```diff
@@ -64,7 +64,7 @@ export class CoreErrorInfoComponent implements OnInit, OnChanges {
     }
 
     @Input() errorDetails!: string;
-    @Input() errorCode?: string;
+    @Input() errorCode!: string;
```
crazyserver: Is this ! permitted or it throws a compilation warning,

src/core/components/error-accordion/error-accordion.ts
```diff
@@ -23,11 +23,11 @@ import { CoreForms } from '@singletons/form';
  * it can be rendered using the static render() method to get the raw HTML.
  */
 @Component({
-    selector: 'core-error-info',
-    templateUrl: 'core-error-info.html',
-    styleUrls: ['error-info.scss'],
+    selector: 'core-error-accordion',
```
crazyserver, line 31: We should probably deprecate the previous component, nit just remove It. Probably admiting both d'electors van be enough.

crazyserver: However se should revise with which components de are invoking with selectors and which creating them in TS code

src/core/components/error-accordion/error-accordion.ts
```diff
@@ -32,41 +37,91 @@ export class CoreErrorAccordionComponent implements OnInit, OnChanges {
...
+                    </div>
+                    <ion-icon name="chevron-down" />
+                </button>
             </div>
         `;
     }
 
-    @Input() errorDetails!: string;
+    /**
+     * Hydrage component.
```
crazyserver, line 87: Typo

crazyserver (changes requested):

src/core/components/error-accordion/error-accordion.ts
NoelDeMartin: I refactored the code to use a service instead, given that this isn't really an Angular Component. It was a component before to use it in storybook, but given that we've removed it now, it's no longer necessary.

PR #3990, opened 2024-03-22
Mobile 4543

crazyserver:


src/core/features/course/services/course-options-delegate.ts
```diff
@@ -403,45 +392,37 @@ export class CoreCourseOptionsDelegateService extends CoreDelegate<CoreCourseOpt
...
 
-        let handlerList: CoreCourseOptionsMenuHandler[] | CoreCourseOptionsHandler[];
-        if (menu) {
-            handlerList = this.coursesHandlers[course.id].enabledMenuHandlers;
-        } else {
-            handlerList = this.coursesHandlers[course.id].enabledHandlers;
-        }
+        const handlerList: CoreCourseOptionsMenuHandler[] | CoreCourseOptionsHandler[] = menu
+            ? this.coursesHandlers[course.id].enabledMenuHandlers
+            : this.coursesHandlers[course.id].enabledHandlers;
```
dpalou, line 425: Now that you're doing this in a single line I think there's no need to manually declare the types, they can be inferred.

src/core/features/course/services/course-options-delegate.ts
```diff
@@ -477,77 +458,13 @@ export class CoreCourseOptionsDelegateService extends CoreDelegate<CoreCourseOpt
...
     }
 
-    /**
-     * Check if a course has any handler enabled for default access, using course object.
-     *
-     * @param course The course object.
-     * @param refresh True if it should refresh the list.
-     * @returns Promise resolved with boolean: true if it has handlers, false otherwise.
-     */
-    async hasHandlersForCourse(course: CoreEnrolledCourseDataWithExtraInfoAndOptions, refresh = false): Promise<boolean> {
```
dpalou, line 487: Deleting these functions is a breaking change that can affect plugins. I guess it's unlikely that a plugin uses them, but we need to be careful when deleting things.

src/core/features/course/services/course.ts
```diff
@@ -104,6 +104,11 @@ export enum CoreCourseModuleCompletionTracking {
     COMPLETION_TRACKING_AUTOMATIC = 2,
 }
 
+export enum CoreCourseAccessDataType {
+    ACCESS_GUEST = 'courses_access_guest',
+    ACCESS_DEFAULT = 'courses_access_default',
+}
```
dpalou, line 110: Since you aren't using things like Object.values or assigning the whole enum to a variable, it's better to use `const enum` instead.

**However**, right now plugins have access to `CoreCourseProvider.ACCESS_GUEST`, and when those properties are deleted the plugins will no longer have access to those values because this `enum` isn't passed to plugins.

IF we want to pass this to plugins, then you should use `const+type` instead of enum. If we decide that plugins shouldn't need this then I would use `const enum`.

src/core/features/course/components/module-summary/module-summary.html
```diff
@@ -24,7 +24,8 @@ <h1>
                 </h1>
             </ion-label>
             <ion-button fill="clear" *ngIf="displayOptions.displayOpenInBrowser && externalUrl" [href]="externalUrl" core-link
-                [showBrowserWarning]="false" [attr.aria-label]="'core.openinbrowser' | translate" slot="end">
+                [showBrowserWarning]="false" [attr.aria-label]="'core.openinbrowser' | translate" slot="end" [class.hidden]="!isTeacher"
```
dpalou, line 27: Can't you use Ionic's `[hidden]` instead of a custom CSS class? Also in course summary.

src/core/features/course/components/module-summary/module-summary.ts
```diff
@@ -269,13 +270,14 @@ export class CoreCourseModuleSummaryComponent implements OnInit, OnDestroy {
...
     protected async fetchCourse(): Promise<void> {
-        // Fix that.
         try {
             this.course = await CoreCourses.getUserCourse(this.courseId, true);
         } catch {
             // The user is not enrolled in the course. Use getCourses to see if it's an admin/manager and can see the course.
             this.course = await CoreCourses.getCourse(this.courseId);
         }
+
+        this.isTeacher = await CoreCourseHelper.guessIsTeacher(this.courseId, this.course);
```
dpalou, line 280: IMO you should use `CoreUtils.ignoreErrors` since this function could fail. The same happens in course summary.

src/addons/mod/bigbluebuttonbn/tests/behat/basic_usage.feature
```diff
@@ -80,6 +80,7 @@ Feature: Test basic usage of BBB activity in app
 
     # Join the session as moderator in a browser.
     When I press "Information" in the app
+    And I set "core-course-module-summary ion-button.core-module-oib-button" styles to "display" "block"
```
dpalou, line 83: I don't like doing this, if this button is not supposed to be seen in the app in this case then IMO it shouldn't be used. IMO you should either use the steps to open a browser tab and navigate to the activity, or create a new custom step to open the activity in browser (if you have enough information to do so, and without using this button).

dpalou (changes requested):

src/core/features/course/components/module-summary/module-summary.html
dpalou: Ok, I understand the problem with BMAs. Let's keep it like this then.

dpalou (approved):

PR #3991, opened 2024-03-25
Mobile 4522

dpalou:


src/addons/mod/forum/services/handlers/module.ts
```diff
@@ -55,6 +57,29 @@ export class AddonModForumModuleHandlerService extends CoreModuleHandlerBase imp
     async getData(module: CoreCourseModuleData, courseId: number): Promise<CoreCourseModuleHandlerData> {
         const data = await super.getData(module, courseId);
 
+        const customData = module.customdata ?
+            CoreTextUtils.parseJSON<{ trackingtype?: string | number } | ''>(module.customdata, {}) : {};
```
alfonso-salces, line 61: Shouldn't we create a new const have more control of the `trackingtype` property?

Maybe we can try something like:

```
export const ADDON_MOD_FORUM_MODULE_TRACKINGTYPES = { off: 0, on: 1, forced: 2 } as const;

export type AddonModForumModuleTrackingTypes = typeof ADDON_MOD_FORUM_MODULE_TRACKINGTYPES[keyof typeof ADDON_MOD_FORUM_MODULE_TRACKINGTYPES];
```

src/addons/mod/forum/services/handlers/module.ts
```diff
@@ -55,6 +57,29 @@ export class AddonModForumModuleHandlerService extends CoreModuleHandlerBase imp
...
+        const trackingType = typeof customData !== 'string' && customData.trackingtype !== undefined ?
+            Number(customData.trackingtype) : undefined;
+
+        if (trackingType === 0) {
```
alfonso-salces, line 65: Following the comment above, we maybe we can do something like
`if (trackingType === ADDON_MOD_FORUM_MODULE_TRACKINGTYPES.off)`

alfonso-salces (changes requested):

src/addons/mod/forum/services/handlers/module.ts
dpalou: There are only 2 values for tracking: 0 (off) or 1 (enabled and optional). There is no forced tracking value. So IMO there's no need to create an enum/constant for this.

dpalou: Nice catch, I didn't notice there was a "forced" value (only available if an admin setting is enabled). I updated the code to use an enum.

alfonso-salces (approved):

PR #3993, opened 2024-03-25
Mobile 4219

alfonso-salces:


src/addons/blog/pages/index/index.html
```diff
@@ -92,20 +94,22 @@ <h3>
                 </div>
             </ion-label>
 
-            @if (entry.lastmodified > entry.created) {
-            <div class="entry-last-modification flex ion-justify-content-between border-bottom ion-padding-top ion-padding-bottom">
-                <ion-note class="flex ion-align-items-center">
+            @if (entry.lastmodified > entry.created || entry.userid === currentUserId && entry.publishstate !== 'draft') {
```
dpalou, line 97: I think this expression is wrong because it lacks parenthesis. Looking at the ifs a few lines below, I think that you want to do:

`entry.lastmodified > entry.created || (entry.userid === currentUserId && entry.publishstate !== 'draft')`

But you didn't add any parenthesis, so I think the code will actually do:

`(entry.lastmodified > entry.created || entry.userid === currentUserId) && entry.publishstate !== 'draft'`

These expressions aren't equivalent. My recommendation is to always put parenthesis when you mix different operators, even if they aren't needed, that way the execution order is more clear.

dpalou (changes requested):

dpalou (approved):

PR #4005, opened 2024-04-10
Mobile 4572

dpalou:


src/core/features/login/components/login-methods/login-methods.ts
```diff
@@ -92,16 +92,24 @@ export class CoreLoginMethodsComponent implements OnInit {
...
-            this.siteUrl,
-            provider,
-            this.siteConfig?.launchurl,
-            this.redirectData,
-        );
-
-        if (!result) {
-            CoreDomUtils.showErrorModal('Invalid data.');
+    async oauthClicked(provider: CoreSiteIdentityProvider): Promise<void> {
+        const modal = await CoreDomUtils.showModalLoading();
```
crazyserver, line 96: What to you think about including the modal and errors inside the helper functions this way it's not necessary to replacate all that code every time.

crazyserver (changes requested):

src/core/features/login/components/login-methods/login-methods.ts
dpalou: Done!

crazyserver (approved):

PR #4017, opened 2024-04-19
Mobile 4430 courses: Update progress changes

alfonso-salces:


src/core/features/course/pages/contents/contents.ts
```diff
@@ -315,20 +319,49 @@ export class CoreCourseContentsPage implements OnInit, OnDestroy, CoreRefreshCon
...
+            }
+
+            await CoreUtils.ignoreErrors(this.invalidateData());
             this.debouncedUpdateCachedCompletion?.();
+        } else {
+            await CoreUtils.ignoreErrors(this.invalidateData());
+            await this.showLoadingAndRefresh(true, false);
+        }
 
+        if (!('progress' in this.course) || !this.course.progress) {
```
dpalou, line 358: I think this check could be wrong. If progress is now 0, this if guard will prevent the event from being triggered but it's possible that the progress has changed and it's now 0, isn't it?

src/core/components/progress-bar/progress-bar.scss
```diff
@@ -47,6 +47,7 @@
         &[value]::-webkit-progress-value {
             background-color: var(--progressbar-color);
             border-radius: var(--height);
+            transition:width 500ms ease-in-out;
```
dpalou, line 50:
```suggestion
            transition: width 500ms ease-in-out;
```

dpalou (changes requested):

dpalou (approved):

PR #4020, opened 2024-04-24
MOBILE-4456 icons: Fix branded icons not being correctly coloured

crazyserver:


src/core/components/mod-icon/mod-icon.ts
```diff
@@ -133,8 +133,6 @@ export class CoreModIconComponent implements OnInit, OnChanges {
                 return;
             }
 
-            this.isBranded = false;
```
NoelDeMartin, line 136: I don't think this is working as expected. By removing this line, the method (which is called `setIsBranded`) no longer sets the `isBranded` property. Maybe we should remove the `return;` as well?

I also noticed that when `colorize` is false, it's not setting it either. It seems like it should set it to false, even if it "doesn't matter", because that's what the function is supposed to be doing. Otherwise, the following code would be possible which doesn't make any sense:

```
await this.setIsBranded();
console.log(this.isBranded); // undefined
```

Furthermore, the function has a `@returns` JSDoc saying that it returns a boolean, but the typescript declaration says it doesn't return anything.

It seems like this method needs some clean up.

NoelDeMartin (changes requested):

NoelDeMartin (approved):

PR #4021, opened 2024-04-24
MOBILE-4577 course: Implement course section link handler

NoelDeMartin:


src/core/features/courses/services/handlers/section-link.ts
```diff
@@ -0,0 +1,92 @@
...
+        // in this method will only use cache.
+
+        const courses = await CoreCourses.getUserCourses(true, siteId, CoreSitesReadingStrategy.ONLY_CACHE);
+
+        for (const course of courses) {
+            const courseSections = await CoreCourse.getSections(
+                course.id,
+                true,
+                true,
+                CoreSites.getReadingStrategyPreSets(CoreSitesReadingStrategy.ONLY_CACHE),
```
dpalou, line 75: The values of `excludeModules` and `excludeContents` can affect whether the cache is found or not, since they are parameters of the WebService. E.g. if you're using excludeModules=true but the app has only cached the request with excludeModules=false, this call will fail.

Since you don't care if the modules and contents are returned or not, you should use the preSet `getCacheUsingCacheKey=true`, that way the cached data is retrieved using the cache key (that doesn't include those params) instead of using the WS params.

```suggestion
                {
                    ...CoreSites.getReadingStrategyPreSets(CoreSitesReadingStrategy.ONLY_CACHE),
                    getCacheUsingCacheKey: true,
                }
```

dpalou (changes requested):

dpalou (approved):

PR #4022, opened 2024-04-24
Mobile 4579

crazyserver:


src/core/features/user/services/user-helper.ts
```diff
@@ -156,6 +157,39 @@ export class CoreUserHelperProvider {
...
+    /**
+     * Get the user initials.
+     *
+     * @param firstname User's first name. Recommended to use this param.
+     * @param lastname User's last name. Recommended to use this param.
+     * @param fullname User's full name. If firstname and lastname are not provided, this param will be used to get them.
+     * @param userId User ID. Recommended to use this param.
+     * @returns User initials.
+     */
+    async getUserInitialsFromParts(firstname?: string, lastname?: string, fullname?: string, userId?: number): Promise<string> {
```
dpalou, line 169: Instead of having 4 optional arguments, I think it's better to have a single parameter that is an object with those 4 optional properties. And then you can create overrides to make sure at least 1 of the required properties is supplied. E.g.:

```
type UserParts = { firstname?: string; lastname?: string; fullname?: string; userId?: number };
async getUserInitialsFromParts(parts: UserParts & Required<Pick<UserParts, 'firstname'|'lastname'>>): Promise<string>;
async getUserInitialsFromParts(parts: UserParts & Required<Pick<UserParts, 'fullname'>>): Promise<string>;
async getUserInitialsFromParts(parts: UserParts & Required<Pick<UserParts, 'userId'>>): Promise<string>;
async getUserInitialsFromParts(parts: UserParts): Promise<string> {
    ... // Implementation
```

src/addons/notifications/services/notifications.ts
```diff
@@ -155,6 +153,15 @@ export class AddonNotificationsProvider {
...
                     // Error getting user. This can happen if device is offline or the user is deleted.
                 }
+            } else {
+                let imgUrl = notification.customdata?.notificationpictureurl || notification.customdata?.notificationiconurl;
+
+                if (imgUrl) {
+                    imgUrl = String(imgUrl);
+                    if (!CoreUrlUtils.isUserAvatarUrl(imgUrl)) {
+                        notification.imgUrl = imgUrl;
+                    }
```
dpalou, line 163: I'm not sure if I understand this code. The problem we have here is that we are displaying the default avatar (theme/image.php) in some notifications, and we want to avoid that. But I think your code doesn't do that, you're hiding real avatars but not the default ones. And I'm not sure if we want to hide real avatars for all types of notifications.

I think that one of the solutions we discussed was always hiding the image for notifications with eventtype="newlogin", that solution will only work for those notifications but maybe that's enough for now. IMO we should double check with Juan.

src/core/services/utils/url.ts
```diff
@@ -565,6 +565,18 @@ export class CoreUrlUtilsProvider {
...
+     * Returns if the URL is a user avatar URL.
+     *
+     * @param url Image URL.
+     * @returns Whether the URL is a user avatar URL.
+     */
+    isUserAvatarUrl(url: string): boolean {
+        const matches = url.match('/tokenpluginfile.php/[0-9a-fA-F]+/[0-9]+/user/icon/(.*)/f[1-2](.*)');
+
+        return !!matches;
+    }
```
dpalou, line 578: Based on my previous comment, maybe this function is no longer needed. But if it is, then I think you should also check the pluginfile.php endpoint too.

dpalou (changes requested):

dpalou (approved):

PR #4029, opened 2024-04-30
Mobile 4560

dpalou:


src/core/features/question/services/question-helper.ts
```diff
@@ -799,38 +800,40 @@ export class CoreQuestionHelperProvider {
...
+            let iconName: string | undefined;
+            let color: string | undefined;
 
             if ('src' in icon) {
                 if ((icon as HTMLImageElement).src.indexOf('correct') >= 0) {
-                    correct = true;
-                } else if ((icon as HTMLImageElement).src.indexOf('incorrect') < 0 ) {
-                    return;
+                    iconName = 'check';
+                    color = 'success';
```
crazyserver, line 809: Better use CoreIonicColorNames here

crazyserver (changes requested):

src/core/features/question/services/question-helper.ts
dpalou: I added a new commit because I also modified a couple of places that didn't use this either.

crazyserver (approved):

PR #4033, opened 2024-05-02
MOBILE-4565 book: Force update slides size after loading

crazyserver:


crazyserver:
This has been already solved

PR #4038, opened 2024-05-07
Mobile 4470

crazyserver:


src/core/singletons/swiper.ts
```diff
@@ -0,0 +1,59 @@
...
+     * Initialize a Swiper instance.
+     *
+     * @param swiper Swiper instance.
+     * @param swiperOpts Swiper options.
+     * @returns Initialized Swiper instance.
+     */
+    static initSwiper(swiper: Swiper, swiperOpts?: SwiperOptions): Swiper {
+        Swiper.use([IonicSlides]);
+
+        this.updateOptions(swiper, swiperOpts);
```
dpalou, line 34: These functions are static, you shouldn't use "this".

src/core/classes/tabs.ts
```diff
@@ -66,25 +66,21 @@ export class CoreTabsBaseComponent<T extends CoreTabBase> implements OnInit, Aft
...
          */
         setTimeout(() => {
-            if (swiperRef?.nativeElement?.swiper && !this.swiper) {
-                this.swiper = swiperRef.nativeElement.swiper as Swiper;
+            if (!swiperRef?.nativeElement?.swiper || this.swiper) {
+                return;
+            }
 
-                this.swiper.changeLanguageDirection(CorePlatform.isRTL ? 'rtl' : 'ltr');
+            this.swiper = CoreSwiper.initSwiper(swiperRef.nativeElement.swiper, this.swiperOpts);
```
dpalou, line 73: Maybe this already happened, but the option slidesPerView can change dynamically but we aren't calling updateOptions after changing it. It would be nice to review if this is working as expected, and also check rich-text-editor since it's possible that it has the same problem.

dpalou (changes requested):

src/core/classes/tabs.ts
dpalou: To clarify Pau's comment, the slidesPerView is set using a swiper-container input in the HTML, e.g.:

`<swiper-container [slidesPerView]="swiperOpts.slidesPerView"`

src/core/classes/tabs.ts
```diff
@@ -66,25 +66,22 @@ export class CoreTabsBaseComponent<T extends CoreTabBase> implements OnInit, Aft
...
 
-                this.swiper.changeLanguageDirection(CorePlatform.isRTL ? 'rtl' : 'ltr');
+            this.swiper = CoreSwiper.initSwiper(swiper, this.swiperOpts);
 
-                Object.keys(this.swiperOpts).forEach((key) => {
-                    if (this.swiper) {
-                        this.swiper.params[key] = this.swiperOpts[key];
-                    }
-                });
+            this.swiper.changeLanguageDirection(CorePlatform.isRTL ? 'rtl' : 'ltr');
```
dpalou, line 76: Maybe this line should be inside `initSwiper`. Right now we don't do this in some places because there is only 1 slide in screen, but I guess it wouldn't hurt to do it. And this way we won't forget doing it when needed.

src/core/singletons/swiper.ts
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Return swiper instance if current is not set or destroyed and new is set and not destroyed.
+     *
+     * @param currentSwiper Current Swiper instance.
+     * @param newSwiperRef New Swiper Element Ref.
+     * @returns New swiper instance if needed.
+     */
+    static renewSwiper(currentSwiper?: Swiper, newSwiperRef?: ElementRef): Swiper | undefined {
```
dpalou, line 63: I'm not a fan of the name of this function, when I read "renew" I thought you would do some modifications on the swiper, but you are just validating and returning it. Maybe it would be better to call this "getNewSwiper", or "getNewValidSwiper" or something like this?

dpalou (changes requested):

dpalou (approved):

PR #4042, opened 2024-05-10
Mobile 4470

crazyserver:


src/core/features/editor/components/rich-text-editor/rich-text-editor.ts
```diff
@@ -78,17 +78,28 @@ export class CoreEditorRichTextEditorComponent implements OnInit, AfterViewInit,
     @Input() draftExtraParams?: Record<string, unknown>; // Extra params to identify the draft.
     @Output() contentChanged: EventEmitter<string | undefined | null>;
 
-    @ViewChild('editor') editor?: ElementRef; // WYSIWYG editor.
+    protected editorElement?: HTMLDivElement; // WYSIWYG editor.
+    @ViewChild('editor')  set editorRef(editorRef: ElementRef<HTMLDivElement>) {
+        this.editorElement = editorRef.nativeElement;
+    }
```
dpalou, line 84: I'm not a fan of creating a setter for no reason, setters are usually used to perform extra operations. IMO it's better to leave the ElementRef and then call .nativeElement when needed. You can change the type to ElementRef<HTMLDivElement> if you want a more specific type.

src/core/features/emulator/classes/wasm-sqlite-object.ts
```diff
@@ -61,20 +61,33 @@ export class WasmSQLiteObject implements SQLiteObject {
...
             const _promiser = sqlite3Worker1Promiser(() => resolve(_promiser));
         });
 
-        await promiser('open', { filename: `file:${this.name}.sqlite3`, vfs: 'opfs' });
+        const response = await promiser('config-get', {});
+        const isEnabled = (response as any).result.opfsEnabled;
+        if (!isEnabled) {
+            // @TODO Fix Workaround for the issue with the opfsEnabled flag.
+            // The flag gets disabled when opening a database, so we need to reload the page to make it work.
+            window.location.reload();
```
dpalou, line 74: Maybe we should write something in the log before reloading, so the developer knows why is the app reloading. Maybe you can use CoreLogger here instead of console directly.

src/core/features/emulator/classes/wasm-sqlite-object.ts
```diff
@@ -61,20 +61,33 @@ export class WasmSQLiteObject implements SQLiteObject {
...
+        }
+
+        try {
+            await promiser('open', { filename: `file:${this.name}.sqlite3`, vfs: 'opfs' });
+        } catch (error) {
+            console.error(error.result.message, this.name);
```
dpalou, line 80: I'm ok with leaving this console.error here (unless the error thrown already provides enough information to determine what happened). But as I said above, we can use CoreLogger and a better message like 'Error opening database {this.name}...'.

src/core/features/editor/components/rich-text-editor/rich-text-editor.ts
```diff
@@ -437,58 +437,12 @@ export class CoreEditorRichTextEditorComponent implements OnInit, AfterViewInit,
...
-        } else {
-            range.selectNode(parent);
-            range.setStart(parent, 0);
-            range = setRange(parent, range, { count: chars });
-        }
+    protected setCurrentCursorPosition(parent: Node): void {
+        const range = document.createRange();
+        // Select all so it will go to the end.
+        range.selectNode(parent);
+        range.selectNodeContents(parent);
```
dpalou, line 445: Deleting so much code about cursor position makes me a bit scared ^^' But I guess we'll notice if something goes wrong.

dpalou (changes requested):

dpalou (approved):

dpalou (approved):

PR #4043, opened 2024-05-10
MOBILE-4547 blog: Support offline blog

alfonso-salces:


src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    /**
+     * Sync all entries on a site.
+     *
+     * @param force Force sync.
+     * @param siteId Site ID to sync.
+     * @returns Promise resolved if sync is successful, rejected if sync fails.
+     */
+    protected async syncAllEntriesFunc(force: boolean, siteId: string): Promise<void> {
```
dpalou, line 63: The function that performs the sync (right now it's this one, but it could be a new function once you apply `getOngoingSync`/`addOngoingSync`) should return `AddonBlogSyncResult`. In this result you should add the warnings I mention in other comments so the page that triggers the manual sync can display them once the sync is done. Also, the "updated" property should be used to determine if the AUTO_SYNC event should be triggered or not.

dpalou (changes requested):

src/addons/blog/pages/index/index.html
```diff
@@ -121,12 +142,13 @@ <h3>
     </core-loading>
 
     <!-- Create a blog entry. -->
-    <ion-fab slot="fixed" core-fab vertical="bottom" horizontal="end"
-        *ngIf="loaded && optionsAvailable && (!filter.userid || (filter.userid && currentUserId === filter.userid))">
+    @if ((filter.userid === currentUserId || showMyEntriesToggle) && loaded() && optionsAvailable) {
```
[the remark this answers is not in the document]
alfonso-salces: This is done because showMyEntriesToggle already checks filter.userid

src/addons/blog/pages/edit-entry/edit-entry.ts
```diff
@@ -368,11 +424,33 @@ export class AddonBlogEditEntryPage implements CanLeave, OnInit {
             { name: 'modassoc', value: this.form.controls.associateWithModule.value && this.modId ? this.modId : 0 },
         ];
 
-        this.addAttachments(attachmentsId, options);
+        if (params.attachmentsId) {
+            options.push({
+                name: 'attachmentsid',
+                value: typeof params.attachmentsId === 'number' ? params.attachmentsId : JSON.stringify(params.attachmentsId),
```
dpalou, line 430: The whole `options` array is stored as a JSON string in database, so there's no need to convert the attachmentsid to a JSON string here. You can leave it as an object and it will be converted along with the rest of options.

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,254 @@
...
+                subject: entry.subject,
+                summary: entry.summary,
+                summaryformat: entry.summaryformat,
+                options: JSON.parse(entry.options),
+                created: entry.created,
+            };
+
+            try {
+                if (entry.id) {
+                    await this.syncUpdatedEntry({ ...entry, id: entry.id, options: JSON.parse(entry.options) });
```
dpalou, line 107: You aren't passing the siteId to `syncUpdatedEntry` or `uploadAttachments`. Please remember to always pass the siteId, otherwise things will go wrong when synchronizing data from a site that isn't the current one.

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,254 @@
...
+     * @returns draftId.
+     */
+    protected async uploadAttachments(params: AddonBlogSyncUploadAttachmentsParams, siteId?: string): Promise<number | undefined> {
+        try {
+            const site = await CoreSites.getSite(siteId);
+            const offlineFiles = await AddonBlogOffline.getOfflineFiles('created' in params ? params.created : params.entryId);
+
+            if (offlineFiles.length === 0) {
+                return;
+            }
```
dpalou, line 180: Even if there are no offline files to upload, when editing an existing entry the user could have deleted attachments so there could be files to delete.

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,254 @@
...
+                return await CoreFileUploader.uploadOrReuploadFiles(
+                    offlineFiles,
+                    AddonBlogProvider.COMPONENT,
+                    params.created,
+                    site.id,
+                );
+            }
+
+            const attachments = AddonBlog.getAttachmentFilesFromOptions(params.options);
+            const filesToDelete = CoreFileUploader.getFilesToDelete(attachments.online, offlineFiles);
```
dpalou, line 192: You fixed part of the function, but broke the part that was ok xD

`attachments.online` contains the "online" attachments that the user left when saving, but that's not the list of "original" attachments because the user could have deleted some files. You need to retrieve the original attachments using `AddonBlog.getEntries` like you did before, the code would be something like this (please notice it's pseudo code):

```
const onlineEntry = AddonBlog.getEntries(...);

const filesToDelete = CoreFileUploader.getFilesToDelete(onlineEntry.attachmentfiles, attachments.online);
```

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,254 @@
...
+            if (filesToDelete.length) {
+                await CoreFileUploader.deleteDraftFiles(params.attachmentsId, filesToDelete, site.id);
+            }
+
+            await CoreFileUploader.uploadFiles(params.attachmentsId, [...attachments.online, ...offlineFiles], site.id);
+        } catch (error) {
+            if (error instanceof CoreError || CoreUtils.isWebServiceError(error)) {
+                throw error;
+            }
+        }
```
dpalou, line 203: I think this `catch` isn't needed, isn't it? It's just ignoring some errors that I think we shouldn't ignore. The calls to `uploadAttachments` are already inside another try/catch so the errors will be handled in there.

src/addons/blog/services/blog-offline.ts
```diff
@@ -0,0 +1,224 @@
...
+export class AddonBlogOfflineService {
+
+    /**
+     * Delete an offline entry.
+     *
+     * @param created Entry creation date.
+     * @param siteId Site ID. If not defined, current site.
+     * @returns Promise resolved if deleted, rejected if failure.
+     */
+    async deleteOfflineEntryRecord(created: number, siteId?: string): Promise<void> {
```
dpalou, line 43: I think deleting entries by `created` isn't the best, because `created` might not be unique in the server. In a Moodle site there could be 2 entries with the same created time, you should only use created when you don't have an id.

src/addons/blog/services/blog-sync.ts
dpalou, line 107:
```suggestion
                    await this.syncUpdatedEntry({ ...entry, id: entry.id, options: formattedEntry.options });
```
You already parsed the options so you can reuse them instead of parsing them again, right?

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,254 @@
...
+
+                return { updated: true, warnings: [] };
+            } catch (error) {
+                if (!error || !CoreUtils.isWebServiceError(error)) {
+                    return { updated: false, warnings: [error] };
+                }
+
+                const warnings = [];
+                await AddonBlogOffline.deleteOfflineEntryRecord(entry.created);
+                this.addOfflineDataDeletedWarning(warnings, entry.summary, error);
```
dpalou, line 130:
```suggestion
                this.addOfflineDataDeletedWarning(warnings, entry.subject, error);
```
I think it's better to use subject, right? Summary is the whole post contents, it can be too long.

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,254 @@
...
+
+                return { updated: false, warnings };
```
dpalou, line 132: You're not sending data to server but you're "discarding" data, so the data displayed to the user was updated. You should return updated=true in this case.

src/addons/blog/services/blog.ts
```diff
@@ -91,10 +98,34 @@ export class AddonBlogProvider {
...
+        };
+
+        if (!CoreNetwork.isOnline()) {
+            return await storeOffline();
+        }
+
+        try {
+            return await site.write('core_blog_add_entry', params);
+        } catch (error) {
+            if (error && !CoreUtils.isWebServiceError(error)) {
```
dpalou, line 121:
```suggestion
            if (!CoreUtils.isWebServiceError(error)) {
```
I think there's no need to check if `error` is set, `isWebServiceError` is already checking it. And if `error` is empty we should probably consider it is as a local error, which `isWebServiceError` already does.

The same happens in the `updateEntry` and `deleteEntry` functions.

src/addons/blog/services/blog.ts
```diff
@@ -103,10 +134,53 @@ export class AddonBlogProvider {
...
+                await AddonBlogOffline.updateEntryOffline(
+                    { ...content, id: params.entryid, created: new Date().getTime() },
+                    { id: params.entryid },
+                );
+            } else if ('created' in params && params.created) {
+                await AddonBlogOffline.updateEntryOffline(
+                    { ...content, id: -params.created, created: params.created },
+                    { created: params.created },
+                );
+            }
```
dpalou, line 162: IMO the code to set the id to a negative number should be all inside blog-offline. In a previous comment I mentioned that we can probably remove the `updateEntryOffline` function, and the function `addOfflineEntry` already handles it correctly inside the function.

src/addons/blog/services/blog.ts
```diff
@@ -134,10 +208,28 @@ export class AddonBlogProvider {
...
+            if (!CoreNetwork.isOnline()) {
+                return await AddonBlogOffline.markEntryAsRemoved(params.entryid, siteId);
+            }
+
+            await site.write('core_blog_delete_entry', params);
+            const isMarkAsRemoved = await AddonBlogOffline.getEntryToRemove(params.entryid);
+
+            if (isMarkAsRemoved) {
+                await AddonBlogOffline.unmarkEntryAsRemoved(params.entryid);
+            }
```
dpalou, line 224:
```suggestion
            await CoreUtils.ignoreErrors(AddonBlogOffline.unmarkEntryAsRemoved(params.entryid));
```
Instead of reading and writing, you can just write and ignore the errors. And the `ignoreErrors` can probably be removed too because you delete using a WHERE, so it won't throw an error if no entries are found, but you can leave it just in case.

src/addons/blog/services/blog.ts
```diff
@@ -183,6 +275,102 @@ export class AddonBlogProvider {
...
+        const options: AddonBlogAddEntryOption[] = JSON.parse(offlineEntry.options);
+        const moduleid = options?.find(option => option.name === 'modassoc')?.value as number | undefined;
+        const courseid = options?.find(option => option.name === 'courseassoc')?.value as number | undefined;
+        const tags = options?.find(option => option.name === 'tags')?.value as string | undefined;
+        const publishstate = options?.find(option => option.name === 'publishstate')?.value as AddonBlogPublishState
+            ?? AddonBlogPublishState.draft;
+        const user = await CoreUser.getProfile(offlineEntry.userid);
+        const offlineFiles = await AddonBlogOffline.getOfflineFiles(offlineEntry.id ?? offlineEntry.created);
+        const optionsFiles = this.getAttachmentFilesFromOptions(options);
+        const attachmentfiles = [...optionsFiles.online, ...offlineFiles];
```
dpalou, line 298:
```suggestion
        const moduleId = options?.find(option => option.name === 'modassoc')?.value as number | undefined;
        const courseId = options?.find(option => option.name === 'courseassoc')?.value as number | undefined;
        const tags = options?.find(option => option.name === 'tags')?.value as string | undefined;
        const publishState = options?.find(option => option.name === 'publishstate')?.value as AddonBlogPublishState
            ?? AddonBlogPublishState.draft;
        const user = await CoreUser.getProfile(offlineEntry.userid);
        const offlineFiles = await AddonBlogOffline.getOfflineFiles(offlineEntry.id ?? offlineEntry.created);
        const optionsFiles = this.getAttachmentFilesFromOptions(options);
        const attachmentFiles = [...optionsFiles.online, ...offlineFiles];
```
Our variables should use camelCase, unless when we're using destructuring and the name comes from a WebService response.

src/addons/blog/services/blog.ts
```diff
@@ -282,14 +467,14 @@ export type AddonBlogAddEntryWSParams = {
     options: AddonBlogAddEntryOption[];
 };
 
-export type AddonBlogUpdateEntryWSParams = AddonBlogAddEntryWSParams & { entryid: number };
+export type AddonBlogUpdateEntryWSParams = AddonBlogAddEntryWSParams & ({ entryid: number } | { created?: number });
```
dpalou, line 470: `created` is not a valid parameter of the WebService. The WSParams type should reflect what the WebService expects, the other parameters used by the app should be in a different type (e.g. you can just put it in the function declaration like you do with `attachments` and `forceOffline`:

`{ attachments?: string; forceOffline?: boolean }`

dpalou (changes requested):

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,255 @@
...
+
+                await AddonBlog.addEntryOnline(formattedEntry, siteId);
+                await AddonBlogOffline.deleteOfflineEntryRecord({ created: entry.created }, siteId);
+                result.updated = true;
+            } catch (error) {
+                if (!error || !CoreUtils.isWebServiceError(error)) {
+                    throw error;
+                }
+
+                await AddonBlogOffline.deleteOfflineEntryRecord(entry.id ? { id: entry.id } : { created: entry.created });
```
dpalou, line 141:
```suggestion
                await AddonBlogOffline.deleteOfflineEntryRecord(entry.id ? { id: entry.id } : { created: entry.created }, siteId);
```

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,255 @@
...
+
+        if (optionsAttachmentsId) {
+            optionsAttachmentsId.value = attachmentsid;
+        } else {
+            entry.options.push({ name: 'attachmentsid', value: attachmentsid });
+        }
+
+        const { options, subject, summary, summaryformat, id } = entry;
+        await AddonBlog.updateEntryOnline({ options, subject, summary, summaryformat, entryid: id }, siteId);
+        await AddonBlogOffline.deleteOfflineEntryRecord({ id });
```
dpalou, line 169:
```suggestion
        await AddonBlogOffline.deleteOfflineEntryRecord({ id }, siteId);
```

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,255 @@
...
+     * Upload attachments.
+     *
+     * @param params entry creation date or entry ID and attachments ID.
+     *
+     * @returns draftId.
+     */
+    protected async uploadAttachments(params: AddonBlogSyncUploadAttachmentsParams, siteId?: string): Promise<number | undefined> {
+        const site = await CoreSites.getSite(siteId);
+        const folder = 'created' in params ? { created: params.created } : { id: params.entryId };
+        const offlineFiles = await AddonBlogOffline.getOfflineFiles(folder);
```
dpalou, line 182:
```suggestion
        const offlineFiles = await AddonBlogOffline.getOfflineFiles(folder, site.id);
```

src/addons/blog/services/blog-offline.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    /**
+     * Retrieve a list of offline files stored.
+     *
+     * @param folderName Folder name.
+     * @returns Offline files for the provided folder name.
+     */
+    async getOfflineFiles(folderName: AddonBlogOfflineParams): Promise<CoreFileEntry[]> {
+        try {
+            const folderPath = await AddonBlogOffline.getOfflineEntryFilesFolderPath(folderName);
```
dpalou, line 164:
```suggestion
    async getOfflineFiles(folderName: AddonBlogOfflineParams, siteId?: string): Promise<CoreFileEntry[]> {
        try {
            const folderPath = await AddonBlogOffline.getOfflineEntryFilesFolderPath(folderName, siteId);
```

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,255 @@
...
+                offlineFiles,
+                AddonBlogProvider.COMPONENT,
+                params.created,
+                site.id,
+            );
+        }
+
+        const { entries } = await AddonBlog.getEntries(
+            { entryid: params.entryId },
+            { readingStrategy: CoreSitesReadingStrategy.PREFER_NETWORK },
```
dpalou, line 195:
```suggestion
            { readingStrategy: CoreSitesReadingStrategy.PREFER_NETWORK, siteId: site.id },
```

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,255 @@
...
+
+    /**
+     * Sync entries to remove.
+     *
+     * @param siteId Site ID.
+     * @returns Entries to remove and result.
+     */
+    protected async syncEntriesToRemove(siteId?: string): Promise<AddonBlogSyncGetPendingToSyncEntries> {
+        let entriesToSync = await AddonBlogOffline.getOfflineEntries(undefined, siteId);
+        const entriesToBeRemoved = await AddonBlogOffline.getEntriesToRemove();
```
dpalou, line 217:
```suggestion
        const entriesToBeRemoved = await AddonBlogOffline.getEntriesToRemove(siteId);
```

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,255 @@
...
+        const warnings = [];
+
+        await Promise.all(entriesToBeRemoved.map(async (entry) => {
+            try {
+                await AddonBlog.deleteEntryOnline({ entryid: entry.id });
```
dpalou, line 222:
```suggestion
                await AddonBlog.deleteEntryOnline({ entryid: entry.id }, siteId);
```

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,255 @@
...
+                const entriesPendingToSync = entriesToSync.filter(entryToSync => entryToSync.id !== entry.id);
+
+                if (entriesPendingToSync.length !== entriesToSync.length) {
+                    await AddonBlogOffline.deleteOfflineEntryRecord({ id: entry.id });
```
dpalou, line 226:
```suggestion
                    await AddonBlogOffline.deleteOfflineEntryRecord({ id: entry.id }, siteId);
```

src/addons/blog/services/blog-sync.ts
```diff
@@ -0,0 +1,255 @@
...
+                    entriesToSync = entriesPendingToSync;
+                }
+            } catch (error) {
+                if (!CoreUtils.isWebServiceError(error)) {
+                    throw error;
+                }
+
+                await AddonBlogOffline.unmarkEntryAsRemoved(entry.id);
```
dpalou, line 234:
```suggestion
                await AddonBlogOffline.unmarkEntryAsRemoved(entry.id, siteId);
```

dpalou (changes requested):

dpalou (approved):

PR #4045, opened 2024-05-13
Mobile 4470

crazyserver:


src/core/components/combobox/combobox.ts
```diff
@@ -50,14 +49,26 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
...
     @Input() label = Translate.instant('core.show'); // Aria label.
     @Input() disabled = false;
     @Input() selection = '';
     @Output() onChange = new EventEmitter<unknown>(); // Will emit an event the value changed.
 
     // Additional options when interface modal is selected.
     @Input() icon?: string; // Icon for modal interface.
+
+    /**
+     * Interface to use, popover or modal.
```
NoelDeMartin, line 68: I think this description is not right, given that it is the same as the `interface` input.

src/core/components/combobox/combobox.ts
```diff
@@ -50,14 +49,26 @@ import { ControlValueAccessor, NG_VALUE_ACCESSOR } from '@angular/forms';
 })
 export class CoreComboboxComponent implements ControlValueAccessor {
 
+    /**
+     * Interface to use, popover or modal.
+     *
+     * @deprecated since 4.4. In a near future, modals won't be supported.
```
NoelDeMartin, line 55: Should we add this to upgrade.txt? Plugins can use this component, right?

src/theme/components/ion-item.scss
```diff
@@ -39,16 +39,12 @@ ion-item.item {
         }
     }
 
+
```
NoelDeMartin, line 42: It seems like this newline was added by mistake.

src/core/components/combobox/combobox.ts
crazyserver: Sorry, I've double checked the usage of the component in workplace and we still use the modal option there so I'm not going to deprecate this yet.

NoelDeMartin (approved):

PR #4049, opened 2024-05-15
Mobile 4470

crazyserver:


src/addons/mod/quiz/components/question-card/question-card.ts
```diff
@@ -0,0 +1,51 @@
...
+        // Get the readable mark for each question.
+        this.readableMark = AddonModQuizHelper.getQuestionMarkFromHtml(this.question.html);
+
+        // Extract the question info box.
+        CoreQuestionHelper.extractQuestionInfoBox(this.question, '.info');
+
+        // Check if the question is blocked. If it is, treat it as a description question.
+        if (!this.inReview && AddonModQuiz.isQuestionBlocked(this.question)) {
+            this.question.type = 'description';
+        }
```
dpalou, line 48: I think it's better to leave this code where it was. This code modifies the question html and type, and these properties are then used by the question component and I'm scared we could be causing some race condition here.

dpalou (changes requested):

PR #4052, opened 2024-05-16
Mobile 4470

crazyserver:


src/core/directives/auto-focus.ts
```diff
@@ -31,7 +31,7 @@ import { CoreDom } from '@singletons/dom';
 })
 export class CoreAutoFocusDirective implements AfterViewInit {
 
-    @Input('core-auto-focus') autoFocus: boolean | string = true;
+    @Input('core-auto-focus') autoFocus = true;
```
dpalou, line 34: Please leave this input as it was, this is a breaking change. In future versions we should check Angular input transforms (we couldn't use them before, but now that we've upgraded Angular we can).

dpalou (changes requested):

dpalou (approved):

PR #4057, opened 2024-05-22
Mobile 4470

crazyserver:


src/core/features/comments/components/comments/comments.ts
```diff
@@ -38,7 +39,7 @@ export class CoreCommentsCommentsComponent implements OnInit, OnChanges, OnDestr
     @Input() itemId!: number;
     @Input() area = '';
     @Input() title?: string;
-    @Output() onLoading: EventEmitter<boolean>; // Event that indicates whether the component is loading data.
+    @Output() onLoading = new EventEmitter<boolean>();  // Event that indicates whether the component is loading data.
```
dpalou, line 42: You're initializing the Output here so you can remove it from the constructor.

src/core/features/settings/pages/licenses/licenses.ts
```diff
@@ -52,7 +52,7 @@ export class CoreSettingsLicensesPage implements OnInit {
     protected allLicenses: CoreSettingsLicense[] = [];
 
     constructor() {
-        this.appLicenseVersion = CoreConstants.CONFIG.versionname.indexOf('-') > 0
+        this.appLicenseVersion = CoreConstants.BUILD.isDevelopment
```
dpalou, line 55: In our webapps do we use isDevelopment = true? Because in that case our "latest" webapp will also say "main", it could be confusing.

dpalou (changes requested):

src/core/features/settings/pages/licenses/licenses.ts
crazyserver: Hmmm the problem is that we're not using -dev anymore... so... what should we do? webapps are not production sites I think it's not a problem.

src/core/features/comments/components/comments/comments.ts
crazyserver: Ooops that was the purpose!

PR #4068, opened 2024-05-27
MOBILE-4470 storagemanager: Fix removing courses

NoelDeMartin:
Fixes a regression caused by https://github.com/moodlehq/moodleapp/commit/91c56256ed40caa6479acd9c2233410d4a25e58f


src/addons/storagemanager/pages/courses-storage/courses-storage.ts
```diff
@@ -50,6 +51,8 @@ export class AddonStorageManagerCoursesStoragePage implements OnInit, OnDestroy
     courseStatusObserver?: CoreEventObserver;
     siteId: string;
 
+    private downloadedCoursesLock = new CoreSemaphore();
```
dpalou, line 54: Instead of implementing a new `CoreSempahore` for this, maybe you can use the `CoreQueueRunner` class. If you do it, I guess you'll need to use allowRepeated=true.

dpalou (changes requested):

src/addons/storagemanager/pages/courses-storage/courses-storage.ts
NoelDeMartin: Thanks for the suggestion, I didn't realize we could use an existing class.

I've refactored the method signature to make it easier to use when you don't need to provide an id, that way it's also possible to achieve what we want without using `allowRepeated`. In fact, I think we could remove `allowRepeated` altogether and use this approach, but I prefer not to modify so much of the existing code since we're in QA now. But it may be worth exploring in the future.

dpalou (approved):

PR #4070, opened 2024-05-28
MOBILE-4470 core: Refactor mod-icon to use signals

NoelDeMartin:
This change is necessary because some icons weren't updated properly inside of components using OnPush (e.g. course-storage page).

### DIFF
--- a/src/core/components/mod-icon/mod-icon.ts
+++ b/src/core/components/mod-icon/mod-icon.ts
@@ -13,7 +13,17 @@
 // limitations under the License.
 
 import { CoreConstants, ModPurpose } from '@/core/constants';
-import { Component, ElementRef, HostBinding, Input, OnChanges, OnInit, SimpleChange } from '@angular/core';
+import {
+    ChangeDetectionStrategy,
+    Component,
+    ElementRef,
+    HostBinding,
+    Input,
+    OnChanges,
+    OnInit,
+    SimpleChange,
+    signal,
+} from '@angular/core';
 import { SafeHtml } from '@angular/platform-browser';
 import { CoreCourse } from '@features/course/services/course';
 import { CoreCourseModuleDelegate } from '@features/course/services/module-delegate';
@@ -42,6 +52,7 @@ const enum IconVersion {
     selector: 'core-mod-icon',
     templateUrl: 'mod-icon.html',
     styleUrls: ['mod-icon.scss'],
+    changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class CoreModIconComponent implements OnInit, OnChanges {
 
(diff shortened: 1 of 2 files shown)


src/core/components/mod-icon/mod-icon.ts
```diff
@@ -133,7 +143,7 @@ export class CoreModIconComponent implements OnInit, OnChanges {
         this.brandedClass = this.isBranded;
 
         // No icon or local icon (not legacy), colorize it.
-        if (!this.iconUrl || this.isLocalUrl) {
+        if (!this.iconUrl || this.isLocalUrl()) {
```
dpalou, line 146:
```suggestion
        if (!this.iconUrl() || this.isLocalUrl()) {
```
Shouldn't this be "this.iconUrl()"? I guess using "!this.iconUrl" will always return true because the signal instance exists, we want to check if the value is empty.

src/core/components/mod-icon/mod-icon.ts
```diff
@@ -146,11 +156,11 @@ export class CoreModIconComponent implements OnInit, OnChanges {
             return;
         }
 
-        this.iconUrl = CoreTextUtils.decodeHTMLEntities(this.iconUrl);
+        this.iconUrl.update(value => CoreTextUtils.decodeHTMLEntities(value));
 
         // If it's an Moodle Theme icon, check if filtericon is set and use it.
-        if (this.iconUrl && CoreUrlUtils.isThemeImageUrl(this.iconUrl)) {
-            const filter = CoreUrlUtils.getThemeImageUrlParam(this.iconUrl, 'filtericon');
+        if (this.iconUrl && CoreUrlUtils.isThemeImageUrl(this.iconUrl())) {
```
dpalou, line 162:
```suggestion
        if (CoreUrlUtils.isThemeImageUrl(this.iconUrl())) {
```
I think this part can be removed, I guess it was added at some point to avoid supplying an undefined value to isThemeImageUrl but I think this can no longer happen.

src/core/components/mod-icon/mod-icon.ts
```diff
@@ -169,14 +179,14 @@ export class CoreModIconComponent implements OnInit, OnChanges {
...
     }
 
     /**
      * Set icon.
      */
     async setIcon(): Promise<void> {
-        this.iconUrl = this.modicon || this.iconUrl;
+        this.iconUrl.update(value => this.modicon || value);
 
         if (!this.iconUrl) {
```
dpalou, line 191:
```suggestion
        if (!this.iconUrl()) {
```

src/core/components/mod-icon/mod-icon.ts
```diff
@@ -185,16 +195,17 @@ export class CoreModIconComponent implements OnInit, OnChanges {
...
+        this.isLocalUrl.set(this.iconUrl().startsWith(assetsPath));
 
         // Cache icon if the url is not the theme generic one.
         // If modname is not set icon won't be cached.
         // Also if the url matches the regexp (the theme will manage the image so it's not cached).
-        this.linkIconWithComponent =
+        this.linkIconWithComponent.set(
             !!this.modname &&
             !!this.componentId &&
             !this.isLocalUrl &&
```
dpalou, line 206:
```suggestion
            !this.isLocalUrl() &&
```

src/core/components/mod-icon/mod-icon.ts
```diff
@@ -317,9 +328,9 @@ export class CoreModIconComponent implements OnInit, OnChanges {
         if (!this.isLocalUrl) {
```
dpalou, line 328:
```suggestion
        if (!this.isLocalUrl()) {
```

dpalou (changes requested):

dpalou (approved):

PR #4073, opened 2024-05-28
MOBILE-4470 core: Inject mod-icon svg into Shadow DOM

NoelDeMartin:
This was necessary in order to completely isolate the styles of an injected SVG. It was previously handled in https://github.com/moodlehq/moodleapp/commit/5735bc4ca96a793fdb555f91acde9ca167cc20d6, but that didn't suffice because some SVGs are styled using classes, not ids.

In v4.3 we didn't need to do any of this because SVGs were displayed using an img tag, so the styles were always isolated.

### DIFF
--- a/src/core/components/mod-icon/mod-icon.scss
+++ b/src/core/components/mod-icon/mod-icon.scss
@@ -40,7 +40,8 @@
             @each $type, $value in $activity-icon-background-colors {
                 &.#{$type}:not(.branded) {
                     background-color: var(--activity-40-#{$type});
-                    ::ng-deep svg, img {
+                    img,
+                    ::ng-deep ::part(svg) {
                         filter: brightness(0) invert(1);
                     }
                 }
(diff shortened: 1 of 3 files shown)


NoelDeMartin:
I'm opening this as a draft because it seems like the colors of custom icons are still no the same as the LMS, but I'm not sure what's the proper approach in this case. We need to discuss further before deciding how to tackle this.

src/core/components/mod-icon/mod-icon.scss
```diff
@@ -58,14 +59,14 @@
     }
 
     &.colorize.version_current:not(.branded) {
-        ::ng-deep svg,
```
crazyserver, line 61: Only to be sure, Is this ng-deep still needed with the shadow Dom?

NoelDeMartin:
Closed in favour of #4079.

PR #4075, opened 2024-05-30
Mobile 4470 blog: Fix activity associations and clear rich text in new entries

alfonso-salces:


src/addons/blog/pages/edit-entry/edit-entry.html
```diff
@@ -23,8 +23,7 @@ <h1>{{ entry ? entry.subject : 'addon.blog.addnewentry' | translate }}</h1>
                 <ion-label position="stacked" for="addon_blog_entry_body">{{ 'addon.blog.entrybody' | translate }}</ion-label>
                 <core-rich-text-editor name="addon_blog_entry_body" [attr.aria-label]="'addon.blog.entrybody' | translate"
                     [control]="form.controls.summary" [placeholder]="'addon.blog.entrybody' | translate" [componentId]="component"
-                    [autoSave]="true" [contextInstanceId]="contextInstanceId" [contextLevel]="contextLevel"
-                    [elementId]="entry?.id ?? 'new_entry'" />
+                    [autoSave]="true" [contextInstanceId]="contextInstanceId" [contextLevel]="contextLevel" elementId="new_entry" />
```
dpalou, line 26: You should leave the elementId input as it was, otherwise all the entries will shame the same draft. Drafts are used for example when the app crashes or it's killed and then you open the view again, to restore the info you had in the editor. It's important that each entry stores its own draft to avoid displaying the draft in other entries.

src/addons/blog/pages/edit-entry/edit-entry.ts
```diff
@@ -354,13 +359,19 @@ export class AddonBlogEditEntryPage implements CanLeave, OnInit {
...
+        const response = this.entry
             ? await AddonBlog.updateEntry({ subject, summary, summaryformat: 1, options , entryid: this.entry.id })
             : await AddonBlog.addEntry({ subject, summary, summaryformat: 1, options });
 
         CoreEvents.trigger(ADDON_BLOG_ENTRY_UPDATED);
         this.forceLeave = true;
 
+        CoreForms.triggerFormSubmittedEvent(
+            this.formElement,
+            !!response || !!this.entry?.id,
```
dpalou, line 378: IMO this check isn't needed, you can always pass "true" because blog right now only works online. If you reach this point it means that the data was sent online.

src/addons/blog/pages/edit-entry/edit-entry.ts
```diff
@@ -133,9 +133,16 @@ export class AddonBlogEditEntryPage implements CanLeave, OnInit {
...
         const courseId = CoreNavigator.getRouteNumberParam('courseId');
+        const cmId = CoreNavigator.getRouteNumberParam('cmId');
         this.userId = CoreNavigator.getRouteNumberParam('userId');
         this.siteHomeId = CoreSites.getCurrentSiteHomeId();
 
+        if (cmId) {
+            this.modId = cmId;
+            this.form.controls.associateWithModule.setValue(true);
+            this.associatedModule = await CoreCourse.getModule(this.modId);
+        }
```
dpalou, line 144: This `await` should be surrounded by a try/catch, because the call could fail. I saw that there's also an "`await CoreCourseHelper.getCourse`" a few lines below this one, IMO you should put both awaits inside a try/catch like you do when you're editing an entry.

dpalou (changes requested):

dpalou (approved):

PR #4086, opened 2024-06-06
MOBILE-4470 workshop: Remove changes confirmation when not editing

crazyserver:


src/addons/mod/workshop/components/assessment-strategy/assessment-strategy.ts
```diff
@@ -273,7 +273,7 @@ export class AddonModWorkshopAssessmentStrategyComponent implements OnInit, OnDe
             return true;
         }
 
-        return AddonWorkshopAssessmentStrategyDelegate.hasDataChanged(
+        return this.edit && AddonWorkshopAssessmentStrategyDelegate.hasDataChanged(
```
dpalou, line 276: If we don't want to display the confirm when we aren't editing, wouldn't it be better to check "this.edit" at the beginning of the function?

dpalou (changes requested):

src/addons/mod/workshop/components/assessment-strategy/assessment-strategy.ts
crazyserver: Nice catch!

dpalou (approved):

PR #4087, opened 2024-06-06
MOBILE-4470 popover: Fix popover position when zoom applied

dpalou:


NoelDeMartin (commented):
I see that you've removed the previous fix, can you add the explanation to the patched files in Ionic? In particular, I think it's important that the patch itself documents what it's for in case it isn't immediately obvious. That way, when we stop using zoom in the future we'll know this can be removed.

NoelDeMartin (approved):

PR #4121, opened 2024-07-12
Mobile 4616 Solve lots of circular dependencies

crazyserver:


src/addons/mod/h5pactivity/constants.ts
```diff
@@ -0,0 +1,33 @@
...
+export const ADDON_MOD_H5PACTIVITY_COMPONENT = 'mmaModH5PActivity';
+
+export const ADDON_MOD_H5PACTIVITY_PAGE_NAME = 'mod_h5pactivity';
+
+export const ADDON_MOD_H5PACTIVITY_TRACK_COMPONENT = 'mod_h5pactivity'; // Component for tracking.
+export const ADDON_MOD_H5PACTIVITY_USERS_PER_PAGE = 20;
+
+// Grade type constants.
+export const enum AddonModH5PActivityGradeMethod {
+    GRADEMANUAL = 0, // No automathic grading using attempt results.
```
dpalou, line 24: Typo: automatic. It was already there, but we can fix it.

src/addons/mod/scorm/constants.ts
```diff
@@ -0,0 +1,57 @@
...
+    GRADEAVERAGE = 2,
+    GRADESUM = 3,
+}
+
+export const enum AddonModScormWhatGrade {
+    HIGHESTATTEMPT = 0,
+    AVERAGEATTEMPT = 1,
+    FIRSTATTEMPT = 2,
+    LASTATTEMPT = 3,
+}
```
dpalou, line 31: IIRC the first variable is about the grade in an attempt, while the second one is about the grade in the whole SCORM. Maybe we can name them ScormAttemptGradeMethod and ScormGradeMethod?

src/addons/mod/scorm/constants.ts
```diff
@@ -0,0 +1,57 @@
...
+
+export const enum AddonModScormMode {
+    MODEBROWSE = 'browse',
+    MODENORMAL = 'normal',
+    MODEREVIEW = 'review',
```
dpalou, line 36: Since the name of the enum already says "Mode" maybe we can just put the names without MODE? And the same happens in the next enums.

dpalou (changes requested):

dpalou (approved):

PR #4122, opened 2024-07-13
probably deployment-target should be also updated to 14.0

hn-88:
when target SDK has been upgraded to 34.


dpalou:
Hi,

the deployment-target affects the minimum iOS version where the app can be installed, it has nothing to do with the Android target SDK.

It was intended to set it to 13.0, so I'm closing this PR :)

Cheers,
Dani

hn-88:
Ah! iOS 13 and not Android 13! Thanks, @dpalou

PR #4123, opened 2024-07-15
Mobile 4632

crazyserver:


src/core/features/siteplugins/services/siteplugins-init.ts
```diff
@@ -101,52 +96,49 @@ const HANDLER_DISABLED = 'core_site_plugins_helper_handler_disabled';
...
         handlerName: string;
         handlerSchema: CoreSitePluginsCourseOptionHandlerData | CoreSitePluginsUserHandlerData;
         handler: CoreSitePluginsCourseOptionHandler | CoreSitePluginsUserProfileHandler;
     }> = {};
 
-    constructor() {
-        this.logger = CoreLogger.getInstance('CoreSitePluginsHelperProvider');
-    }
+    protected static readonly HANDLER_DISABLED = 'core_site_plugins_helper_handler_disabled';
+    protected static logger = CoreLogger.getInstance('CoreSitePluginsHelper');
```
dpalou, line 109: Shouldn't the name passed to CoreLogger be CoreSitePluginsInitializer?

src/core/features/compile/components/compile-html/compile-html.ts
```diff
@@ -163,6 +172,30 @@ export class CoreCompileHtmlComponent implements OnChanges, OnDestroy, DoCheck {
...
+        }
+
+        // Prepend all CSS rules with :host to avoid conflicts.
+        if (this.cssCode) {
+            if (!this.cssCode.includes(':host')) {
+                this.cssCode =  CoreDom.prefixCSS(this.cssCode, ':host ::ng-deep', ':host');
+            }
+
+        }
+    }
```
dpalou, line 197: This can be simplified to a single if.

src/core/features/siteplugins/services/siteplugins-init.ts
```diff
@@ -101,52 +96,49 @@ const HANDLER_DISABLED = 'core_site_plugins_helper_handler_disabled';
...
             } finally {
                 CoreSitePlugins.setPluginsFetched();
             }
         });
 
         // Re-load plugins restricted for courses when the list of user courses changes.
         CoreEvents.on(CoreCoursesProvider.EVENT_MY_COURSES_CHANGED, (data) => {
-            if (data && data.siteId && data.siteId == CoreSites.getCurrentSiteId() && data.added && data.added.length) {
-                this.reloadCourseRestrictHandlers();
+            if (data?.siteId === CoreSites.getCurrentSiteId() && data?.added.length) {
```
dpalou, line 140: Please check that siteId is not undefined, otherwise if the user logs out it could happen that the event passes an undefined siteId and getCurrentSiteId() returns undefined too.

src/core/features/siteplugins/services/siteplugins-init.ts
```diff
@@ -182,9 +174,9 @@ export class CoreSitePluginsHelperProvider {
         const uniqueName = CoreSitePlugins.getHandlerUniqueName(plugin, handlerName);
         const componentId = uniqueName + '#main';
 
-        // Remove the CSS files for this handler that aren't used anymore. Don't block the call for this.
+        // Remove the CSS files for this handler that aren't used anymore. Don't block the call for CoreSitePluginsInitializer.
```
dpalou, line 177: This automatic replace wasn't needed, "this" was the right word.

src/core/features/siteplugins/services/siteplugins-init.ts
```diff
@@ -182,9 +174,9 @@ export class CoreSitePluginsHelperProvider {
...
         const files = await CoreUtils.ignoreErrors(
-            CoreFilepool.getFilesByComponent(site.getId(), CoreSitePluginsProvider.COMPONENT, componentId),
+            CoreFilepool.getFilesByComponent(site.getId(), CORE_SITE_PLUGINS_UPDATE_COURSE_CONTENT, componentId),
```
dpalou, line 179: This should be CORE_SITE_PLUGINS_COMPONENT.

src/core/features/siteplugins/services/siteplugins-init.ts
```diff
@@ -209,7 +201,7 @@ export class CoreSitePluginsHelperProvider {
             site.getId(),
             url,
             false,
-            CoreSitePluginsProvider.COMPONENT,
+            CORE_SITE_PLUGINS_UPDATE_COURSE_CONTENT,
```
dpalou, line 204: As above, it should be CORE_SITE_PLUGINS_COMPONENT. You probably typed the wrong name in the replace. There are more, please review them.

src/core/features/siteplugins/services/siteplugins-init.ts
```diff
@@ -101,52 +96,49 @@ const HANDLER_DISABLED = 'core_site_plugins_helper_handler_disabled';
  * @todo Support ViewChild and similar in site plugins. Possible solution: make components and directives inject the instance
  * inside the host DOM element?
  */
-@Injectable({ providedIn: 'root' })
```
dpalou, line 104: What benefit do we have to make this class be a static class instead of an Angular service? IMO it makes the class itself a bit less readable, since you need to put CoreSitePluginsInitializer everywhere and this makes the lines longer (and in many places you had to split the code into several lines because of this). Also, if for some reason we need to override a certain function or create a testing stub we wouldn't be able to do it using Angular injection, we would need to use a different strategy.

dpalou (changes requested):

src/core/features/siteplugins/services/siteplugins-init.ts
crazyserver: Ooooos!

src/core/features/siteplugins/services/siteplugins-init.ts
crazyserver: ok!

dpalou (approved):

PR #4124, opened 2024-07-22
MOBILE-4630 pluginfile: Add lang to file download request

alfonso-salces:


src/core/services/utils/url.ts
```diff
@@ -211,7 +211,8 @@ export class CoreUrlUtilsProvider {
             url = this.addParamsToUrl(url, { token });
         }
 
-        return this.addParamsToUrl(url, { offline: '1' }); // Always send offline=1 (it's for external repositories).
+        // Always send offline=1 (it's for external repositories).
+        return this.addParamsToUrl(url, { offline: '1', lang: Translate.currentLang });
```
dpalou, line 215: The app uses "-" for child languages (e.g. en-us), while Moodle uses "_" (e.g. en_us). So I guess this won't work fine for child languages.

Also, we usually prefer using our `CoreLang` service instead of using `Translate` directly. The problem is that our `getCurrentLanguage` function is async. Maybe we could create a new function that is synchronous, something like `getCurrentLanguageInstant` or `getCurrentLanguageSync`, and this function wouldn't have the code to detect the current language (this means it can return undefined). Also, this function also converts the format to Moodle format if you indicate it using the parameter, so the problem with "_" and "-" would be easily fixed.

dpalou, line 215: One thing to take into account is that, if the user changes the language of the app, the downloaded files will still have the old language. One possible solution is to "invalidate" all files when the app language changes, but that means all the files will be re-downloaded when probably most files won't have multi lang. We should probably discuss this internally with Juan to decide what we want to do.

dpalou (changes requested):

src/core/services/utils/url.ts
alfonso-salces: As we dicussed internally with Juan, this is expected and does not require to take care of that right now.

dpalou (approved):

PR #4127, opened 2024-07-22
Mobile 4616

crazyserver:


src/core/classes/sites/authenticated-site.ts
```diff
@@ -1269,10 +1268,40 @@ export class CoreAuthenticatedSite extends CoreUnauthenticatedSite {
...
+
+            // Check is a valid number.
+            if (Number(version) >= 24) {
+                // Append release number.
+                docsUrl = docsUrl.replace('https://docs.moodle.org/', 'https://docs.moodle.org/' + version + '/');
+            }
+        }
+
+        try {
+            let lang = await CoreLang.getCurrentLanguage(CoreLangFormat.LMS);
```
dpalou, line 1298: Now that we have a synchronous function to get the current language maybe we should use it instead? This function should always be called after the app has been initialized.

Btw, I think this function isn't used in our public code (I haven't checked our private code).

dpalou (changes requested):

src/core/classes/sites/authenticated-site.ts
crazyserver: No it's isn't used.... Do you want me to remove it? It generates some circular deps, and that's why I moved it, but I can directly warn the user it will be removed

src/core/singletons/url.ts
```diff
@@ -344,4 +350,573 @@ export class CoreUrl {
...
+     * Remove the parameters from a URL, returning the URL without them.
+     *
+     * @param url URL to treat.
+     * @returns URL without params.
+     */
+    static removeUrlParams(url: string): string {
+        const matches = url.match(/^[^?]+/);
+
+        return matches ? matches[0] : '';
+    }
```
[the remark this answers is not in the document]
crazyserver: :-)

src/core/singletons/url.ts
```diff
@@ -101,7 +120,8 @@ export class CoreUrl {
 
         // Prepare parts replacing empty strings with undefined.
         return {
-            protocol: match[2] || undefined,
+            protocol: match[3]?.startsWith('//') ? match[2]: undefined || undefined,
+            scheme: match[2] || undefined,
```
dpalou, line 124: As we discussed, I think it's better to only treat protocol here. Also in assemble function.

src/core/singletons/url.ts
```diff
@@ -68,6 +79,13 @@ interface UrlParts {
 
 }
 
+export const enum CoreUrlPartNames {
+    Scheme = 'scheme',
```
dpalou, line 83: As we discussed, I guess it's better to name this part "protocol" instead of scheme.

src/core/singletons/url.ts
```diff
@@ -344,4 +366,572 @@ export class CoreUrl {
...
+     *
+     * @param url URL to treat.
+     * @returns Protocol, undefined if no protocol found.
+     * @todo Use CoreUrl.parse
+     */
+    static getUrlProtocol(url: string): string | void {
+        const urlParts = CoreUrl.parse(url);
+
+        return urlParts?.protocol;
+    }
```
dpalou, line 663: If we don't differentiate between protocol and scheme in `parse`, the behaviour of this function will change. Before, it only captured protocols followed by "://", now it will also capture "javascript:", "mailto:", etc. Maybe it's ok like this, but we need to be careful this change doesn't break anything.

If we force that `parse` only captures protocols with "://", then it will break the function `isLocalFileUrl`.

If this becomes too problematic maybe we can stop using parse in some of these functions for now.

src/core/singletons/url.ts
```diff
@@ -344,4 +366,572 @@ export class CoreUrl {
...
+     * E.g. javascript:doSomething() returns 'javascript'.
+     *
+     * @param url URL to treat.
+     * @returns Scheme, undefined if no scheme found.
+     */
+    static getUrlScheme(url: string): string | void {
+        const urlParts = CoreUrl.parse(url);
+
+        return urlParts?.scheme;
+    }
```
dpalou, line 676: As we discussed, maybe this code should be moved somewhere else, like CoreCustomURLSchemes because a string with a URL scheme might not be a valid URL.

src/core/singletons/url.ts
```diff
@@ -344,4 +366,572 @@ export class CoreUrl {
...
+        }
+
+        // Get URL without protocol.
+        const withoutProtocol = url.replace(/^[^?@/]*:\/\//, '');
+        const matches = withoutProtocol.match(/[^@]*/);
+
+        // Make sure that @ is at the start of the URL, not in a param at the end.
+        if (matches && matches.length && !matches[0].match(/[/|?]/)) {
+            return matches[0];
+        }
```
dpalou, line 697: This function can use `parse` too, right?

src/core/singletons/url.ts
```diff
@@ -344,4 +366,572 @@ export class CoreUrl {
...
+    }
+
+    /**
+     * Returns the URL without the desired parts.
+     *
+     * @param url URL to treat.
+     * @param parts Parts to remove.
+     * @returns URL without the parts.
+     */
+    static removeUrlParts(url: string, parts: CoreUrlPartNames[]): string {
```
dpalou, line 877: A tiny improvement for this function would be to accept a single part as a param too:

`parts: CoreUrlPartNames | CoreUrlPartNames[]`

That way you can do things like `removeUrlParts(url, CoreUrlPartNames.Scheme)` without having to pass an array. But it's a tiny improvement, feel free to leave it like this if you prefer.

src/core/singletons/url.ts
```diff
@@ -344,4 +366,572 @@ export class CoreUrl {
...
+        if (!url) {
+            return url;
+        }
+
+        parts.forEach((part) => {
+            switch (part) {
+                case CoreUrlPartNames.WWWInDomain:
+                    // Remove www.
+                    url = url.replace(/^www./, '');
```
dpalou, line 886: Sine this RegExp uses "^", the order of the part names is important. E.g. if I do:

[CoreUrlPartNames.WWWInDomain, CoreUrlPartNames.Scheme]

It won't work properly, because first it will look for www but it won't be found at the beginning because the URL still has the protocol. And it won't work either if the user only wants to remove "www". We should probably modify the regexp to search it after the protocol too.

dpalou (changes requested):

dpalou (approved):

dpalou (dismissed):

dpalou (approved):

PR #4133, opened 2024-07-29
Mobile 4587

dpalou:


src/addons/qtype/ddwtos/component/ddwtos.scss
```diff
@@ -47,19 +46,18 @@ core-format-text ::ng-deep, .drags ::ng-deep {
...
+        cursor: var(--cursor);
     }
     .draghome, .drag.unplaced{
         border: 1px solid var(--core-dd-question-border);
     }
     .draghome {
         visibility: hidden;
     }
-    .drag {
-        z-index: 2;
```
crazyserver, line 58: I would leave z-index in .drag to avoid problems with clicks (I'm not sure about visibility hidden behaviour). I think opacity: 0 is clickable but visibility hidden, not. But I would leave that here just in case.

crazyserver (changes requested):

crazyserver (approved):

PR #4134, opened 2024-07-29
Mobile 4616

crazyserver:


src/core/services/qrscan.ts
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /**
+     * Open a modal to scan a QR code.
+     *
+     * @param title Title of the modal. Defaults to "QR reader".
+     * @returns Promise resolved with the captured text or undefined if cancelled or error.
+     */
+    async scanQR(title?: string): Promise<string | undefined> {
+        return CoreViewer.scanQR(title);
```
dpalou, line 50: I'm not sure if CoreViewer is the best place to have this logic. The main goal of this function is to **Scan** a QR, not to **View** something. IMO it would be better to have all the logic encapsulated in this QRScan service.

dpalou (changes requested):

dpalou (approved):

PR #4136, opened 2024-07-31
Mobile 4612

dpalou:


src/core/utils/signals.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * Return an effect wrapper that can be used to create an effect with a certain injection context.
+ * Example:
+ *
+ * const effectWrapper = effectWithInjectionContext(injector);
+ *
+ * effectWrapper(() => {
+ *    // Your effect code here.
+ * });
```
alfonso-salces, line 35: Code samples in comments should be between ``` to be highlighted on function hover. This also applies to `modelWithInjectionContext` function.

alfonso-salces (approved):

PR #4138, opened 2024-08-01
MOBILE-4581 utils: Add lang to openInBrowser urls

alfonso-salces:


src/core/services/utils/utils.ts
```diff
@@ -1181,12 +1181,16 @@ export class CoreUtilsProvider {
...
-            type: CoreAnalyticsEventType.OPEN_LINK,
-            link: originaUrl,
-        });
+        const site = CoreSites.getCurrentSite();
+        const siteUrl = site?.getURL();
 
-        window.open(url, '_system');
+        CoreAnalytics.logEvent({ type: CoreAnalyticsEventType.OPEN_LINK, link: originaUrl });
+        window.open(
+            siteUrl && url.includes(siteUrl)
```
dpalou, line 1189:
```suggestion
            site.containsUrl(url)
```

src/core/services/utils/utils.ts
```diff
@@ -1181,12 +1181,16 @@ export class CoreUtilsProvider {
             }
         }
 
-        CoreAnalytics.logEvent({
-            type: CoreAnalyticsEventType.OPEN_LINK,
-            link: originaUrl,
-        });
+        const site = CoreSites.getCurrentSite();
```
dpalou, line 1184: Now that we're looking at circular dependencies, this is adding more coupling between CoreUtilsProvider and CoreSites. We're already using CoreSites in another function in this file so this isn't a new circular dependency, I guess we can look at it later because it's not a trivial fix.

dpalou (changes requested):

src/core/services/utils/utils.ts
```diff
@@ -1181,12 +1181,14 @@ export class CoreUtilsProvider {
...
-            type: CoreAnalyticsEventType.OPEN_LINK,
-            link: originaUrl,
-        });
-
-        window.open(url, '_system');
+        const site = CoreSites.getCurrentSite();
+        CoreAnalytics.logEvent({ type: CoreAnalyticsEventType.OPEN_LINK, link: originaUrl });
+        window.open(
+            site?.containsUrl(url)
+                ? CoreUrl.addParamsToUrl(url, { lang: CoreLang.getCurrentLanguage(CoreLangFormat.LMS) })
```
dpalou, line 1188: `CoreLang.getCurrentLanguage` return a Promise, you need an `await` here.

dpalou (changes requested):

dpalou (approved):

PR #4141, opened 2024-08-05
MOBILE-4631 login: Hide the login form in the app when is hidden in LMS

albertgasset:


src/core/classes/sites/unauthenticated-site.ts
```diff
@@ -464,6 +464,7 @@ export type CoreSitePublicConfigResponse = {
     tool_mobile_setuplink?: string; // App download page.
     tool_mobile_qrcodetype?: CoreSiteQRCodeType; // eslint-disable-line @typescript-eslint/naming-convention
     warnings?: CoreWSExternalWarning[];
+    showloginform?: number; // Display default login form.
```
dpalou, line 467:
```suggestion
    showloginform?: number; // @since 4.5.Display default login form. 
```

We usually indicate the LMS version where this value was introduced.

src/core/features/login/pages/credentials/credentials.ts
```diff
@@ -137,6 +143,10 @@ export class CoreLoginCredentialsPage implements OnInit, OnDestroy {
                 }
             });
         }
+
+        this.alwaysShowLoginFormObserver = CoreEvents.on<number>(ALWAYS_SHOW_LOGIN_FORM_CHANGED, async () => {
```
dpalou, line 147:
```suggestion
        this.alwaysShowLoginFormObserver = CoreEvents.on(ALWAYS_SHOW_LOGIN_FORM_CHANGED, async () => {
```

To define the parameters of an event we do it differently so it's detected automatically without having to specify the type in every listener. You should add the type in here:

https://github.com/moodlehq/moodleapp/blob/main/src/core/features/login/services/login-helper.ts#L1695

In this code we're basically adding new values to the `CoreEventsData` interface that is used by `CoreEvents`. The keys of the object are the names of the events, and the value is the type of the parameter passed to the event. If you add your event in here, all the listeners will automatically infer the type without having to specify it manually. So you can remove the `<number>` type from here and from reconnect page.

PD: Please notice the type of the parameter you're passing is `{ value: number }`, not just `number`.

dpalou (changes requested):
The format of commit names is "ISSUE component: Description". Your commit lacks a component, e.g. "login".

albertgasset:
Thanks @dpalou  for the review, I've updated the branch with the changes.

dpalou (approved):

PR #4142, opened 2024-08-06
MOBILE-4580 courses: Use new WS parameter to fetch course by section id

albertgasset:


src/core/features/courses/services/handlers/section-link.ts
```diff
@@ -66,12 +66,18 @@ export class CoreCoursesSectionLinkHandlerService extends CoreCoursesLinksHandle
...
+    private async getSectionCourse(sectionId: number | false, siteId: string | false): Promise<CoreCourseBasicData | null> {
         if (!siteId || !sectionId) {
             return null;
         }
 
-        // Ideally, we would use a webservice to get this information; but such webservice doesn't exists.
+        const site = await CoreSites.getSite(siteId);
+
+        if (site.isVersionGreaterEqualThan('4.5')) {
+            return CoreCourses.getCourseByField('sectionid', sectionId, siteId);
```
dpalou, line 77: If this WS call fails (e.g. the user is offline) we should try to use the fallback that we use for 4.4 sites.

dpalou (changes requested):

dpalou (approved):

PR #4147, opened 2024-08-13
MOBILE-4070 attachments: Collapse mimetypes

alfonso-salces:


src/core/components/attachments/core-attachments.html
```diff
@@ -13,16 +13,31 @@
                 <ion-icon name="fas-plus" slot="icon-only" aria-hidden="true" />
             </ion-button>
         </ion-item>
-        <ion-item class="ion-text-wrap" *ngIf="fileTypes && fileTypes.mimetypes && fileTypes.mimetypes.length">
```
crazyserver, line 16: Lately we're using ion-accordion to solve this behaviour because it's easier to implement. Have you checked that?

crazyserver (changes requested):

src/core/components/attachments/core-attachments.html
```diff
@@ -13,16 +13,30 @@
...
-                </ul>
-            </ion-label>
-        </ion-item>
+
+        @if (fileTypes && fileTypes.mimetypes && fileTypes.mimetypes.length) {
+        <ion-accordion-group>
+            <ion-accordion toggleIconSlot="start">
+                <ion-item class="core-attachments-header ion-text-wrap" aria-controls="core-attachments" slot="header">
+                    <p>{{ 'core.fileuploader.filesofthesetypes' | translate }}</p>
+                    <ion-icon name="fas-circle-info" aria-hidden="true" slot="end" flip-rtl />
```
crazyserver, line 22: flip-rtl is not needed here because icon is not directional (arrows, etc)

src/core/components/attachments/core-attachments.html
```diff
@@ -13,16 +13,30 @@
...
-                    </li>
-                </ul>
-            </ion-label>
-        </ion-item>
+
+        @if (fileTypes && fileTypes.mimetypes && fileTypes.mimetypes.length) {
+        <ion-accordion-group>
+            <ion-accordion toggleIconSlot="start">
+                <ion-item class="core-attachments-header ion-text-wrap" aria-controls="core-attachments" slot="header">
+                    <p>{{ 'core.fileuploader.filesofthesetypes' | translate }}</p>
```
crazyserver, line 21: This item does not have ion-label, also, p tag will need item-heading class to match  styles after addind ion-label

src/core/components/attachments/core-attachments.html
```diff
@@ -13,16 +13,30 @@
...
+                    <ion-icon name="fas-circle-info" aria-hidden="true" slot="end" flip-rtl />
+                </ion-item>
+
+                <div class="core-attachments" slot="content">
+                    <ion-label>
```
crazyserver, line 26: This ion-label is not needed because you're outsite the ion-item

src/core/components/attachments/attachments.scss
```diff
@@ -1,3 +1,19 @@
...
-    --loading-inline-min-height: 60px;
+@use "theme/globals" as *;
+
+:host {
+    core-loading {
+        --loading-inline-min-height: 60px;
+    }
+
+    .core-attachments {
+        @include margin-horizontal(null, .8rem);
```
crazyserver, line 9: Margins on design system are on a 4x pixel basis. I think it would be better to use 12px or var(--mdl-spacing-3);

src/core/components/attachments/attachments.scss
```diff
@@ -1,3 +1,19 @@
...
+
+        &-header {
+            ion-icon {
+                @include margin-horizontal(null, .8rem);
```
crazyserver, line 13: Same margin as above

src/core/components/attachments/attachments.scss
```diff
@@ -1,3 +1,19 @@
...
+                color: var(--ion-color-tint);
```
crazyserver, line 14: I preffer using --text-color here, tint it could be confusing and I think it is not defined. You can also try using color="dark" on the html tag and remove this line. Check dark mode as well.

src/core/components/attachments/core-attachments.html
```diff
@@ -13,16 +13,30 @@
...
+
+        @if (fileTypes && fileTypes.mimetypes && fileTypes.mimetypes.length) {
+        <ion-accordion-group>
+            <ion-accordion toggleIconSlot="start">
+                <ion-item class="core-attachments-header ion-text-wrap" aria-controls="core-attachments" slot="header">
+                    <p>{{ 'core.fileuploader.filesofthesetypes' | translate }}</p>
+                    <ion-icon name="fas-circle-info" aria-hidden="true" slot="end" flip-rtl />
+                </ion-item>
+
+                <div class="core-attachments" slot="content">
```
crazyserver, line 25: IMO div is not needed you can add the class and slot to the ul tag directly.

src/core/components/attachments/core-attachments.html
```diff
@@ -13,16 +13,30 @@
...
+                    <ion-label>
+                        <ul class="list-with-style">
```
crazyserver, line 27: I think this class list-with-style is not defined. I would remove it.

crazyserver (changes requested):
Thanks Alfonso for taking care of this issue. I did some comments. Some of them are not mandatory changes since they are based on my opinion and they are all easy to fix. Can you review?

crazyserver (approved):

PR #4159, opened 2024-08-26
MOBILE-4574 badges: Support links to badges by hash

albertgasset:


src/addons/badges/services/badges.ts
```diff
@@ -106,6 +106,58 @@ export class AddonBadgesProvider {
...
+    }
+
+    /**
+     * Invalidate get badge by hash WS call.
+     *
+     * @param hash Badge issued hash.
+     * @param siteId Site ID. If not defined, current site.
+     * @returns Promise resolved when data is invalidated.
+     */
+    async invalidateUserBadgeByHash(hash: string, siteId?: string): Promise<void> {
```
dpalou, line 155: This function isn't used, you should use it in `refreshBadges` if needed.

src/addons/badges/services/badges.ts
```diff
@@ -167,6 +219,8 @@ export type AddonBadgesUserBadge = {
     dateissued: number; // Date issued.
     dateexpire: number; // Date expire.
     visible?: number; // Visible.
+    recipientid?: number; // @sice 4.5. Id of the awarded user.
+    recipientfullname?: string; // @sice 4.5. Full name of the awarded user.
```
dpalou, line 223: Typo: sice instead of since.

src/addons/badges/badge-lazy.module.ts
```diff
@@ -0,0 +1,32 @@
...
+import { NgModule } from '@angular/core';
+import { RouterModule, Routes } from '@angular/router';
+
+import { AddonBadgesIssuedBadgePage } from './pages/issued-badge/issued-badge';
+
+const routes: Routes = [
+    {
+        path: ':badgeHash',
+        component: AddonBadgesIssuedBadgePage,
+    },
```
dpalou, line 24: If this page is opened without split view, then the following code shouldn't be executed in the page:

```
const source = CoreRoutedItemsManagerSourcesTracker.getOrCreateSource(
    AddonBadgesUserBadgesSource,
    [this.courseId, this.userId],
);

this.badges = new CoreSwipeNavigationItemsManager(source);
```
This code is used to allow changing the current badge by swiping left/right when there is a spit view. Since there is no split view in this case, the code shouldn't be used.

I guess the best way to detect this is to use the "data" property of the route to identify it has split view or not, something like:

```suggestion
    {
        path: ':badgeHash',
        component: AddonBadgesIssuedBadgePage,
        data: { usesSplitView: false },
    },
```

Please notice you will need to add "usesSplitView: true" in the routes that use split view. 

If you search "discussionsPathPrefix" in the app you will see an example of how to use this route data.

dpalou (changes requested):

src/addons/badges/services/badges-helper.ts
```diff
@@ -0,0 +1,55 @@
...
+        if (site.isVersionGreaterEqualThan('4.5')) {
+            // The WS to fetch a badge by hash is available and it returns the name of the recipient.
+            return true;
+        }
+
+        // Open in browser if badge is one of the user badges.
+        const badges = await AddonBadges.getUserBadges(0, site.getUserId());
+        const badge = badges.find((badge) => badgeHash == badge.uniquehash);
+
+        return badge != null;
```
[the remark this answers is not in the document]
albertgasset: Ok, but it should be` !== undefined`

src/addons/badges/badge-lazy.module.ts
albertgasset: I've used the name `usesSwipeNavigation", because swipe navigation it's also used without the split view.

src/addons/badges/services/badges.ts
albertgasset: Now it's called in `refreshBadges`

src/addons/badges/pages/issued-badge/issued-badge.ts
```diff
@@ -96,16 +100,29 @@ export class AddonBadgesIssuedBadgePage implements OnInit, OnDestroy {
...
+                    badge.recipientfullname = user.fullname;
+                }
+            } else {
+                // The badge is awarded to another user, try to fetch the badge by hash.
+                if (site.isVersionGreaterEqualThan('4.5')) {
+                    badge = await AddonBadges.getUserBadgeByHash(this.badgeHash);
+                }
+                if (!badge) {
+                    // Should never happen. The app opens the badge in the browser if it can't be fetched.
+                    throw '';
```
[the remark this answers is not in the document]
albertgasset: Fixed

src/addons/badges/services/badges-helper.ts
```diff
@@ -0,0 +1,55 @@
...
+        }
+
+        const site = await CoreSites.getSite(siteId);
+
+        if (site.isVersionGreaterEqualThan('4.5')) {
+            // The WS to fetch a badge by hash is available and it returns the name of the recipient.
+            return true;
+        }
+
+        // Open in browser if badge is one of the user badges.
```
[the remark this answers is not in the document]
albertgasset: Fixed

src/addons/badges/services/badges.ts
albertgasset: Fixed

dpalou (approved):

PR #4165, opened 2024-09-02
MOBILE-2256 privatefiles: Remove private files

alfonso-salces:


src/addons/privatefiles/pages/index/index.ts
```diff
@@ -279,4 +291,180 @@ export class AddonPrivateFilesIndexPage implements OnInit, OnDestroy {
...
+        try {
+            await AddonPrivateFiles.deleteFile(file);
+            await this.refreshFiles();
+        } catch (error) {
+            await CoreDomUtils.showErrorModalDefault(error, 'An error occourred while file was being deleted.');
+        }
+
+        if (loading) {
+            loading.dismiss();
+        }
```
dpalou, line 316:
```suggestion
        loading?.dismiss();
```

src/addons/privatefiles/components/file-actions.html
```diff
@@ -0,0 +1,29 @@
...
+    <ion-item button="false" detail="false" lines="none">
+        <ion-icon slot="start" name="fas-cloud" aria-hidden="true" />
+        <ion-label>{{ 'addon.privatefiles.availableoffline' | translate }}</ion-label>
+        <ion-toggle slot="end" [(ngModel)]="toggleValue" (ngModelChange)="close({ status: toggleValue ? 'download' : 'deleteOffline' })" />
+    </ion-item>
+
+    <ion-item detail="false" (click)="close({ status: 'deleteOnline' })" lines="none">
+        <ion-icon slot="start" name="fas-trash" aria-hidden="true" color="danger" />
+        <ion-label> {{ 'core.delete' | translate }} </ion-label>
+    </ion-item>
```
dpalou, line 28: This modal doesn't allow re-downloading files that are outdated (downloaded but there is a newer version). We'll talk with UX to find a way to support it.

dpalou (changes requested):

src/addons/privatefiles/components/file-actions/file-actions.html
```diff
@@ -0,0 +1,30 @@
+<header>
```
dpalou, line 1: While testing this I noticed that clicking the backdrop doesn't close this modal. Is that intended? IMO it's a bit weird from the user's POV. The other modal we have with `openSheet` (CoreFileUploaderAudioRecorderComponent) I guess it makes sense not to close it because the user is doing things that last a while (record audio), but in this case we're only displaying some options so IMO the expected behaviour is to close the modal when clicking the backdrop.

src/addons/privatefiles/components/file-actions/file-actions.html
```diff
@@ -0,0 +1,30 @@
...
+
+<ion-list>
+    @if (isDownloaded) {
+    <ion-item button="false" detail="false" lines="none" (click)="close({ status: 'deleteOffline' })">
+        <ion-icon slot="start" name="fam-cloud-x" aria-hidden="true" />
+        <ion-label>{{ 'core.removedownloadeddata' | translate }}</ion-label>
+    </ion-item>
+    }
+
+    <ion-item detail="false" (click)="close({ status: 'deleteOnline' })" lines="none">
```
dpalou, line 26:
```suggestion
    <ion-item button detail="false" lines="none" (click)="close({ status: 'deleteOffline' })">
        <ion-icon slot="start" name="fam-cloud-x" aria-hidden="true" />
        <ion-label>{{ 'core.removedownloadeddata' | translate }}</ion-label>
    </ion-item>
    }

    <ion-item button detail="false" (click)="close({ status: 'deleteOnline' })" lines="none">
```

IMO we should use "button" in these items, since they're clickable. E.g. this makes the webapp change the cursor when hovering, and maybe it has other effects too.

src/addons/privatefiles/components/file/file.html
```diff
@@ -0,0 +1,51 @@
...
+        <ion-thumbnail slot="start" (click)="download($event, true)">
+            <img [src]="fileIcon" alt="" role="presentation" />
+        </ion-thumbnail>
+        }
+
+        <ion-label (click)="download($event, true)">
+            <p class="item-heading">
+                {{fileName}}
+
+                @if (state === 'downloaded') {
```
dpalou, line 18: In core-download-refresh component, we set the constant value to a variable to avoid hardcoding the value in the template:
```
    TS:
    statusDownloaded = DownloadStatus.DOWNLOADED;
    HTML:
    *ngIf="status === statusDownloaded
```

src/addons/privatefiles/components/file/file.html
```diff
@@ -0,0 +1,51 @@
...
+            <ion-button fill="clear" *ngIf="isDownloaded && isIOS" (click)="openFile($event, true)" [title]="openButtonLabel | translate">
+                <ion-icon slot="icon-only" [name]="openButtonIcon" aria-hidden="true" />
+            </ion-button>
+
+            @if (!showCheckbox) {
+
+            @if (!isDownloaded) {
+            <core-download-refresh [status]="state" [enabled]="canDownload" [loading]="isDownloading" [canTrustDownload]="!alwaysDownload"
+                (action)="download()" />
+            }
```
dpalou, line 41: Please notice that `isDownloaded` is also true for outdated files, because the file is downloaded (but outdated). This means that this button won't be displayed for outdated files, and that isn't what we want. You should check that `state` is different than `DownloadStatus.DOWNLOADED`.

src/addons/privatefiles/pages/index/index.ts
```diff
@@ -279,4 +289,156 @@ export class AddonPrivateFilesIndexPage implements OnInit, OnDestroy {
...
+            try {
+                await CoreDomUtils.showDeleteConfirm('core.confirmremoveselectedfiles');
+            } catch {
+                return;
+            }
+        }
+
+        const siteId = CoreSites.getCurrentSiteId();
+        const loading = await CoreLoadings.show();
+        await AddonPrivateFiles.deleteFiles(this.selectedFiles);
```
dpalou, line 306: This WS call can fail, but the error isn't captured so the user won't know what happened.

Please notice that, if this function fails, the 'deleteOnline' case from the management menu shouldn't display the toast saying that the file was deleted.

src/addons/privatefiles/pages/index/index.ts
```diff
@@ -279,4 +289,156 @@ export class AddonPrivateFilesIndexPage implements OnInit, OnDestroy {
...
+
+        for (const file of this.selectedFiles) {
+            try {
+                await this.deleteOfflineFile(file, siteId);
+            } catch (error) {
+                await CoreDomUtils.showErrorModalDefault(error, 'An error occourred while file was being deleted.');
```
dpalou, line 312: The function deleteOfflineFile already captures errors and displays them to the user, so there's no need to do it here too.

src/addons/privatefiles/pages/index/index.ts
```diff
@@ -279,4 +289,156 @@ export class AddonPrivateFilesIndexPage implements OnInit, OnDestroy {
...
+        }
+
+        if (status === 'deleteOnline') {
+            try {
+                await CoreDomUtils.showDeleteConfirm('core.confirmremoveselectedfile', { filename: file.filename });
+            } catch {
+                return;
+            }
+
+            const loading = await CoreLoadings.show();
```
dpalou, line 386: The `deleteSelectedFiles` function already displays a loading so I guess you can remove this one.

src/addons/privatefiles/pages/index/index.ts
```diff
@@ -279,4 +289,156 @@ export class AddonPrivateFilesIndexPage implements OnInit, OnDestroy {
...
+            }
+        }
+
+        const siteId = CoreSites.getCurrentSiteId();
+        const loading = await CoreLoadings.show();
+        await AddonPrivateFiles.deleteFiles(this.selectedFiles);
+
+        for (const file of this.selectedFiles) {
+            try {
+                await this.deleteOfflineFile(file, siteId);
```
dpalou, line 310: When testing this in the webapp, I noticed that calling `deleteOfflineFile` throws an error in webapp if the file isn't downloaded, but it works fine in real devices. This is controlled [here](https://github.com/moodlehq/moodleapp/blob/main/src/core/services/filepool.ts#L2834), but I guess that in webapp we aren't throwing the right error.

IMO it would be nice to improve [our removeFile implementation for webapp](https://github.com/moodlehq/moodleapp/blob/main/src/core/features/emulator/services/file.ts#L685) to throw an error with code = 1 too if the file doesn't exist, to make it consistent with real devices. Another approach would be to only call deleteOfflineFile if we know that the file is downloaded, but that means having to add code here to check the status of the file.

The exact error thrown in real device is this, we should throw the same in webapp when the file doesn't exist:

{code: 1, message: 'NOT_FOUND_ERR'}

src/addons/privatefiles/pages/index/index.ts
```diff
@@ -279,4 +289,156 @@ export class AddonPrivateFilesIndexPage implements OnInit, OnDestroy {
...
+
+    /**
+     * Open file management menu.
+     *
+     * @param instance CoreFileComponent instance.
+     * @param file File to manage.
+     *
+     * @returns Promise done.
+     */
+    async openManagementFileMenu(instance: CoreFileComponent, file: AddonPrivateFilesFile): Promise<void> {
```
dpalou, line 365:
```suggestion
     * @param instance AddonPrivateFilesFileComponent instance.
     * @param file File to manage.
     *
     * @returns Promise done.
     */
    async openManagementFileMenu(instance: AddonPrivateFilesFileComponent, file: AddonPrivateFilesFile): Promise<void> {
```

src/addons/privatefiles/pages/index/index.ts
```diff
@@ -279,4 +289,156 @@ export class AddonPrivateFilesIndexPage implements OnInit, OnDestroy {
...
+        const siteId = CoreSites.getCurrentSiteId();
+        const fileState = await CoreFilepool.getFileStateByUrl(siteId, file.fileurl, file.timemodified);
+        const isFileDownloaded = CoreFileHelper.isStateDownloaded(fileState);
```
dpalou, line 368: Since now you receive the instance I guess you can use `instance.isDownloaded` instead of recalculating it.

src/addons/privatefiles/pages/index/index.ts
```diff
@@ -279,4 +289,156 @@ export class AddonPrivateFilesIndexPage implements OnInit, OnDestroy {
...
+                await CoreDomUtils.showDeleteConfirm('core.confirmremoveselectedfile', { filename: file.filename });
+            } catch {
+                return;
+            }
+
+            const loading = await CoreLoadings.show();
+            this.selectedFiles = [file];
+            await this.deleteSelectedFiles();
+            const message = Translate.instant('core.filedeletedsuccessfully', { filename: file.filename });
+            await CoreToasts.show({ message, translateMessage: false, duration: ToastDuration.SHORT });
```
dpalou, line 390: The `deleteSelectedFiles` function already displays a Toast, you're displaying it twice.

src/addons/privatefiles/services/privatefiles.ts
```diff
@@ -92,7 +93,7 @@ export class AddonPrivateFilesProvider {
             return [];
         }
 
-        return result.files.map((entry) => {
+        return await Promise.all(result.files.map(async (entry) => {
```
dpalou, line 96: This change is no longer needed.

dpalou (changes requested):

src/core/services/modals.ts
```diff
@@ -67,15 +67,22 @@ export class CoreModalsService {
...
+     * @param componentProps Component to render inside the modal.
+     * @param backdropDismiss Dismiss on backdrop click.
+     *
      * @returns Modal result once it's been closed.
      */
-    async openSheet<T>(component: Constructor<CoreModalComponent<T>>): Promise<T> {
+    async openSheet<T>(
+        component: Constructor<CoreModalComponent<T>>,
+        componentProps: Record<string, unknown> = {},
+        backdropDismiss = true,
```
dpalou, line 78: IMO you should pass `backdropDismiss=false` [in here](https://github.com/moodlehq/moodleapp/blob/main/src/core/features/fileuploader/services/fileuploader.ts#L164), because we don't want to change an existing feature.

src/core/services/modals.ts
```diff
@@ -85,6 +92,11 @@ export class CoreModalsService {
 
         viewContainer?.setAttribute('aria-hidden', 'true');
 
+        if (backdropDismiss) {
+            const backdrop = document.getElementsByTagName('ion-backdrop')?.[0];
```
dpalou, line 96:
```suggestion
            const backdrop = element.querySelector('ion-backdrop');
```
It's better to search inside `element` to find the proper backdrop, since the app could be displaying more than one.

Also, if you use querySelector you can get rid of the "?.[0]" part.

dpalou (changes requested):

dpalou (approved):

PR #4171, opened 2024-09-06
MOBILE-4547 blog: Fix remove offline entries sync

alfonso-salces:


src/addons/blog/services/blog-sync.ts
```diff
@@ -95,8 +95,8 @@ import { AddonBlogOfflineEntryDBRecord } from './database/blog';
      * @returns Syncronization result.
      */
     async performEntriesSync(siteId: string): Promise<AddonBlogSyncResult> {
-        const result: AddonBlogSyncResult = { updated: false, warnings: [] };
         const entriesToSync = await this.syncEntriesToRemove(siteId);
+        const result: AddonBlogSyncResult = entriesToSync.result;
```
dpalou, line 99:
```suggestion
        const result = entriesToSync.result;
```
This cast is no longer needed, now the type can be inferred. You can also do this:

`        const { entries, result } = await this.syncEntriesToRemove(siteId);`

dpalou (changes requested):

src/addons/blog/pages/index/index.ts
```diff
@@ -417,6 +417,12 @@ export class AddonBlogIndexPage implements OnInit, OnDestroy {
      * @param entryToRemove Entry.
      */
     async deleteEntry(entryToRemove: AddonBlogOfflinePostFormatted | AddonBlogPostFormatted): Promise<void> {
+        try {
+            await CoreDomUtils.showDeleteConfirm('addon.mod_data.confirmdeleterecord');
```
dpalou, line 421: I just noticed that you're using a mod_data string, and you shouldn't use strings from other addons. I missed this in my previous review.

I saw that LMS uses this string:

`$string['blogdeleteconfirm'] = 'Delete the blog entry \'{$a}\'?';`

You should add this string in the app and use it here.

dpalou (changes requested):

dpalou (approved):

PR #4180, opened 2024-09-17
Mobile 4442

crazyserver:


src/core/features/course/components/course-format/course-format.ts
```diff
@@ -326,6 +326,25 @@ export class CoreCourseFormatComponent implements OnInit, OnChanges, OnDestroy {
             this.loaded = true;
             this.sectionChanged(sections[0]);
         } else if (this.initialSectionId || this.initialSectionNumber !== undefined) {
+            const subSection = this.subSections.find((section) => section.id === this.initialSectionId);
```
dpalou, line 329: Shouldn't you also check `initialSectionNumber`? Or subsection don't have a number?

src/core/features/course/components/course-section/course-section.html
```diff
@@ -36,13 +36,69 @@ <h2 *ngIf="section.name" class="big" [id]="'core-section-name-' + section.id">
...
+            [viewedModules]="viewedModules" [collapsible]="true" />
+        } @else {
+        <core-course-module *ngIf="module.visibleoncoursepage !== 0" [module]="module" [section]="section"
+            [showActivityDates]="course.showactivitydates" [showCompletionConditions]="course.showcompletionconditions"
+            [isLastViewed]="lastModuleViewed && lastModuleViewed.cmId === module.id" [class.core-course-module-not-viewed]="
+                    !viewedModules[module.id] &&
+                    (!module.completiondata || module.completiondata.state === completionStatusIncomplete)" />
+        }
+    </ng-container>
+</div>
```
dpalou, line 104: Most of this HTML is the same as above, I think the only differences are:

slot="header"
<div slot="content">
<ng-container *ngIf="section.expanded">

IMO it would be better to move all the common code to ng-templates to reuse it. And that way it's easier to view that the content is the same in both cases.

src/core/features/course/components/course-section/course-section.ts
```diff
@@ -41,10 +44,11 @@ import { CoreCourseFormatDelegate } from '@features/course/services/format-deleg
         CoreCourseComponentsModule,
     ],
 })
-export class CoreCourseSectionComponent implements OnInit {
+export class CoreCourseSectionComponent implements OnInit, OnChanges {
 
     @Input({ required: true }) course!: CoreCourseAnyCourseData; // The course to render.
     @Input({ required: true }) section!: CoreCourseSectionToDisplay;
+    @Input() sections: CoreCourseSectionToDisplay[] = []; // List of sections in the course.
```
dpalou, line 51: This comment says "List of sections in the course.". But then you either pass a list of subsections, or nothing at all. This is confusing.

src/core/features/course/services/course.ts
```diff
@@ -1770,6 +1770,8 @@ type CoreCourseGetContentsWSSection = {
     uservisible?: boolean; // Is the section visible for the user?.
     availabilityinfo?: string; // Availability information.
     modules: CoreCourseGetContentsWSModule[]; // List of module.
+    component?: string; // @since 4.5 The delegate component of this section if any.
+    itemid?: number; // @since 4.5 The optional item id delegate component can use to identify its instance.
```
dpalou, line 1774: I put the comment here, but it's not related to this line of code. This commit has a submessage that only contains "a".

src/core/features/course/components/course-index/course-index.ts
```diff
@@ -187,6 +178,115 @@ export class CoreCourseCourseIndexComponent implements OnInit {
...
+                            id: module.id,
+                            name: module.name,
+                            course: module.course,
+                            visible: !!module.visible,
+                            uservisible: CoreCourseHelper.canUserViewModule(module, section),
+                            indented: true,
+                            subSection: subSectionToRender,
+                        };
+                    }
+                }
```
dpalou, line 216: If the subsection isn't found for some reason, shouldn't you also "return"? Otherwise you'll try to get the completion status when you shouldn't.

src/core/features/course/components/course-index/course-index.ts
```diff
@@ -187,6 +178,115 @@ export class CoreCourseCourseIndexComponent implements OnInit {
...
+            modules,
+        };
+    }
+
+    /**
+     * Toogle the visibility of a section (expand/collapse).
+     *
+     * @param ev The event of the accordion.
+     */
+    accordionMultipleChange(ev: AccordionGroupChangeEventDetail): void {
```
dpalou, line 251: Is this function used? You aren't using ion-accordion in course-index.

src/core/features/course/components/course-format/course-format.ts
```diff
@@ -487,19 +487,30 @@ export class CoreCourseFormatComponent implements OnInit, OnChanges, OnDestroy {
...
         this.sectionChanged(section);
 
+        if (data.subSectionId) {
+            section = this.subSections.find((section) => section.id === data.subSectionId);
+            if (!section) {
+                return;
+            }
+
+            // Use this section to find the module.
+            this.setSectionExpanded(section);
```
dpalou, line 511: If the user has selected a subsection but not a module, shouldn't we scroll to the subsection? Otherwise the user might not see it.

dpalou (changes requested):

src/core/features/course/components/course-index/course-index.ts
crazyserver: True, I tried to add accordion on course index but I had a limitation when clicking on the icon to expand collapse (i took the ion-item action instead). Now it has been removed.

src/core/features/course/components/course-format/course-format.ts
crazyserver: True! thx!

src/core/features/course/components/course-section/course-section.ts
crazyserver: changed!

src/core/features/course/components/course-index/course-index.ts
crazyserver: I'd return but without the subsection. thx

src/addons/mod/subsection/services/handlers/index-link.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    /**
+     * @inheritdoc
+     */
+    getActions(
+        siteIds: string[],
+        url: string,
+        params: Record<string, string>,
+        courseId?: number,
+    ): CoreContentLinksAction[] | Promise<CoreContentLinksAction[]> {        return [{
```
dpalou, line 44: Wrong indentation (the return).

src/addons/mod/subsection/services/subsection.ts
```diff
@@ -0,0 +1,58 @@
...
+
+}
+export const AddonModSubsection = makeSingleton(AddonModSubsectionProvider);
+
+/**
+ * Custom Data for subsection modules.
+ */
+export type AddonModSubsectionCustomData = {
+    sectionid: number;
+};
```
dpalou, line 58: I think this type isn't used.

src/core/features/course/components/course-format/course-format.ts
```diff
@@ -581,16 +614,14 @@ export class CoreCourseFormatComponent implements OnInit, OnChanges, OnDestroy {
     }
 
     /**
-     * Scroll to a certain module.
+     * Scroll to a certain module or section.
      *
-     * @param moduleId Module ID.
+     * @param id ID of the module or section to scroll to.
+     * @param isSubsection Whether the module is inside a subsection.
```
dpalou, line 620: IMO this JSDoc/parameter names is wrong.

The first line says "or section", but the parameter is called "isSubsection". The code actually allows scrolling to a section, so I guess the parameter should be named "isSection".

Also, the JSDoc of the parameter says "Whether the module is inside a subsection.", but it's actually used is to determine if the ID is a section or a module.

src/core/features/course/components/course-index/course-index.scss
```diff
@@ -87,3 +87,7 @@ ion-item.item {
         }
     }
 }
+
+div.core-course-index-subsection{
```
dpalou, line 91:
```suggestion
div.core-course-index-subsection {
```

src/core/features/course/components/course-section/course-section.ts
```diff
@@ -0,0 +1,99 @@
...
+                        section.component === 'mod_subsection' && section.itemid === module.instance);
+                }
+            });
+        }
+    }
+
+}
+
+type CoreCourseModuleToDisplay = CoreCourseModuleData & {
+    subsection?: CoreCourseSectionToDisplay;
```
dpalou, line 93: In course index, you name the module property "subSection", but here you name it "subsection" in lowercase. IMO we should be consistent.

I was concerned that this could cause problems if we used the same "module" object in both places. But I see that in course index we create a copy of the module, so as far as JS is concerned I think it should be ok. Maybe TS complains because the types don't match and you need to do an `Omit` in course index.

dpalou (changes requested):

src/core/features/course/components/course-section/course-section.ts
crazyserver: OK, done, thanks. About your concern, TS is not complaining so I left it like it is now

dpalou (dismissed):

dpalou (approved):

PR #4181, opened 2024-09-17
MOBILE-4639 badges: Support links to badges/badgeclass.php?id=X

albertgasset:


src/addons/badges/pages/badge-class/badge-class.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    protected badgeId = 0;
+    protected logView: (badge: AddonBadgesBadgeClass) => void;
+
+    badge?: AddonBadgesBadgeClass;
+    badgeLoaded = false;
+    currentTime = 0;
+
+    constructor(protected route: ActivatedRoute) {
+        this.badgeId = parseInt(CoreNavigator.getRouteParam('badgeId') ?? '');
```
dpalou, line 41: Can't you use `CoreNavigator.getRouteNumberParam` instead? It automatically converts to number. If the parameter is undefined, it will return undefined.

src/addons/badges/pages/badge-class/badge-class.ts
```diff
@@ -0,0 +1,91 @@
...
+
+    /**
+     * Refresh the badge class.
+     *
+     * @param refresher Refresher.
+     */
+    async refreshBadgeClass(refresher?: HTMLIonRefresherElement): Promise<void> {
+        await CoreUtils.ignoreErrors(AddonBadges.invalidateBadgeClass(this.badgeId));
+
+        await CoreUtils.ignoreErrors(this.fetchBadgeClass());
```
dpalou, line 86: This `ignoreErrors` isn't needed because `fetchBadgeClass` already captures errors. It doesn't hurt, so if you prefer leaving it it's ok for me.

dpalou (changes requested):

src/addons/badges/pages/badge-class/badge-class.ts
albertgasset: I've change it to `CoreNavigator..getRequiredRouteNumberParam`

src/addons/badges/pages/badge-class/badge-class.ts
albertgasset: I've remove it

dpalou (approved):

PR #4187, opened 2024-09-26
MOBILE-4505 course: Avoid show two info icons in navbar

alfonso-salces:


src/core/features/course/pages/index/index.html
```diff
@@ -10,11 +10,19 @@ <h1>
...
                 <ion-icon name="fas-circle-info" slot="icon-only" aria-hidden="true" />
             </ion-button>
+
             <!-- Add an empty context menu so tab pages can add items, otherwise the menu disappears in some cases. -->
             <core-context-menu />
+            } @else {
+            <core-context-menu>
+                <core-context-menu-item [priority]="400" [content]="'core.course.coursesummary' | translate" (action)="openCourseSummary()"
+                    iconAction="fas-circle-info" />
+            </core-context-menu>
```
dpalou, line 24: Talking with Juan, we thought it would be better to display this option last in the menu, so we should use a really low priority. We don't allow priority=0, so I suggest using "1".

Also, instead of using fas-circle-info, in this case it would be better to use fas-graduation-cap to have a different icon.

src/addons/mod/assign/components/index/addon-mod-assign-index.html
```diff
@@ -1,8 +1,15 @@
 <!-- Buttons to add to the header. -->
 <core-navbar-buttons slot="end">
+    @if (!courseContentsPage) {
     <ion-button fill="clear" (click)="openModuleSummary()" aria-haspopup="true" [ariaLabel]="'core.info' | translate">
         <ion-icon name="fas-circle-info" slot="icon-only" aria-hidden="true" />
     </ion-button>
+    } @else {
+    <core-context-menu>
+        <core-context-menu-item [priority]="500" [content]="'core.info' | translate" (action)="openModuleSummary()"
```
dpalou, line 9: Also talking with Juan we thought we should try to display this option first, so I recommend using a bigger priority value in all activities. Something like "5000", or even bigger if you want.

src/addons/mod/workshop/components/index/addon-mod-workshop-index.html
```diff
@@ -1,8 +1,15 @@
...
+    @if (!courseContentsPage) {
     <ion-button fill="clear" (click)="openModuleSummary()" aria-haspopup="true" [ariaLabel]="'core.info' | translate">
         <ion-icon name="fas-circle-info" slot="icon-only" aria-hidden="true" />
     </ion-button>
+    } @else {
+    <core-context-menu>
+        <core-context-menu-item [priority]="500" [content]="'core.info' | translate" (action)="openModuleSummary()"
+            iconAction="fas-circle-info" />
+    </core-context-menu>
+    }
```
dpalou, line 12: Adding the comment here, but it's not related to this file. You should also apply this change in core-siteplugins-module-index.html, that way the fix will also be applied to activity plugins.

dpalou (changes requested):

dpalou (approved):

PR #4193, opened 2024-10-01
Mobile 4660

crazyserver:


src/addons/storagemanager/pages/course-storage/course-storage.ts
```diff
@@ -213,146 +227,158 @@ export class AddonStorageManagerCourseStoragePage implements OnInit, OnDestroy {
...
 
                 // Recalculate the status.
-                await CoreCourseHelper.calculateSectionStatus(section, this.courseId, false);
+                await this.calculateSectionsStatus([sectionFinder.section]);
 
-                if (section.isDownloading && !CoreCourseModulePrefetchDelegate.isBeingDownloaded(downloadId)) {
+                if (sectionFinder.section.isDownloading && !CoreCourseModulePrefetchDelegate.isBeingDownloaded(downloadId)) {
                     // All the modules are now downloading, set a download all promise.
-                    this.prefecthSection(section);
+                    this.prefetchSection(sectionFinder.section);
```
dpalou, line 240: When `data.sectionId` is a subsection, you're checking if the parent section is downloading (`section.isDownloading`) but in `isBeingDownloaded` you pass `downloadId` (which is calculated using data.sectionId, the subsection id). So in one condition you use the section and in the other you use the subsection. This seems wrong.

I'm not sure if in this case you should check the subsection, the parent section or both. But right now the check doesn't seem consistent.

src/addons/storagemanager/pages/course-storage/course-storage.ts
```diff
@@ -213,146 +227,158 @@ export class AddonStorageManagerCourseStoragePage implements OnInit, OnDestroy {
...
+            }));
 
-            if (!module) {
+            if (!moduleFound) {
                 return;
             }
 
             // Call determineModuleStatus to get the right status to display.
-            const status = CoreCourseModulePrefetchDelegate.determineModuleStatus(module, data.status);
+            const status = CoreCourseModulePrefetchDelegate.determineModuleStatus(moduleFound, data.status);
```
dpalou, line 286: Should we call this function for mod_subsection modules? They don't have a status, it's always not downloaded because their prefetch handler doesn't store a status (and it's ok that it doesn't store it, because the status is calculated dynamically in this page).

src/core/features/course/components/course-format/course-format.ts
```diff
@@ -800,6 +808,10 @@ export class CoreCourseFormatComponent implements OnInit, OnChanges, OnDestroy {
             this.currentSite?.getLocalSiteConfig<string>(`${COURSE_EXPANDED_SECTIONS_PREFIX}${this.course.id}`),
         );
 
+        if (expandedSections === undefined) {
+            this.accordionMultipleValue = [];
+        }
+
         // Expand all sections if not defined.
         if (expandedSections === undefined) {
```
dpalou, line 816: This empty array line can be moved to the existing if:

```suggestion
        // Expand all sections if not defined.
        if (expandedSections === undefined) {
            this.accordionMultipleValue = [];
```

src/addons/storagemanager/pages/course-storage/course-storage.ts
```diff
@@ -519,40 +530,36 @@ export class AddonStorageManagerCourseStoragePage implements OnInit, OnDestroy {
...
      * Confirm and prefetch a section. If the section is "all sections", prefetch all the sections.
      *
      * @param section Section to download.
      */
-    async prefecthSection(section: AddonStorageManagerCourseSection): Promise<void> {
+    async prefetchSection(section: AddonStorageManagerCourseSection): Promise<void> {
         section.isCalculating = true;
         this.changeDetectorRef.markForCheck();
         try {
             await CoreCourseHelper.confirmDownloadSizeSection(this.courseId, section, this.sections);
```
dpalou, line 542: While testing this I noticed that this function calls `CoreCourseModulePrefetchDelegate.getDownloadSize` passing `section.modules`, and `section.modules` can include a mod_subsection. But subsection don't have a download size, they return -1 (cannot calculate), so for those sections it will always say that the size couldn't be calculated.

dpalou (changes requested):
Taking into account all the problems we're facing here, maybe it would have been better to take a different approach: when we receive the data from WS, we could modify the data to replace all the mod_subection activities with their section, and remove the subsection from the list of sections. That way in section contents we would have modules and sections, and all the app would receive this new structure (there would be no need to filter the list of sections in the page, or calculate module.subsection and so on). For the app, mod_subsection wouldn't exist, only the function converting the WS data would know that there is an activity called mod_subsection.

I guess it's too late now to try this approach, and maybe it causes other problems, so for now I guess it's better to keep your approach.

src/addons/storagemanager/pages/course-storage/course-storage.ts
crazyserver: I've change the conditions to find it.

dpalou:
Closing this PR in favor of https://github.com/moodlehq/moodleapp/pull/4194, which includes these commits along with others.

PR #4218, opened 2024-10-21
MOBILE-4686 user-menu: Fix routes with canLeave cannot leave

alfonso-salces:


src/core/services/navigator.ts
```diff
@@ -726,7 +726,13 @@ export class CoreNavigatorService {
      * @returns Whether the current route page can block leaving the route.
      */
     currentRouteCanBlockLeave(): boolean {
```
dpalou, line 728: You're reversing the result of this function for no reason. Before, this function returned true if the route could block, now it returns false if it blocks, which doesn't make sense IMO.

src/core/services/navigator.ts
```diff
@@ -726,7 +726,13 @@ export class CoreNavigatorService {
...
-        return !!this.getCurrentRoute().snapshot?.routeConfig?.canDeactivate?.length;
+        const canDeactivate = this.getCurrentRoute().snapshot?.routeConfig?.canDeactivate;
+
+        if (!canDeactivate?.length) {
+            return true;
+        }
+
+        return canDeactivate.every(method => typeof method === 'function' && method());
```
dpalou, line 735: This code doesn't work. The canDeactivate function (in here, `method`) expects to receive an instance of the component of the current route, and you aren't passing anything. This throws an error in console:

```
ERROR TypeError: Cannot use 'in' operator to search for 'canLeave' in undefined
    at isCanLeave (can-leave.ts:25:37)
```

I checked the contents of `getCurrentRoute()` and I see we have access to the class of the component, but not the instance so I'm not sure if this is feasible.

dpalou, line 735: Even if we manage to fix this, I think this will cause a bad UX for the user. Your code is calling the `canDeactivate` function to check if the user can leave or not, which in most cases will display a confirm modal to the user, but you're not leaving the page.

After the user confirms that he wants to leave, the logout process will continue and this process performs a navigation to '/login/sites', which will trigger Angular's canDeactivate function again, displaying the confirm modal a second time. This is not what we want.

src/core/services/navigator.ts
dpalou, line 728: Your change is a breaking change that will break the usage in behat-blocking. IMO you should leave the previous function as it was and create a new one for the logout/switch account use case.

dpalou (changes requested):

dpalou:
@alfonso-salces now that MOBILE-4028 has been fixed I guess we can close this PR, right?

alfonso-salces:
@dpalou I think so!

PR #4221, opened 2024-10-22
MOBILE-4616 core: Bump version code to 45002

crazyserver:


albertgasset (changes requested):
I think the  ios-CFBundleVersion needs to be updated

albertgasset (approved):

PR #4236, opened 2024-11-13
MOBILE-4688 file: Fix files with unencoded % in their name

dpalou:


src/core/singletons/opener.ts
```diff
@@ -128,17 +128,27 @@ export class CoreOpener {
...
         }
 
-        const openFile = async (mimetype?: string) => {
-            if (CoreOpener.shouldOpenWithDialog(options)) {
-                await FileOpener.showOpenWithDialog(path, mimetype || '');
-            } else {
-                await FileOpener.open(path, mimetype || '');
+        const openFile = async (path: string, mimetype?: string) => {
+            try {
+                if (this.shouldOpenWithDialog(options)) {
```
crazyserver, line 133: This is inside an static function and should use CoreOpener instead of this ;-)

crazyserver (changes requested):

PR #4241, opened 2024-11-19
MOBILE-4704 navigator: Exit from full screen mode while navigate

alfonso-salces:


src/core/services/navigator.ts
```diff
@@ -150,6 +150,13 @@ export class CoreNavigatorService {
             ? await NavController.navigateRoot(url, navigationOptions)
             : await NavController.navigateForward(url, navigationOptions);
 
+        // This is done to exit full screen if the user navigate.
+        if (document.exitFullscreen) {
+            document.exitFullscreen();
+        } else if (document['webkitExitFullscreen']) {
+            document['webkitExitFullscreen']();
+        }
```
dpalou, line 158: I did some tests in Android and I noticed that using `exitFullscreen()` works fine even if the full screen was opened using `webkitExitFullscreen`, so that's good.

But I also noticed that `exitFullscreen()` returns a Promise, according to documentation: a Promise which is resolved once the user agent has finished exiting fullscreen mode. That promise is rejected when the app is NOT in fullscreen, and this means an error is displayed in the JS console. The 'webkit' function doesn't return anything. So IMO you should do:

```suggestion
        // This is done to exit full screen if the user navigate.
        if (document.exitFullscreen) {
            await CorePromiseUtils.ignoreErrors(document.exitFullscreen());
        } else if (document['webkitExitFullscreen']) {
            document['webkitExitFullscreen']();
        }
```

dpalou (changes requested):

dpalou (approved):

PR #4242, opened 2024-11-19
MOBILE-4138 core: Wait for ready on delegates

crazyserver:


src/core/classes/delegate.ts
```diff
@@ -111,15 +102,7 @@ export class CoreDelegate<HandlerType extends CoreDelegateHandler> {
     }
 
     /**
-     * Check if the delegate is enabled so handlers are not updated if not..
-     *
-     * @returns Whether the delegate is enabled.
-     */
-    async isEnabled(): Promise<boolean> {
-        return true;
-    }
```
dpalou, line 120: This function is overridden in some delegates, like AddonModQuizAccessRuleDelegate. If the delegate is disabled there's no need to calculate the handlers, that's why this function was implemented.

dpalou (changes requested):

PR #4245, opened 2024-11-21
Mobile 4028

dpalou:


src/core/services/sites.ts
```diff
@@ -1494,8 +1499,13 @@ export class CoreSitesProvider {
      * @param siteId Site that will be opened after logout.
      * @param redirectData Page/url to open after logout.
      * @returns Promise resolved with boolean: true if app will be reloaded after logout.
+     * @deprecated since 5.0. Use CoreSites.logout instead, it automatically handles redirects.
      */
-    async logoutForRedirect(siteId: string, redirectData: CoreRedirectPayload): Promise<boolean> {
+    async logoutForRedirect(
+        siteId: string,
+        redirectData: CoreRedirectPayload,
+        logoutOptions?: InternalLogoutOptions,
```
crazyserver, line 1507: It's a bit weird to add a param to a deprecated function because it should not be used...  You can leave it this way but the JSdoc is missing.

src/core/services/sites.ts
```diff
@@ -1457,10 +1449,23 @@ export class CoreSitesProvider {
...
+    }
+
+    /**
+     * Logout the user.
+     * This function is for internal usage, please use CoreSites.logout instead.
+     *
      * @param options Logout options.
      * @returns Promise resolved when the user is logged out.
      */
-    async logout(options: CoreSitesLogoutOptions = {}): Promise<void> {
```
crazyserver, line 1463: Can you add a comment to explain why this function cannot be protected or private?

crazyserver (changes requested):

crazyserver (approved):

PR #4249, opened 2024-11-28
MOBILE-4653 chore: Improve formatPixelsSize to admit more units

crazyserver:


src/core/singletons/dom.ts
```diff
@@ -739,6 +739,51 @@ export class CoreDom {
...
+     * If the size is already valid (like '500px' or '50%') it won't be modified.
+     * Returned size will have a format like '500px'.
+     *
+     * @param size Size to format.
+     * @returns Formatted size. If size is not valid, returns an empty string.
+     */
+    static formatSizeUnits(size: string | number): string {
+        // Check for valid pixel units.
+        if (typeof size === 'string') {
+            size = size.replace(' ', ''); // Trim and remove all spaces.
```
dpalou, line 753: This will only replace the first space, not all of them. You should do:

```suggestion
            size = size.replace(/ /g, ''); // Trim and remove all spaces.
```

dpalou (changes requested):

dpalou (approved):

PR #4252, opened 2024-12-04
MOBILE-4653 core: Add site logo component

crazyserver:


src/core/features/mainmenu/components/user-menu/user-menu.html
```diff
@@ -13,30 +13,22 @@ <h1>
...
             <ion-item button class="core-usermenu-siteinfo ion-text-wrap" *ngIf="siteInfo && displaySiteUrl" lines="full" [detail]="false"
                 [href]="siteUrl" core-link>
                 <ion-label>
                     <!-- Show site logo. -->
-                    <img class="core-usermenu-site-logo" *ngIf="siteLogo && siteLogoLoaded" [src]="siteLogo" role="presentation" alt=""
-                        core-external-content onError="this.class='image-not-found'">
-                    <p class="core-usermenu-sitename">
-                        <core-format-text [text]="siteName" contextLevel="system" [contextInstanceId]="0" [wsNotFiltered]="true" />
-                    </p>
+                    <core-site-logo [hideOnError]="true" fallbackLogo="assets/img/login_logo.png" />
```
dpalou, line 23: I think that using the new component in user-menu has changed the behaviour. Before this change, the user-menu called `getPublicConfig` (if needed) to fetch the logo, but the `core-site-logo` component doesn't do that.

AFAIK when you login in a site we don't set the public config, we only call `getPublicConfig` in the places that we need it, so it's possible that the site instance doesn't have this info (and so the logo won't be displayed).

dpalou (changes requested):

src/core/features/courses/pages/my/my.html
```diff
@@ -5,8 +5,7 @@
         </ion-buttons>
         <ion-title>
             <h1>
-                <core-format-text [text]="siteName" contextLevel="system" [contextInstanceId]="0" class="core-header-sitename" />
-                <img src="assets/img/top_logo.png" class="core-header-logo" [alt]="siteName">
```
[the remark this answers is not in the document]
crazyserver: The visibility of the logo is controlled by CSS. So it shows the logo or the title and this is still the same.

src/core/components/site-logo/site-logo.ts
```diff
@@ -0,0 +1,115 @@
...
+    siteId?: string;
+    siteLogo?: string;
+    logoLoaded = false;
+    fallbackLogo = '';
+    showSiteName = true;
+
+    protected updateSiteObserver: CoreEventObserver;
+
+    constructor() {
+        this.siteId = CoreSites.getCurrentSiteId();
```
dpalou, line 52: If you receive a `site` instance via input then you should use its ID (if it has an ID). In credentials we won't have an ID, but in reconnect we will.

src/core/components/site-logo/site-logo.html
```diff
@@ -0,0 +1,13 @@
+<div class="core-logo-container" *ngIf="showLogo && logoLoaded">
+    <img *ngIf="siteLogo" [src]="siteLogo" class="core-logo" [alt]="siteName" core-external-content (error)="imageError()">
```
dpalou, line 2: You're passing the siteId to format-text, but not to this core-external-content.

src/core/components/site-logo/site-logo.html
```diff
@@ -0,0 +1,13 @@
...
+    <img *ngIf="!siteLogo" [src]="fallbackLogo" class="core-logo" [alt]="siteName">
+</div>
+
+<p *ngIf="showSiteName && siteNameMode === 'p' && siteName" class="ion-no-margin ion-no-padding core-logo-sitename">
+    <core-format-text [text]="siteName" [filter]="false" contextLevel="system" [contextInstanceId]="0" [siteId]="siteId" />
```
dpalou, line 7: In some cases you're using [filter]="false", but in others you don't. I guess this is because of copy paste. Since we cannot know where will this component be used IMO we should remove this filter=false (and maybe we should improve the format-text directive to avoid calling the filter functions if there is no site ID?

src/core/features/courses/pages/my/my.html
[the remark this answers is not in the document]
dpalou: I don't agree that the behaviour is the same now :P But as we talked in chat we'll support using an online URL for top logo too, so it's ok like this.

dpalou (approved):

PR #4266, opened 2024-12-16
Mobile 3063

crazyserver:


src/core/features/viewer/services/viewer.ts
```diff
@@ -97,6 +99,50 @@ export class CoreViewerService {
...
+        return settings;
+    }
+
+    protected async applyReadingModeSettings(settings: CoreViewerReadingModeSettings): Promise<void> {
+        document.body.style.setProperty('--reading-mode-zoom', settings.zoom + '%');
+        Object.values(CoreViewerReadingModeThemes).forEach((theme) => {
+            document.body.classList.remove(`core-reading-mode-theme-${theme}`);
+        });
+        document.body.classList.add(`core-reading-mode-theme-${settings.theme}`);
+        document.body.classList.toggle('core-reading-mode-multimedia-hidden', !settings.showMultimedia);
```
dpalou, line 129: If I'm not mistaken, when exiting the reading mode you're leaving all this CSS in the DOM. Shouldn't you remove it too? I know it's not applied because other classes are removed, but maybe we should clean it up too.

src/core/singletons/dom.ts
```diff
@@ -739,6 +739,32 @@ export class CoreDom {
         return css.replace(regExp, prefix + ' $1 $2');
     }
 
+    /**
+     * Remove all classes from all the DOM.
+     *
+     * @returns Clean text.
+     */
+    static removeAllStyles(text: string | undefined): string {
```
dpalou, line 747: I think this function isn't used, I'm not sure if you want to keep it or not.

dpalou (changes requested):

src/core/directives/reading-mode.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    /**
+     * Enters the reading mode.
+     */
+    protected async enterReadingMode(): Promise<void> {
+        this.enabled = true;
+        CoreViewer.getReadingModeSettings(true);
+
+        document.body.classList.add('core-reading-mode-enabled');
+        this.contentEl?.classList.add('core-reading-mode-content');
```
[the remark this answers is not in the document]
crazyserver: Yep, this is something we can do only once, so I moved it to addTextViewerButton instead

scripts/langindex.json
```diff
@@ -2683,6 +2683,19 @@
...
+  "core.viewer.decreasetextsize": "local_moodlemobileapp",
+  "core.viewer.default": "moodle",
+  "core.viewer.enterreadingmode": "local_moodlemobileapp",
+  "core.viewer.exitreadingmode": "local_moodlemobileapp",
+  "core.viewer.increasetextsize": "local_moodlemobileapp",
+  "core.viewer.openreadingmodesettings": "local_moodlemobileapp",
+  "core.viewer.readingthemeauto": "local_moodlemobileapp",
+  "core.viewer.readingthemedark": "local_moodlemobileapp/core.settings.colorscheme-dark",
+  "core.viewer.readingthemehcm": "local_moodlemobileapp",
+  "core.viewer.readingthemelight": "local_moodlemobileapp/core.settings.colorscheme-light",
```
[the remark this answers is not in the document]
crazyserver: Not to duplicate the string that is yet in local_moodlemobileapp

src/core/features/viewer/services/viewer.ts
crazyserver: I think it's ok to leave it there, it defines the settings but disabled.

dpalou (approved):

PR #4282, opened 2025-01-13
Mobile 4724

dpalou:


src/core/services/utils/dom.ts
```diff
@@ -111,85 +81,7 @@ export class CoreDomUtilsProvider {
...
-            message = message || (size.size === 0 ? 'core.course.confirmdownloadzerosize' : 'core.course.confirmdownload');
-
-            return this.showConfirm(
-                wifiPrefix + Translate.instant(
-                    message,
-                    { size: readableSize, availableSpace: availableSpace },
-                ),
-            );
-        }
+        return CoreAlerts.confirmDownloadSize(size, { message, unknownMessage, wifiThreshold, limitedThreshold, alwaysConfirm });
```
crazyserver, line 84: I'd import CoreAlerts here since it's deprecated and async function

src/core/services/utils/dom.ts
```diff
@@ -393,104 +285,40 @@ export class CoreDomUtilsProvider {
...
-            const matches = errorMessage.match(/token"?[=|:]"?(\w*)/);
-            if (matches?.[1]) {
-                errorMessage = errorMessage.replace(new RegExp(matches[1], 'g'), 'secret');
-            }
-        } else {
-            errorMessage = error;
-        }
-
-        let message = CoreText.decodeHTML(needsTranslate ? Translate.instant(errorMessage) : errorMessage);
+        const message = CoreAlerts.getErrorMessage(error);
```
crazyserver, line 297: I'd import CoreAlerts here since it's deprecated and async function

src/core/services/utils/dom.ts
```diff
@@ -393,104 +285,40 @@ export class CoreDomUtilsProvider {
...
     /**
      * Check whether an error is an error caused because the user canceled a showConfirm.
      *
      * @param error Error to check.
      * @returns Whether it's a canceled error.
+     * @deprecated since 5.0. Use CoreErrorHelper.isCanceledError instead.
      */
     isCanceledError(error: CoreAnyError): boolean {
-        return error instanceof CoreCanceledError;
+        return CoreErrorHelper.isCanceledError(error);
```
crazyserver, line 310: I'd import CoreErrorHelper here since it's deprecated and async function

src/core/services/utils/dom.ts
```diff
@@ -903,7 +662,12 @@ export class CoreDomUtilsProvider {
         message = message ? Translate.instant(message) : message;
         buttonText = buttonText ? Translate.instant(buttonText) : buttonText;
 
-        return this.showAlert(header, message, buttonText, autocloseTime);
+        return CoreAlerts.show({
```
crazyserver, line 665: Same as the other comments and there are more in this file

src/core/services/utils/dom.ts
```diff
@@ -1384,32 +937,6 @@ export class CoreDomUtilsProvider {
         return element.children;
     }
 
-    /**
-     * Treat anchors inside alert/modals.
-     *
-     * @param container The HTMLElement that can contain anchors.
-     */
-    treatAnchors(container: HTMLElement): void {
```
crazyserver, line 1392: Shouldn't this function be deprecated?

crazyserver (changes requested):

src/core/services/utils/dom.ts
dpalou: I cannot change it because I'd need to make the function async, which is a breaking change. The purpose of keeping the functions deprecated is to avoid breaking changes, so IMO it's better to keep it like this.

src/core/services/overlays/modals.ts
```diff
@@ -188,28 +187,10 @@ export class CoreModalsService {
...
-        );
-
-        if (modalData === undefined) {
-            throw new CoreCanceledError();
-        } else if (modalData instanceof CoreWSError) {
-            throw modalData;
-        }
-
-        return modalData;
+        return CorePrompts.promptPassword(passwordParams);
```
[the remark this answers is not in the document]
dpalou: Changed!

src/core/services/utils/dom.ts
dpalou: As above, I cannot change this because I'd need to make the function async.

src/core/services/utils/dom.ts
dpalou: I wasn't sure what to do about this. This function should have been protected IMO, and it's protected now in CoreAlerts. I don't think any plugin is using this, so I decided to remove it.

crazyserver (approved):

PR #4287, opened 2025-01-15
Mobile 4736

crazyserver:


src/addons/qtype/ordering/component/ordering.ts
```diff
@@ -79,6 +79,12 @@ export class AddonQtypeOrderingComponent extends CoreQuestionBaseComponent<Addon
                 '.icon.fa-circle-xmark, .icon.fa-square-check, .icon.circle-half-stroke',
             )?.remove();
 
+            // New icon used for partially correct answers (MDL-83639).
+            const otherIcon = itemContentEl?.querySelector<HTMLImageElement>('img.icon[src]');
+            if (otherIcon && otherIcon.src.indexOf('grade_partiallycorrect') >= 0) {
+                otherIcon.remove();
+            }
```
dpalou, line 86: This can be added in the previous querySelector, using this selector:

`img.icon[src*="grade_partiallycorrect"]`

src/core/features/question/services/question-helper.ts
```diff
@@ -887,7 +901,10 @@ export class CoreQuestionHelperProvider {
...
             const partiallyCorrectIcon = this.getPartiallyCorrectIcon();
             if ('src' in icon) {
-                if ((icon as HTMLImageElement).src.indexOf('correct') >= 0) {
+                if ((icon as HTMLImageElement).src.indexOf('grade_partiallycorrect') >= 0) {
+                    iconData = partiallyCorrectIcon;
+                    color = CoreIonicColorNames.WARNING;
+                } else if((icon as HTMLImageElement).src.indexOf('correct') >= 0) {
                     iconData = correctIcon;
                     color = CoreIonicColorNames.SUCCESS;
                 } else if ((icon as HTMLImageElement).src.indexOf('incorrect') >= 0 ) {
```
dpalou, line 910: This code was already like this, but I think it's wrong. If the src contains the word "incorrect" it will always enter in the "correct" condition, it will never reach the "incorrect" case. The app should check "incorrect" first.

dpalou (changes requested):

dpalou (approved):

PR #4289, opened 2025-01-15
Mobile 4690

crazyserver:


src/core/components/dynamic-component/dynamic-component.ts
```diff
@@ -124,6 +126,13 @@ export class CoreDynamicComponent<ComponentClass> implements OnChanges, DoCheck
         }
     }
 
+    /**
+     * @inheritdoc
+     */
+    async ready(): Promise<void> {
```
dpalou, line 132: Since dynamic-component now implements the `ready` callback I would specify that it `implements AsyncDirective`. That way we make sure the signature is consistent with the rest of the app.

src/core/components/dynamic-component/dynamic-component.ts
```diff
@@ -183,6 +192,17 @@ export class CoreDynamicComponent<ComponentClass> implements OnChanges, DoCheck
...
+        if (typeof this.instance['onReadyPromise'] === 'undefined') {
+            this.onReadyPromise.resolve();
+        } else {
+            // eslint-disable-next-line promise/catch-or-return
+            this.instance.onReadyPromise.then(() => {
+                this.onReadyPromise.resolve();
+
+                return;
+            });
+        }
```
dpalou, line 204: Instead of relying in `onReadyPromise`, which is something that is usually protected and we don't have any defined signature, I think it would be better to use the `ready` callback since it's more standard thanks to `AsyncDirective`.

This means changing the approach, e.g. you will probably need a promise to wait until instance is loaded, and then in the `ready` function you could do something like:

```
const instance = await this.promisedInstance;

if (typeof instance['ready'] === 'function') {
    await instance.ready();
}
```

If we do this you could remove the `instance` property and use either `await this.promisedInstance` or `this.promisedInstance.value` depending if you need it to be sync or not.

Also, I guess in base-question-component you will need to implement `AsyncDirective`.

src/addons/qtype/essay/services/handlers/essay.ts
```diff
@@ -148,6 +150,10 @@ export class AddonQtypeEssayHandlerService implements CoreQuestionHandler {
             return;
         }
 
+        if (this.isCompleteResponse(question, answers, component, componentId) === 0) {
+            return;
+        }
```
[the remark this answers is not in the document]
crazyserver: Ooops good catch!

src/core/components/dynamic-component/dynamic-component.ts
crazyserver: Thanks It was done this way but for any reason I've lost that change.

src/core/components/dynamic-component/dynamic-component.ts
```diff
@@ -112,18 +119,32 @@ export class CoreDynamicComponent<ComponentClass> implements OnChanges, DoCheck
...
     ngDoCheck(): void {
-        if (this.instance) {
-            // Check if there's any change in the data object.
-            const changes = this.differ.diff(this.data || {});
-            if (changes) {
-                this.setInputData();
-                if (this.instance.ngOnChanges) {
-                    this.instance.ngOnChanges(CoreDomUtils.createChangesFromKeyValueDiff(changes));
-                }
+        const instance = this.promisedInstance.value;
```
dpalou, line 122: Since you created a getter to retrieve the `instance`, you can use it instead of using `promisedInstance` directly. That way most of the code in this directive can remain as it was, just make sure to check it's not undefined (e.g. in `setInputData` it's not done right now).

src/core/features/question/constants.ts
```diff
@@ -37,3 +37,12 @@ export const enum QuestionDisplayOptionsValues {
     VISIBLE = 1,
     EDITABLE = 2,
 }
+
+/**
+ * Possible values for the question complete response or gradable response (compatible).
+ */
+export const enum QuestionCompleteGradableResponse {
+    UNDETERMINATED = -1,
```
dpalou, line 45: I'm not sure if UNDETERMINATED is a valid word (or the right word here), if I search in Google it says if I meant UNDETERMINED. Maybe we should use "UNKNOWN"?

dpalou (changes requested):

src/core/features/question/classes/base-question-component.ts
```diff
@@ -726,12 +729,19 @@ export class CoreQuestionBaseComponent<T extends AddonModQuizQuestion = AddonMod
         return questionEl;
     }
 
+    /**
+     * @inheritdoc
+     */
+    async ready(): Promise<void> {
+        return this.onReadyPromise;
+    }
```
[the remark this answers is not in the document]
crazyserver: Sorry I forgot to add it

PR #4303, opened 2025-01-29
Mobile 4653

crazyserver:


src/core/directives/long-press.ts
```diff
@@ -49,10 +50,10 @@ export class CoreLongPressDirective implements OnInit, OnDestroy {
             gestureName: 'longpress',
             onStart: (event) => {
                 this.timeout = window.setTimeout(() => {
-                    this.longPress.emit(event);
+                    this.longPress.emit(event.event);
```
dpalou, line 53: Instead of changing this here, which changes the data passed to the `Output` and can break any usage of this directive, I think it's better to fix the problem in device-info and properly type the output here (right now it passes `GestureDetail`). I also think it's better to put the type `GestureDetail` in `copyItemInfo` instead of just event.

dpalou (changes requested):

dpalou (approved):

PR #4304, opened 2025-01-30
MOBILE-4653 loading: Add placeholder types for loading

crazyserver:


crazyserver:
Thanks Dani! I've solved all what you said. I think the new pipe is not needed to be exported to use in site plugins. However I changed my mind and I think the function I added to the last commit could be a good idea.

dpalou:
If this was only needed in 1 component I would agree about using a function, but you're using the same approach in calendar too, and maybe we need it in other places in the future. IMO a pipe is easier to reuse.

About exporting it to site plugins, it could be useful in some situations, specially if a plugin now wants to use their own placeholders as you do in calendar. They can use workarounds to achieve it, but having this pipe makes it easier. Since it's a small pipe with no imports IMO it's ok to export it.

BTW in the pipe description it could be nice to add an example of how it can be used (this way it's easier to understand why it's called "times" when the pipe just creates an array).

crazyserver:
Sooo I think that's it :-)

dpalou (approved):

PR #4305, opened 2025-01-30
Mobile 4653

crazyserver:


src/core/components/swipe-slides/swipe-slides.html
```diff
@@ -1,6 +1,11 @@
-<swiper-container #swiperRef *ngIf="loaded" [attr.aria-busy]="activeSlideIndex ? 'true' : null">
-    <swiper-slide *ngFor="let item of items; index as index" [attr.aria-hidden]="!isActive(index)"
-        [attr.tabindex]="!isActive(index) ? -1 : null">
-        <ng-container *ngIf="template" [ngTemplateOutlet]="template" [ngTemplateOutletContext]="{item: item, active: isActive(index)}" />
-    </swiper-slide>
-</swiper-container>
+@if (loaded) {
+    <swiper-container #swiperRef [attr.aria-busy]="activeSlideIndex ? 'true' : null">
+        @for (item of items; track item.id; let index = $index) {
```
dpalou, line 3: `item.id` doesn't exist, but TS doesn't know it because `Item` is a generic with value `unknown`. The way to get the ID of an item is to use `manager?.getSource().getItemId(item)`. I'm not sure how will the performance be when having a function in the `track` part.

src/core/components/tabs-outlet/core-tabs-outlet.html
```diff
@@ -1,37 +1,51 @@
...
-                                    <span *ngIf="tab.badgeA11yText" class="sr-only">
-                                        {{ tab.badgeA11yText | translate: {$a : tab.badge } }}
-                                    </span>
-                                </ion-badge>
-                            </ion-label>
-                        </ion-tab-button>
-                    </swiper-slide>
-                </ng-container>
+                @for (tab of tabs; track tab.id) {
+                    @if (tab.id) {
```
dpalou, line 14: The id of a tab is set when initializing the tab, so I guess we could improve the code to avoid having this check (e.g. maybe we could have a transform to add the required data when the input changes). But this can be done later, no need to do it now.

EDIT: I just saw that in core-tabs.html we just do `tab.id!`, but that isn't the best approach either.

dpalou (changes requested):

src/core/components/tabs-outlet/tabs-outlet.ts
```diff
@@ -51,14 +51,15 @@ import { CorePath } from '@singletons/path';
...
+export class CoreTabsOutletComponent extends CoreTabsBaseComponent<CoreTabsOutletTabWithId>
     implements AfterViewInit, OnChanges, OnDestroy {
 
     /**
      * Determine tabs layout.
      */
     @Input() layout: 'icon-top' | 'icon-start' | 'icon-end' | 'icon-bottom' | 'icon-hide' | 'label-hide' = 'icon-hide';
-    @Input() tabs: CoreTabsOutletTab[] = [];
+    @Input({ transform: (tabs: CoreTabsOutletTab[]): CoreTabsOutletTabWithId[] =>
+        tabs.map((tab) => CoreTabsOutletComponent.formatTab(tab)) }) tabs: CoreTabsOutletTabWithId[] = [];
```
dpalou, line 62: This input isn't marked as required, so it can be undefined (that's why we initialized it to an empty array). So the tabs param in the transform could be undefined, in that case we should return an empty array.

src/core/components/sites-list/sites-list.html
```diff
@@ -1,38 +1,48 @@
...
-    </ion-card>
+                @for (site of accountsList.sameSite; track site.id) {
+                    <ng-container *ngTemplateOutlet="siteItem; context: {site: site, isCurrentSite: false}" />
+                }
+            </ion-card>
+        }
 
-    <ion-card *ngFor="let sites of accountsList.otherSites">
-        <ng-container *ngTemplateOutlet="siteCardHeader; context: {site: sites[0], isCurrentSite: false}" />
+        @for (sites of accountsList.otherSites; track sites?.[0].siteUrl) {
```
dpalou, line 15: Instead of:

sites?.[0].siteUrl

I think it should be:

sites[0]?.siteUrl

dpalou (changes requested):

dpalou (approved):

PR #4308, opened 2025-02-04
Mobile 4759

crazyserver:


src/core/features/siteplugins/components/assign-feedback/assign-feedback.ts
```diff
@@ -57,6 +58,8 @@ export class CoreSitePluginsAssignFeedbackComponent extends CoreSitePluginsCompi
         this.jsData.edit = this.edit;
         this.jsData.canEdit = this.canEdit;
 
+        this.extraImports = await getModAssignComponentModules();
```
dpalou, line 61: Moving these extra imports (and the ones in quiz and workshop) from compile to the components means making the API more limited, which is a breaking change. However, we already mentioned several times that we should limit this more and it doesn't make sense that all plugins have access to assign components for example, so I'm ok with this change. However, it should be documented in upgrade.txt.

src/core/features/siteplugins/components/question-behaviour/question-behaviour.ts
```diff
@@ -71,6 +72,8 @@ export class CoreSitePluginsQuestionBehaviourComponent extends CoreSitePluginsCo
         this.jsData.buttonClicked = this.buttonClicked;
         this.jsData.onAbort = this.onAbort;
 
+        this.extraImports = await getModQuizComponentModules();
```
dpalou, line 75: Quiz uses qbehaviour and questions, but questions and qbehaviours usually don't import anything from quiz, they should be independent. So I guess we can remove the extra imports from here and also from question.

The only place where it could make sense to have them is for quiz access rules, because that code does belong to quiz. Looking at the components we have right now I'm not sure if any access rule will need to use those components, it doesn't seem likely. Since we're limiting the API maybe we want to remove them from there too, I leave that up to you.

src/core/features/user/components/user-profile-field/user-profile-field.ts
```diff
@@ -21,13 +21,18 @@ import { CoreUserProfileFieldDelegate } from '@features/user/services/user-profi
...
 
 /**
  * Directive to render user profile field.
  */
 @Component({
     selector: 'core-user-profile-field',
     templateUrl: 'core-user-profile-field.html',
+    standalone: true,
+    imports: [
+        CoreDynamicComponent,
```
dpalou, line 34: If we want to be consistent I guess we should import SharedModule here too.

src/core/features/usertours/components/user-tour/user-tour.ts
```diff
@@ -49,6 +50,10 @@ const BACKDROP_DISMISS_SAFETY_TRESHOLD = 1000;
     selector: 'core-user-tours-user-tour',
     templateUrl: 'core-user-tours-user-tour.html',
     styleUrl: 'user-tour.scss',
+    standalone: true,
+    imports: [
+        NgIf,
```
dpalou, line 55: As mentioned previous, for consistency maybe we want to use SharedModule here.

dpalou (changes requested):

dpalou (approved):

PR #4312, opened 2025-02-10
Mobile 4653 fixes

crazyserver:


src/core/components/user-avatar/core-user-avatar.html
```diff
@@ -1,28 +1,29 @@
...
 
-} @else if (!avatarUrl && initials) {
-
-    @if (linkProfile) {
-        <div class="userinitials" [attr.aria-label]="fullname" [title]="fullname" (ariaButtonClick)="gotoProfile($event)"
-            [attr.data-initials]="initials">
-        </div>
+}
+@if (!avatarUrl || imageError){
+    @if(initials) {
```
dpalou, line 13: You lack a space between if and the parenthesis.

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -115,10 +116,10 @@ export class CoreUserAvatarComponent implements OnInit, OnChanges, OnDestroy {
     }
 
     /**
-     * Avatar image loading error handler.
+     * Avatar image loading handler.
      */
-    loadImageError(): void {
-        this.avatarUrl = undefined;
+    loadImageSuccess(success: boolean): void {
```
dpalou, line 121: Since this function is called both on load and error I think we shouldn't call it "Success". I'd use the same name you set in the logo component: `imageLoaded`.

dpalou (changes requested):

src/core/features/mainmenu/services/home-delegate.ts
```diff
@@ -67,6 +67,13 @@ export interface CoreMainMenuHomeHandlerData {
      * Name of the icon to display for the handler.
      */
     icon?: string; // Name of the icon to display in the tab.
+
+    /**
+     * If tab is enabled or not. When disabled, tab will be hidden.
+     * This is necessary to manage swiper when tabs are hidden.
+     * If tab is hidden, but enabled the lateral buttons won't work properly.
```
dpalou, line 74: I don't understand the last line of this comment. If tabs are hidden when disabled, how can a tab be hidden but enabled?

src/core/classes/tabs.ts
```diff
@@ -271,7 +271,7 @@ export class CoreTabsBaseComponent<T extends CoreTabBase> implements AfterViewIn
      * Init the component.
      */
     protected async init(): Promise<void> {
-        if (!this.hideUntil || !this.swiper) {
+        if (!this.hideUntil || !this.swiper || this.firstSelectedTab) {
```
dpalou, line 274: Before this change, calling `init` also called `calculateSlides` if tabs were shown, but now that isn't true anymore. I'm not sure if that change can cause problems, maybe it would be better to call `calculateSlides` if tabs have already been initialized.

src/core/classes/tabs.ts
```diff
@@ -271,7 +271,7 @@ export class CoreTabsBaseComponent<T extends CoreTabBase> implements AfterViewIn
...
             // Hidden, do nothing.
```
dpalou, line 275: This comment isn't true anymore, this if can also mean that tabs are already initialized and not hidden.

dpalou (changes requested):

src/core/features/mainmenu/services/home-delegate.ts
crazyserver: I think now it's clearer, I meant hidden using CSS

src/core/classes/tabs.ts
crazyserver: Now it's true again

dpalou (approved):

PR #4319, opened 2025-02-18
Mobile 3671 Assignment previous attempts

crazyserver:


dpalou (changes requested):

src/addons/mod/assign/components/edit-feedback-plugin-modal/edit-feedback-plugin-modal.ts
```diff
@@ -79,7 +79,7 @@ export class AddonModAssignEditPluginFeedbackModalComponent {
      * @returns Object with the data.
      */
     protected getInputData(): CoreFormFields {
-        return CoreForms.getDataFromForm(document.forms['addon-mod_assign-edit-feedback-form']);
+        return CoreForms.getDataFromForm(document.forms['addon-mod_assign-edit-feedback-plugin-form']);
```
dpalou, line 82: Since now the idea is to show the form inline instead of using a modal, I guess we should also deprecate this component (`AddonModAssignEditPluginFeedbackModalComponent`) and also the function `editFeedback` from `AddonModAssignFeedbackPluginBaseComponent`, right?

src/addons/mod/assign/components/submission/submission.ts
```diff
@@ -156,53 +133,19 @@ export class AddonModAssignSubmissionComponent implements OnInit, OnDestroy, Can
...
-
-                // Don't refresh if the user has modified some data.
-                const hasDataToSave = await this.hasDataToSave();
-
-                if (!hasDataToSave) {
-                    this.invalidateAndRefresh(false);
-                }
-            },
-            this.siteId,
-        );
```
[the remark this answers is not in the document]
dpalou: I tried it agan and, after an automatic sync, the view wasn't refreshed. I guess some "if" is stopping the execution. Can you please review it?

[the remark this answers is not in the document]
crazyserver: He, he was listeing to ADDON_MOD_ASSIGN_MANUAL_SYNCED, ADDON_MOD_ASSIGN_MANUAL_SYNCED (Both manual) now it works

src/addons/mod/assign/services/feedback-delegate.ts
```diff
@@ -146,6 +149,7 @@ export interface AddonModAssignFeedbackHandler extends CoreDelegateHandler {
      * @param userId User ID.
      * @param plugin The plugin object.
      * @param pluginData Object where to store the data to send.
+     * @param inputData Data entered in the feedback form.
```
dpalou, line 152: This is a breaking change for plugins, you're adding a new parameter before an existing one. Site plugins can implement this function, and this change can break those plugins.

As far as I know there is no assign feedback plugin adapted to the app, so maybe we can document this in upgrade.txt and if a plugin is affected they will need to change their code.

src/core/classes/sites/site.ts
```diff
@@ -550,8 +550,8 @@ export class CoreSite extends CoreAuthenticatedSite {
      */
     getConfig(name?: undefined, ignoreCache?: boolean): Promise<CoreSiteConfig>;
     getConfig(name: string, ignoreCache?: boolean): Promise<string>;
-    getConfig(name?: string, ignoreCache?: boolean): Promise<string | CoreSiteConfig> {
-        return firstValueFrom(
+    async getConfig(name?: string, ignoreCache?: boolean): Promise<string | CoreSiteConfig> {
+        return await firstValueFrom(
```
dpalou, line 554: I guess this can happen in all the places where we use `firstValueFrom`. Can you please change them all? Adding an await shouldn't hurt, and it can fix this problem in other places.

dpalou (changes requested):

dpalou (approved):

PR #4325, opened 2025-02-25
MOBILE-4653 behat: Fix failure when opening PDF files

albertgasset:


local_moodleappbehat/tests/behat/behat_app_helper.php
```diff
@@ -829,4 +829,25 @@ protected function wait_animations_done() {
...
     }
+
+    /**
+     * Get window names, excluding Chrome extensions.
+     * Workaround for bug: https://github.com/SeleniumHQ/selenium/issues/15330
+     *
+     * @return array
+     */
+    protected function get_window_names(): array {
+        $wndowNames = [];
```
dpalou, line 840: Typo in the variable name: wndowNames instead of windowNames.

local_moodleappbehat/tests/behat/behat_app_helper.php
```diff
@@ -829,4 +829,25 @@ protected function wait_animations_done() {
...
+        foreach ($this->getSession()->getWindowNames() as $windowName) {
+            $this->getSession()->switchToWindow($windowName);
+            $windowUrl = $this->getSession()->getCurrentUrl();
+            if (strpos($windowUrl, 'chrome-extension://') !== 0) {
+                $wndowNames[] = $windowName;
+            }
+        }
+
+        $this->getSession()->switchToWindow($wndowNames[0]);
```
dpalou, line 849: Instead of assuming that the app was in the first window, I guess it would be better to return to "current" window when this function is executed. What do you think?

dpalou (changes requested):

local_moodleappbehat/tests/behat/behat_app_helper.php
albertgasset: Fixed

local_moodleappbehat/tests/behat/behat_app_helper.php
albertgasset: I agree, I've fixed it

dpalou (approved):

PR #4327, opened 2025-02-26
Mobile 4773

crazyserver:


src/core/singletons/time.ts
```diff
@@ -202,4 +202,32 @@ export class CoreTime {
...
+        const now = new Date();
+
+        const year = now.getFullYear();
+        const month = String(now.getMonth() + 1).padStart(2, '0');
+        const day = String(now.getDate()).padStart(2, '0');
+        const hours = String(now.getHours()).padStart(2, '0');
+        const minutes = String(now.getMinutes()).padStart(2, '0');
+        const seconds = String(now.getSeconds()).padStart(2, '0');
+
+        return `${year}${month}${day}${hours}${minutes}${seconds}`;
```
dpalou, line 250: I think this change could break behats. In the `initializeAutomatedTests` function, we set the timezone of the moment library to 'Australia/Perth' when using behats. This means that, before this change, this function always returned the readable time in Australia/Perth timezone for behats, but now it will use the machine's timezone.

dpalou (changes requested):

src/core/singletons/time.ts
crazyserver: Thanks Dani!

I'd consider solving [MOBILE-4549](https://tracker.moodle.org/browse/MOBILE-4549) on a near future, starting by deprecating the moment export.

dpalou (approved):

PR #4330, opened 2025-02-28
Mobile 4549

crazyserver:


src/addons/mod/feedback/tests/behat/groups.feature
```diff
@@ -0,0 +1,51 @@
...
+      | feedback1 | user2 | I don't know             |
+      | feedback1 | user3 | Not at all               |
+      | feedback1 | user4 | Yes of course            |
+      | feedback1 | user5 | Yes of course            |
+      | feedback1 | user6 | Not at all               |
+      | feedback1 | user7 | I don't know             |
+
+    When I entered the feedback activity "Course feedback" on course "C1" as "teacher" in the app
+    And I press "Analysis" in the app
+    And the UI should match the snapshot
```
dpalou, line 51:
```suggestion
    Then the UI should match the snapshot
```

src/core/singletons/time.ts
```diff
@@ -133,15 +133,62 @@ export class CoreTime {
...
+                ...dayjs.Ls['en'].relativeTime,
+                ss: '%d seconds',
+                mm: '%d minutes',
+                hh: '%d hours',
+                dd: '%d days',
+                // eslint-disable-next-line @typescript-eslint/naming-convention
+                MM: '%d months',
+                yy: '%d years',
+            },
+        });
```
dpalou, line 191: Why do we need to change the locale? Is it to make it match with moment?

dpalou (changes requested):

src/core/singletons/logger.ts
```diff
@@ -108,7 +108,7 @@ export class CoreLogger {
     private static prepareLogFn(logFn: LogFunction, className: string): LogFunction {
         // Return our own function that will call the logging function with the treated message.
         return (...args): void => {
-            const now = moment().format('l LTS');
+            const now = dayjs().format('l LTS');
```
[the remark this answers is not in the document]
crazyserver: dayjs is not equivalent to dayjs.tz, on tz you don't have some general functions like locale, localeData, months, monthsShort... But I changed that, yes!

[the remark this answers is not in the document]
crazyserver: Also to use tz, the plugin must be initialized and this is not the case.

src/core/singletons/time.ts
crazyserver: This is needed to solve an error on relative translations I've added a new comment

dpalou (approved):

dpalou (approved):

PR #4338, opened 2025-03-06
Mobile 4675

crazyserver:


src/core/services/filepool-queue.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+    /**
+     * Remove a file from the queue.
+     *
+     * @param siteId The site ID.
+     * @param fileId The file ID.
+     */
+    async remove(siteId: string, fileId: string): Promise<void> {
+        await CorePromiseUtils.ignoreErrors(this.table.deleteByPrimaryKey({ siteId, fileId }));
```
dpalou, line 201: IMO this function shouldn't ignoreErrors, because then there's no way to tell if the operation was successful or not. The code calling this function should be the one deciding if the error should be ignored or not.

src/core/services/filepool.ts
```diff
@@ -2297,7 +2196,7 @@ export class CoreFilepoolProvider {
      * @returns Resolved with file object from DB on success, rejected otherwise.
      */
     protected async hasFileInQueue(siteId: string, fileId: string): Promise<CoreFilepoolQueueEntry> {
```
dpalou, line 2195: This function is no longer used and can be removed.

src/core/services/filepool-queue.ts
```diff
@@ -0,0 +1,338 @@
...
+import { CoreText } from '@singletons/text';
+import { CorePromiseUtils } from '@singletons/promise-utils';
+import { CorePromisedValue } from '@classes/promised-value';
+import { CoreFilepoolOnProgressCallback } from './filepool';
+
+/**
+ * Service to handle the filepool queue.
+ */
+@Injectable({ providedIn: 'root' })
+export class CoreFilepoolQueueService {
```
dpalou, line 45: Since this class should only be used inside Filepool service and it doesn't seem likely that we need to override its functions, maybe we should have it inside the same file as the filepool service, as a normal class instead of a service. What do you think?

src/core/services/filepool-queue.ts
```diff
@@ -172,19 +177,18 @@ export class CoreFilepoolQueueService {
...
+            const entry = await this.table.getOne({}, {
+                sorting: [
+                    { priority: 'desc' },
+                    { added: 'asc' },
+                ],
+            });
+
+            return entry;
+        } catch {
+            throw new CoreError('Queue is empty.');
```
dpalou, line 190: In this case, instead of throwing an error maybe we can just return undefined if it isn't found, since you aren't asking for a specific entry, you just want the first. If there's no first, you get undefined. What do you think?

src/core/services/filepool-queue.ts
```diff
@@ -0,0 +1,338 @@
...
+    }
+
+}
+
+export const CoreFilepoolQueue = makeSingleton(CoreFilepoolQueueService);
+
+/**
+ * File actions.
+ */
+export const enum CoreFilepoolFileActions {
```
dpalou, line 360: This enum is now declared here but it hasn't been removed from the filepool class.

src/core/singletons/file-utils.ts
```diff
@@ -26,6 +26,15 @@ export class CoreFileUtils {
...
     }
 
+    /**
+     * Check if the plugin is available.
+     *
+     * @returns Whether the plugin is available.
+     */
+    static cordovaPluginFileIsAvailable(): boolean {
+        return window.resolveLocalFileSystemURL !== undefined;
+    }
```
dpalou, line 36: I was just wondering if we really need to keep this check in the app. There are several reasons:

- If the app is configured properly, this function should always return true. We should never publish an app where this returns false.
- I think we usually don't do this for other plugins.
- If this function returns false, the `CoreFile` service will throw the following error in all operations: "Error getting device OS to initialize file system". If we remove this function and the file plugin isn't properly installed I guess the error will be something like "cordova.file is undefined", but it will still be an error. And it's an error meant for devs, as I said we shouldn't publish an app with this returning false so end users should never see this error.

So maybe we should just deprecate the isAvailable function in CoreFile and stop using it. What do you think?

dpalou (changes requested):

src/core/singletons/file-utils.ts
crazyserver: I agree with you! :-)

dpalou (approved):

PR #4346, opened 2025-03-11
MOBILE-4783 completion: Don't refresh course if completion isn't used

dpalou:


dpalou:
Closing the PR, please see the details in the tracker issue.

PR #4347, opened 2025-03-11
Mobile 4653

crazyserver:


src/addons/mod/choice/services/choice.ts
```diff
@@ -267,20 +268,26 @@ export class AddonModChoiceProvider {
      * @param options Other options.
      * @returns Promise resolved when the choice is retrieved.
      */
-    getChoice(courseId: number, cmId: number, options: CoreSitesCommonWSOptions = {}): Promise<AddonModChoiceChoice> {
-        return this.getChoiceByDataKey(courseId, 'coursemodule', cmId, options);
```
dpalou, line 271: You should delete the getChoiceByDataKey functon, right?

src/addons/mod/glossary/services/glossary.ts
```diff
@@ -775,13 +776,7 @@ export class AddonModGlossaryProvider {
...
         const glossaries = await this.getCourseGlossaries(courseId, options);
 
-        const glossary = glossaries.find((glossary) => glossary.coursemodule == cmId);
-
-        if (glossary) {
-            return glossary;
-        }
-
-        throw new CoreError(Translate.instant('core.course.modulenotfound'));
+        return CoreCourseUtils.getActivityByCmId(glossaries, cmId);
```
dpalou, line 779: Here in glossary, you can also change getGlossaryById to use getActivityByField.

src/core/features/course/services/course-utils.ts
```diff
@@ -0,0 +1,67 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { CoreError } from '@classes/errors/error';
+import { Translate } from '@singletons';
+
+/**
+ * Singleton with utils functions for course.
+ */
+export class CoreCourseUtils {
```
dpalou, line 21: This class is a singleton, not a service. IMO it should be in a `singletons` folder instead of the `services` folder.

src/core/features/course/services/course-utils.ts
```diff
@@ -0,0 +1,67 @@
...
+     * Get an activity by key. Will throw an error if not found.
+     *
+     * @param activities List of activities.
+     * @param fieldName Field name to search by.
+     * @param value Activity value to match the key.
+     * @returns Activity.
+     */
+    static getActivityByField<T extends CoreCourseModuleData>(
+        activities: T[],
+        fieldName: 'id' | 'coursemodule' | 'context',
```
dpalou, line 50: Do we really want to limit this in here? This function can use any field, not just these ones. E.g. if in workplace or similar we need to use a different field we would need to change this. IMO it's ok to have a limited list in the modules because each module knows the field it uses, but in here maybe it would be better to keep it generic.

dpalou (changes requested):

src/core/features/course/services/course-module-helper.ts
```diff
@@ -0,0 +1,77 @@
...
+     * @param activities List of activities.
+     * @param fieldName Field name to search by.
+     * @param value Activity value to match the key.
+     * @returns Activity.
+     */
+    getActivityByField<T extends Record<J, number>, J extends keyof T>(
+        activities: T[] = [],
+        fieldName: J,
+        value: number,
+    ): T {
```
dpalou, line 56: This function can be used to search using any type of field, so IMO it would be better to declare value as number|string|boolean instead of just number. Also, the Record value should be unknown IMO because it can also contain objects or arrays inside, even if we don't use them to search.

Also, for the key you can use a name that better identifies what it is instead of using just "J" but this is optional. 

```suggestion
    getActivityByField<T extends Record<string, unknown>, FieldName extends keyof T>(
        activities: T[] = [],
        fieldName: FieldName,
        value: number|string|boolean,
    ): T {
```

src/core/features/filter/services/filter.ts
```diff
@@ -248,7 +248,7 @@ export class CoreFilterProvider {
             options.filter = false;
         }
 
-        if (options.filter) {
+        if (options.filter && filters?.length) {
```
dpalou, line 251: If this can cause problems maybe it should be handled by the `filterText` function to avoid having the same problem in other places. E.g. if the parameter filters is empty, don't do anything else.

One thing that concerns me is that the `filterText` function also removes <nolink> tags at the end (this code was extracted from LMS). Maybe we should do it even if there are no filters?

dpalou (changes requested):

dpalou (approved):

PR #4353, opened 2025-03-14
MOBILE-4584 H5P: Autoplay embedded H5P when downloaded or outdated

crazyserver:


src/core/features/h5p/components/h5p-player/core-h5p-player.html
```diff
@@ -1,13 +1,16 @@
-<div *ngIf="!showPackage && urlParams" class="core-h5p-placeholder">
-    <ion-button class="core-h5p-placeholder-play-button" fill="clear" (click)="play($event)" [ariaLabel]="'core.h5p.play' | translate">
-        <ion-icon name="far-circle-play" slot="icon-only" aria-hidden="true" />
-    </ion-button>
+@if (showPackage) {
+    <core-h5p-iframe *ngIf="showPackage" [fileUrl]="urlParams!.url" [displayOptions]="displayOptions" [onlinePlayerUrl]="src"
```
dpalou, line 2: The ngIf is no longer needed.

Also, the "!" in urlParams is not needed either because `urlParams` is initialized to an empty object, but maybe we want to change that (I'll add another comment about that). In case urlParams can be undefined, this "!" can be changed to "?".

src/core/features/h5p/components/h5p-player/core-h5p-player.html
```diff
@@ -1,13 +1,16 @@
...
+        [component]="component" [componentId]="componentId" [fileTimemodified]="fileTimemodified" />
+} @else if (urlParams) {
```
dpalou, line 4: We used to check `urlParams` when this template was implemented, but that isn't valid anymore because 3 years ago we changed the code and `urlParams` is now initialized to an empty object so this check will always be true. 

Maybe we should check another property to know that the data has been initialized properly? Or maybe we shouldn't initialize urlParams to an empty object, but in that case in the future we might initialize it again by mistake.

src/core/features/h5p/components/h5p-player/h5p-player.ts
```diff
@@ -80,23 +82,34 @@ export class CoreH5PPlayerComponent implements OnInit, OnChanges, OnDestroy {
...
     ngOnChanges(changes: {[name: string]: SimpleChange}): void {
         // If it's already playing there's no need to check if it can be downloaded or auto-played.
         if (changes.src && !this.showPackage) {
-            this.checkCanDownload();
-            if (this.autoPlay) {
+            this.playIfRequested();
+        }
+    }
+
+    protected async playIfRequested(): Promise<void> {
```
dpalou, line 98: This function lacks JSDoc. Also, I don't like the word "Requested" here because it makes me think that the user or someone requested it, but that's not true when the package is downloaded.

src/core/features/h5p/components/h5p-player/h5p-player.ts
```diff
@@ -80,23 +82,34 @@ export class CoreH5PPlayerComponent implements OnInit, OnChanges, OnDestroy {
...
+
+        if (!this.autoPlay) {
+            await this.checkCanDownload();
```
dpalou, line 101: With your change, the function `checkCanDownload` was always called, but now it's only called when `autoPlay` is false. IMO that's wrong, `checkCanDownload` also calculates the `urlParams` and that's always needed.

If we don't care about `state`, `canDownload$` and so on in this case, then maybe we should put the `urlParams` part outside of `checkCanDownload`.

src/core/features/h5p/components/h5p-player/h5p-player.ts
```diff
@@ -80,23 +82,34 @@ export class CoreH5PPlayerComponent implements OnInit, OnChanges, OnDestroy {
...
+
+            if (this.canDownload$.getValue() &&
+                (this.state === DownloadStatus.OUTDATED || this.state === DownloadStatus.DOWNLOADED)) {
```
dpalou, line 104: You can use `CoreFileHelper.isStateDownloaded` here if you want.

dpalou (changes requested):

dpalou (approved):

PR #4362, opened 2025-03-18
MOBILE-4575 user: Sort user initials depending on translation

crazyserver:


src/core/features/user/tests/user-helper.test.ts
```diff
@@ -23,13 +31,53 @@ describe('getUserInitialsFromParts', () => {
...
 
   it('should return initials based on fullname if firstname and lastname are missing', async () => {
-    let parts = {
+    mockSingleton(CoreUser, {
+        getProfile: () => Promise.resolve({
+          firstname: 'John',
+          lastname: 'Doe',
+          fullname: 'John Doe',
+          id: 123,
+    }) });
```
dpalou: Is this mock needed? If I'm not mistaken, getProfile is called when there is no firstname, lastname and no fullname. But in the tests of this scenario there is always a fullname.

If it is needed for some reaason, then the indentation is wrong.

src/core/features/user/tests/user-helper.test.ts
```diff
@@ -12,9 +12,17 @@
...
+import { mockSingleton, mockTranslate } from '@/testing/utils';
 import { CoreUserHelper } from '../services/user-helper';
+import { CoreUser } from '../services/user';
 
 describe('getUserInitialsFromParts', () => {
+  beforeEach(() => {
+    mockTranslate({
+      'core.user.fullnamedisplay': '{{$a.firstname}} {{$a.lastname}}',
+    });
+  });
```
dpalou: Some of the indentation is this file is wrong, it uses 2 spaces instead of 4. Part of it was already wrong. Can you fix the indentation in the whole file?

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -32,7 +32,27 @@ import { CORE_USER_PROFILE_PICTURE_UPDATED } from '@features/user/constants';
 /**
  * Component to display a "user avatar".
  *
- * Example: <core-user-avatar [user]="participant"></core-user-avatar>
+ * Main inputs:
+ * - url: URL of the avatar image. If not provided, initials will be shown.
+ * - userId: User ID. It will be used to link the image to the profile if linkProfile is true.
+ * - siteId: Site ID. It will be used to link image external content. Current site will be used if not provided.
```
dpalou: IMO this comment is hard to understand, specially for people who isn't used to how the app works or what the external-content directive is.

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -128,53 +165,56 @@ export class CoreUserAvatarComponent implements OnInit, OnChanges, OnDestroy {
...
      * @returns boolean
      */
-    isOnline(): boolean {
-        if (!this.user) {
-            return false;
-        }
+    computeIsOnline(): boolean {
+        const user = this.user; // eslint-disable-line @typescript-eslint/no-deprecated
+        const isOnline = this.isOnline ?? user?.isonline;
+        const lastaccess = this.lastAccess ?? user?.lastaccess;
```
dpalou: This variable should use camelCase.

src/addons/competency/pages/competency/competency.html
```diff
@@ -154,30 +154,34 @@ <h2 class="ion-margin-horizontal">{{ 'addon.competency.evidence' | translate }}<
...
-                            <core-user-avatar [user]="evidence.actionuser" slot="start" [linkProfile]="false" />
+                @for (evidence of competency.evidence; track evidence.id) {
+                    <ion-card>
+                        @if (evidence.actionuser) {
+                            @let user = evidence.actionuser;
+                            <ion-item class="ion-text-wrap" core-user-link [userId]="user.id" [courseId]="courseId">
+                                <core-user-avatar [userId]="user.id" [url]="user.profileimageurl" [fullname]="user.fullname"
+                                    slot="start" [linkProfile]="false" />
+                                <ion-label>
+                                    <p class="item-heading">{{ evidence.actionuser.fullname }}</p>
```
dpalou, line 164: You can do user.fullname instead.

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -46,19 +66,33 @@ import { CORE_USER_PROFILE_PICTURE_UPDATED } from '@features/user/constants';
...
+    @Input() isOnline?: boolean;
+    @Input() lastAccess?: number;
+
+    // Deprecated inputs.
+    /** @deprecated since 5.3 Use url instead. */
+    @Input() profileUrl?: string;
+    /** @deprecated since 5.3 Pass data using separate inputs instead. */
+    @Input() site?: CoreSiteBasicInfo | CoreSiteInfo;
+    /** @deprecated since 5.3 Use separate data inputs instead. */
+    @Input() user?: CoreUserWithAvatar;
```
dpalou: As we discussed in chat, this change makes using the component too verbose in most cases. I think it's better to do the opposite change: make the "user" object the required input, and deprecate the inputs to override some input fields. If there is a special case that wants to override one of the fields in the user object, it will need to do it outside of the user-avatar component and then pass the object with the right data.

IMO we should have these inputs: user, linkProfile, courseId, checkOnline, siteId. Remember to add 'initials' to the CoreUserWithAvatar type. In cases where we currently used [site] as the user, you will now need to use the [user] input.

src/core/features/user/services/user.ts
```diff
@@ -382,7 +383,14 @@ export class CoreUserProvider {
         if ('country' in user && user.country) {
             user.country = CoreCountries.getCountryName(user.country);
         }
-        this.storeUser(user.id, user.fullname, user.profileimageurl);
+        user.initials = CoreUserHelper.getUserInitials({
+            firstname: user.firstname,
+            lastname: user.lastname,
+            fullname: user.fullname,
+            initials: user.initials,
+        });
```
dpalou: Since this calculation is now needed in CoreUser, then I think it's better to move the getUserInitials function to the CoreUser service instead of having it in the helper. To avoid having a circular dependency.

src/core/features/user/services/user.ts
```diff
@@ -665,24 +673,31 @@ export class CoreUserProvider {
...
+            initials: record.initials,
         };
 
+        if (!record.fullname || !record.profileimageurl || !record.initials) {
+            const localRecord = await this.getUserFromLocalDb(record.id, siteId);
+
+            record.fullname = record.fullname || localRecord.fullname;
+            record.profileimageurl = record.profileimageurl || localRecord.profileimageurl;
+            record.initials = record.initials || localRecord.initials;
+        }
```
dpalou: I guess this code is to avoid deleting data from DB if we miss some field, right? Another possible option would be to check if there is a local record, and use updateRecords instead of insertRecords in that case. But your code achieves the same, so no need to change it unless you like the updateRecords approach more.

I have 2 comments with this implementation though:

1. If the `record` to store lacks one of these values and the user isn't stored in DB yet, this code will fail. E.g. it's the first time I receive data for user with ID 10 but I don't know the `profileimageurl`. The data won't be stored because `getUserFromLocalDb` will fail.
2. What happens if at some point we want to delete some field info? I guess it won't happen, right? E.g. a user deletes his avatar so we set it as null in DB. I think in that case we also receive a profileimageurl pointing to the default avatar, so I guess we won't have that need, but I'm just saying so you analyze it, just in case.

src/core/features/user/services/user-ws.ts
```diff
@@ -0,0 +1,392 @@
...
+    async updatePicture(
+        userId: number,
+        draftItemId: number,
+        wsOptions: CoreSitesCommonWSOptions = {},
+    ): Promise<CoreUserUpdatePictureWSResponse> {
+        const site = await CoreSites.getSite(wsOptions.siteId);
+
+        const params: CoreUserUpdatePictureWSParams = {
+            draftitemid: draftItemId,
+            delete: !draftItemId,
```
dpalou: This logic is new, before this change delete was always false. Can it happen that draftItemId is 0 and is a valid value? Because in that case we would delete the image when we don't want to.

src/core/features/user/services/user-ws.ts
```diff
@@ -0,0 +1,392 @@
...
+     * @param sortBy Sort by. Can be 'firstname', 'lastname', 'email' or 'siteorder' (default).
+     * @param wsOptions Options to pass to the WS.
+     * @returns Promise resolved when the participants are retrieved.
+     */
+    async getEnrolledUsers(
+        courseId: number,
+        limitFrom = 0,
+        limitNumber = CORE_USER_PARTICIPANTS_LIST_LIMIT,
+        sortBy = 'siteorder',
+        wsOptions: CoreSitesCommonWSOptions = {},
```
dpalou: As we talked in chat, for consistency I think it's better that these new function in WhateverWSService always have 2 parameters: an object with the WS parameters, and an object with the WSOptions. These way we don't mix both things, we have the same signature in all WS methods and we don't need to change the signature if the parameters change for some reason.

dpalou (changes requested):

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -32,7 +32,17 @@ import { CORE_USER_PROFILE_PICTURE_UPDATED } from '@features/user/constants';
 /**
  * Component to display a "user avatar".
  *
- * Example: <core-user-avatar [user]="participant"></core-user-avatar>
+ * - user: User data. It can be used to get the avatar URL and full name if not provided separately.
```
dpalou: IMO this comment is wrong, isn't it? Because now we don't want them to be provided separately.

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -44,138 +54,143 @@ import { CORE_USER_PROFILE_PICTURE_UPDATED } from '@features/user/constants';
...
-            CORE_USER_PROFILE_PICTURE_UPDATED,
-            (data) => {
-                if (data.userId === this.userId) {
-                    this.avatarUrl = data.picture;
-                }
-            },
-            CoreSites.getCurrentSiteId(),
-        );
-    }
+    /** @deprecated since 5.3 Use userid in user object instead */
```
dpalou: I would say "Use userid or id in user object instead", since we also accept id.

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -44,138 +54,143 @@ import { CORE_USER_PROFILE_PICTURE_UPDATED } from '@features/user/constants';
...
-                    : this.site.userId)
-                    ?? (await CoreSites.getSite(this.siteId)).getUserId(),
-                fullname: this.site.fullname ?? '',
-                firstname: this.site.firstname ?? '',
-                lastname: this.site.lastname ?? '',
-                profileimageurl: this.site.userpictureurl ?? '',
-            };
-        }
+        // eslint-disable-next-line @typescript-eslint/no-deprecated
+        return user?.userid || this.userId() || user?.id || // Take always userid from user first.
```
dpalou: this.userId() used to take priority over the rest. IMO we should keep it that way to avoid changing behaviour.

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -44,138 +54,143 @@ import { CORE_USER_PROFILE_PICTURE_UPDATED } from '@features/user/constants';
...
-    /**
-     * Avatar image loading handler.
-     *
-     * @param success Whether the image was loaded successfully or not.
-     */
-    imageLoaded(success: boolean): void {
-        this.imageError = !success;
-    }
+        // eslint-disable-next-line @typescript-eslint/no-deprecated
+        return user?.fullname ?? user?.userfullname ?? this.fullname() ?? this.site()?.fullname ?? '';
```
dpalou: As above, this.fullname() used to take priority over the rest, this can change the behaviour.

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -44,138 +54,143 @@ import { CORE_USER_PROFILE_PICTURE_UPDATED } from '@features/user/constants';
...
-    protected async setFields(): Promise<void> {
-        const profileUrl = this.profileUrl || this.user?.profileimageurl || this.user?.userprofileimageurl ||
-            this.user?.userpictureurl || this.user?.profileimageurlsmall || this.user?.urls?.profileimage;
+    /** @deprecated since 5.3 Use profileimageurl in user object instead */
+    readonly profileUrl = input<string>();
 
-        if (typeof profileUrl === 'string') {
-            this.avatarUrl = profileUrl;
-        }
+    readonly url = linkedSignal<string | undefined>(() => {
```
dpalou: Since this component also links to the user profile, naming it just "url" could be confused with the user profile URL instead of the image URL. Maybe we should name it "avatarUrl" or "imageUrl"?

src/core/components/user-avatar/user-avatar.ts
```diff
@@ -44,138 +54,143 @@ import { CORE_USER_PROFILE_PICTURE_UPDATED } from '@features/user/constants';
...
+        const user = this.user();
 
-        this.fullname = this.fullname || this.user?.fullname || this.user?.userfullname;
+        const profileUrl = user?.profileimageurl || user?.userprofileimageurl ||
+            user?.userpictureurl || user?.profileimageurlsmall || user?.urls?.profileimage ||
+            // eslint-disable-next-line @typescript-eslint/no-deprecated
+            this.profileUrl() || this.site()?.userpictureurl;
```
dpalou: Again, to avoid breaking functionality, I think this.profileUrl() should be the first.

src/addons/notifications/pages/notification/notification.ts
```diff
@@ -59,6 +60,7 @@ export default class AddonNotificationsNotificationPage implements OnInit, OnDes
     protected contextUrl?: string;
     protected courseId?: number;
     protected actionsData?: Record<string, string|number>; // Extra data to handle the URL.
+    protected user?: CoreUserWithAvatar; // User who sent the notification.
```
dpalou: Is this being used?

src/core/features/user/services/user-helper.ts
```diff
@@ -70,40 +70,14 @@ export class CoreUserHelperProvider {
...
-                if (split.length > 1) {
-                    parts.lastname = split[split.length - 1];
-                }
-            }
-        }
-
-        const initials = availableFieldsSorted.reduce((initials, fieldName) =>
-            initials + (parts[fieldName]?.charAt(0) ?? ''), '');
+    async getUserInitialsFromParts(userParts: CoreUserNameParts & { userId?: number }): Promise<string> {
+        const { CoreUser } = await import('./user');
```
dpalou: CoreUser was already imported in the header. Why did you remove it and put a dynamic import here?

My guess is that it's because it's a deprecated function. But in this specific case, since it's quite normal that the helper services import the "standard" service of the same component, I don't think it's a big deal to keep it as a normal import. I guess it doesn't add a new circular dependency that wasn't already there, right?

src/core/features/user/services/user.ts
```diff
@@ -665,25 +672,24 @@ export class CoreUserProvider {
...
+            id: record.id,
+            fullname: record.fullname,
+            profileimageurl: record.profileimageurl,
+            initials: record.initials,
         };
 
-        await site.getDb().insertRecord(USERS_TABLE_NAME, userRecord);
+        const db = await CoreSites.getSiteDb(siteId);
+
+        await db.updateRecords(USERS_TABLE_NAME, userRecord, { id: userRecord.id });
```
dpalou: If a user doesn't exist, I think this statement won't do anything, right? Because there won't be any record to update with that id. So if I'm not mistaken, if a user installs the app from a scratch or adds a new site I think no users will be stored in the DB, ever.

dpalou (changes requested):

dpalou (approved):

PR #4363, opened 2025-03-18
Mobile 4653

crazyserver:


src/core/features/course/services/course-download-status-helper.ts
```diff
@@ -0,0 +1,244 @@
...
+     * @see https://www.typescriptlang.org/docs/handbook/declaration-merging.html#module-augmentation
+     */
+    export interface CoreEventsData {
+        [COURSE_STATUS_CHANGED_EVENT]: CoreEventCourseStatusChanged;
+    }
+
+}
+
+/**
+ * Service that course status features.
```
dpalou, line 44: You lack some word here. Maybe "Service that **contains** ..."?

src/core/features/courses/services/courses.ts
```diff
@@ -230,6 +234,7 @@ export class CoreCoursesProvider {
      *
      * @param siteId Site Id. If not defined, use current site.
      * @returns Promise resolved with true if disabled, rejected or resolved with false otherwise.
+     * @deprecated since 5.0. Use isDownloadCourseDisabledInSite instead.
```
dpalou, line 237: This is a pattern we use in a lot of places for isEnabled/isDisabled (having an async function that accepts a siteId and a sync one that accepts a site instance). So I wouldn't deprecate this. The same applies for other similar functions in this file that you deprecated.

If we want to change this pattern we should do it in the whole app. And IMO it's not a big deal to have 2 functions, even if in some cases we only use one of them.

src/core/features/courses/services/courses.ts
```diff
@@ -374,12 +383,9 @@ export class CoreCoursesProvider {
      * @param ids List of IDs of the courses to get.
      * @param siteId Site to get the courses from. If not defined, use current site.
      * @returns Promise resolved with the courses.
+     * @deprecated since 5.0. Use getCourses instead. This function will be protected.
```
dpalou, line 386: I don't understand this comment. You say we should use getCourses, but getCourses is this function that is now deprecated.

dpalou (changes requested):

dpalou (approved):

PR #4368, opened 2025-03-25
Mobile 4653

crazyserver:


src/core/singletons/mimetype.ts
```diff
@@ -0,0 +1,618 @@
...
+}
+
+/**
+ * "Utils" service with helper functions for mimetypes and extensions.
+ */
+export class CoreMimetype {
+
+    protected static readonly EXTENSION_REGEX = /^[a-z0-9]+$/;
+
+    protected static logger= CoreLogger.getInstance('CoreMimetype');
```
dpalou, line 51: Missing space between variable name and =

src/core/singletons/mimetype.ts
```diff
@@ -0,0 +1,618 @@
...
+    protected static extToMime: Record<string, MimeTypeInfo> = extToMime;
+    protected static mimeToExt: Record<string, string[]> = mimeToExt;
```
dpalou, line 53: The extToMime and mimeToExt variables are protected and you aren't transforming the original variables, so I guess you can just use the original imported variables.

src/core/services/sites.ts
```diff
@@ -1423,14 +1407,13 @@ export class CoreSitesProvider {
      *
      * @param siteId Site id.
      * @returns Site.
+     * @deprecated since 5.0. Use getSite instead.
      */
     async getSiteFromDB(siteId: string): Promise<CoreSite> {
```
dpalou, line 1412: This function is used in our private code, and according to the JSDoc, the purpose is to get the data from DB without any optimizations, so I'm not sure if it's a good idea to use getSite instead. I guess it was implemented like this for a reason.

dpalou (changes requested):

src/core/services/sites.ts
crazyserver: That's fine. I though to use directly the code inside the funtion in the code of the other function. If not, I think this function should be protected, What do you think?

src/addons/mod/lesson/services/handlers/report-link.ts
```diff
@@ -91,17 +90,9 @@ export class AddonModLessonReportLinkHandlerService extends CoreContentLinksHand
...
-                undefined,
-                undefined,
-                false,
-                false,
-                siteId,
-            );
+            const module = await CoreCourseModuleHelper.getModuleNavigationInfo(moduleId, undefined, siteId);
 
             const params = {
-                module,
```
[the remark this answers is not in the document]
crazyserver: Ops, you're true, sorry

src/core/features/course/services/course.ts
```diff
@@ -585,14 +582,16 @@ export class CoreCourseProvider {
...
-            }
-
-            foundModule = section.modules.find((module) => module.id == moduleId);
+            foundModule = section.modules.find((module) => {
+                CoreCourseModuleHelper.storeModuleIndexRecord({
+                    id: module.id,
+                    modname: module.modname,
+                    instance: module.instance,
+                    course: courseId!,
+                }, siteId);
```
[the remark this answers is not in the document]
crazyserver: I've changed a bit the code to:
1. If only a record is inserted it will use insert instead insert or replace
2. When inserting multiple records, a transaction is used now. (we can decide to limit the number of inserts per trassactions).

dpalou:
Just for the record, in the end we decided to remove from this PullRequest the commit to store the activities IDs in database. The reason is that right now there's no way to differentiate if the data received by `site.read` comes from WebService or from the cache, so this could cause a lot of writes in the DB that could make the performance worse when opening courses or activities.

We'll look at it again in the 5.1 release.

dpalou (approved):

PR #4377, opened 2025-03-31
Mobile 4692 forum: Add warning messages

alfonso-salces:


src/addons/mod/forum/components/index/index.html
```diff
@@ -67,6 +67,45 @@
...
+                    <ion-label>
+                        @if (userGroups.length > 0) {
+                            <span>{{ 'addon.mod_forum.cannotadddiscussionall' | translate }}</span>
+                        } @else {
+                            <span>{{ 'addon.mod_forum.cannotadddiscussiongroup' | translate }}</span>
+                        }
+                    </ion-label>
+                </ion-item>
+            </ion-card>
+        }
```
dpalou, line 107: This block logic seems wrong. To begin with, this page already has a warning about groups (`usesGroups && canAddDiscussion && !canAddDiscussionToGroup`), so now we could have 2 warnings at the same time about groups. IMO you should have a single `if` handling all the groups warnings, with several cases.

On the other hand, this logic doesn't seem equivalent to the Moodle LMS logic. Please take a look at the function `get_notifications` in the `discussion_list` class in LMS and try to replicate the logic. Please notice this logic is modified in MDL-83558, but it hasn't been integrated yet. I guess we can use the new logic, but it hasn't been integrated so it could change, maybe you should confirm it with Juan to decide if we want to replicate the new logic or not.

I think you should have most of the information needed to replicate the logic, e.g. for `$capabilitymanager->can_access_all_groups` you have `this.groupInfo.canAccessAllGroups`. The part I think we don't know is `can_post_to_my_groups`, but maybe we can assume it using other properties like `allPartsPermissions`, I'm not sure about that.

Right now I don't have time to look at all the logic and I'll be off for a few days, so please take a look at this to see if you can improve it.

dpalou (changes requested):

src/addons/mod/forum/components/index/index.html
alfonso-salces: As we discussed with Juan, we will include these changes in another issue.

dpalou (approved):

PR #4388, opened 2025-04-07
MOBILE-4653 choice: Add closed and open choice tests

crazyserver:


src/addons/block/searchforums/tests/behat/block_searchforums.feature
```diff
@@ -0,0 +1,50 @@
+@addon_block_search_forums @app @block @block_search_forums @mod_forum @javascript
```
dpalou, line 1: Shouldn't this test be in a different commit? The commit name is "Add closed and open choice tests", and this file has nothing to do with that.

src/addons/block/searchforums/tests/behat/block_searchforums.feature
```diff
@@ -0,0 +1,50 @@
...
+      | activity | name         | intro               | course | idnumber |
+      | forum    | Test forum 1 | Test forum 1 intro  | C1     | forum1   |
+    And the following "mod_forum > discussions" exist:
+      | forum   | name                 | subject              | message                      |
+      | forum1  | Initial discussion 1 | Initial discussion 1 | Initial discussion message 1 |
+    And the following "blocks" exist:
+      | blockname     | contextlevel | reference | pagetypepattern | defaultregion |
+      | search_forums | System       | 1         | site-index      | side-pre      |
+
+  # The tests does not check if the search is correct.
```
dpalou, line 32: Since you're already doing a search in the test I guess it shouldn't be hard to search a valid word, right? I know we might replicate a test already done in forum, but since it's just a few steps IMO it shouldn't be a big deal.

src/addons/mod/choice/tests/behat/choice_availability.feature
```diff
@@ -0,0 +1,107 @@
...
+      | teacher1 | Teacher | 1 | teacher1@example.com |
+      | student1 | Student | 1 | student1@example.com |
+    And the following "courses" exist:
+      | fullname | shortname | category |
+      | Course 1 | C1 | 0 |
+    And the following "course enrolments" exist:
+      | user | course | role |
+      | teacher1 | C1 | editingteacher |
+      | student1 | C1 | student |
+    And I log in as "teacher1"
```
dpalou, line 19: IMO this line should be after the generators, before "I am on Course 1 course homepage", otherwise it's easier to miss it.

src/addons/mod/choice/tests/behat/choice_availability.feature
```diff
@@ -0,0 +1,107 @@
...
+      | timeopen[year] | 2037 |
+      | timeopen[hour] | 00 |
+      | timeopen[minute] | 00 |
+    And I press "Save and display"
+    When I entered the choice activity "Choice name" on course "Course 1" as "student1" in the app
+    Then I should not find "Option 1" in the app
+    And I should not find "Option 2" in the app
+    And I should not find "Save my choice" in the app
+    And I should not find "The results are not currently viewable" in the app
+    And the UI should match the snapshot
```
dpalou, line 41: Maybe we should also check that the open time is displayed? I know that's done by a reusable component not specific to choice, but maybe in the future we change some info and then we use the component in a wrong way. The snapshot also checks it, but snapshots don't work in some environments like Jenkins so it's better to search it in the DOM too.

This applies to other tests too, not just this one.

src/addons/mod/choice/tests/behat/choice_availability.feature
```diff
@@ -0,0 +1,107 @@
...
+      | timeopen[day] | 30 |
+      | timeopen[month] | December |
+      | timeopen[year] | 2007 |
+      | timeopen[hour] | 00 |
+      | timeopen[minute] | 00 |
+    And I press "Save and display"
+    When I entered the choice activity "Choice name" on course "Course 1" as "student1" in the app
+    Then I should find "Option 1" in the app
+    And I should find "Option 2" in the app
+    And I should find "Save my choice" in the app
```
dpalou, line 72: Since "Save my choice" is a button I guess it's better to use "I should be able to press". When the button is not there it's better to use "find" as you're already doing.

src/addons/mod/choice/tests/behat/choice_availability.feature
```diff
@@ -0,0 +1,107 @@
...
+  Scenario: Enable the choice activity with a end deadline in the future
+    Given I set the following fields to these values:
+      | timeclose[enabled] | 1 |
+      | timeclose[day] | 30 |
+      | timeclose[month] | December |
+      | timeclose[year] | 2037 |
+      | timeclose[hour] | 00 |
+      | timeclose[minute] | 00 |
+    And I press "Save and display"
+    When I log in as "student1"
```
dpalou, line 85: This line isn't needed because you're already specifying the user in the next step.

dpalou (changes requested):

src/addons/block/searchforums/tests/behat/block_searchforums.feature
crazyserver: ¿? I think your confused. There are 2 commits here

src/addons/block/searchforums/tests/behat/block_searchforums.feature
crazyserver: I considered this way because it implies telling the expected results in the tests..

Something like:
Given global search expects the query "message" and will return:
      | type     | idnumber |
      | activity | forum1   |
      | activity | forum2   |
      
So I think this is not necessary here.

src/addons/mod/choice/tests/behat/choice_availability.feature
crazyserver: Since the button is disabled in the begining I decided to test the button.

dpalou (approved):

PR #4391, opened 2025-04-09
Mobile 4653

crazyserver:


src/addons/block/courselist/tests/behat/block_courselist.feature
```diff
@@ -0,0 +1,50 @@
...
+  Scenario: View and navigate the courses block in a course
+    Given I entered the course "Course 1" as "teacher1" in the app
+    When I press "Open block drawer" in the app
+    Then I should find "My courses" in the app
+    When I press "My courses" in the app
+    Then the header should be "My courses" in the app
+    Then I should find "Course 1" in the app
+    Then I should find "Course 2" in the app
+    Then I should find "Course 3" in the app
+    Then I should not find "Course 4" in the app
```
dpalou, line 39: These Then should be And.

src/addons/block/globalsearch/tests/behat/block_globalsearch.feature
```diff
@@ -0,0 +1,39 @@
...
+      | blockname     | contextlevel | reference | pagetypepattern | defaultregion |
+      | globalsearch   | Course       | C1        | course-view-*   | side-pre      |
+
+  Scenario: View and navigate the global search block in a course
+    Given I entered the course "Course 1" as "student1" in the app
+    When I press "Open block drawer" in the app
+    Then I should find "Search" in the app
+    When I press "Search" in the app
+    Then the header should be "Search" in the app
+    And I set the field "Search" to "Moodle" in the app
```
dpalou, line 26: This And should be When.

src/addons/block/glossaryrandom/tests/behat/block_glossaryrandom.feature
```diff
@@ -0,0 +1,66 @@
...
+      | Title                           | Animal of the day   |
+      | Take entries from this glossary | Animals             |
+      | How a new entry is chosen       | Last modified entry |
+    And I press "Save changes"
+
+  Scenario: View and navigate the glossay random block in a course
+    Given I entered the course "Course 1" as "student1" in the app
+    When I press "Open block drawer" in the app
+    Then I should find "Animal of the day" in the app
+    And I should find "Zebra" in the app
```
dpalou, line 38: In here I would also test that the other animals aren't found.

src/addons/block/glossaryrandom/tests/behat/block_glossaryrandom.feature
```diff
@@ -0,0 +1,66 @@
...
+    And I press "Add a new entry" in the app
+    Then the header should be "Animals" in the app
+    When I set the following fields to these values in the app:
+      | Concept    | Donkey |
+      | Definition | Esel   |
+    And I press "Save" in the app
+    And I press "Animals" in the app
+    And I pull to refresh in the app
+    Then I should find "Aardvark" in the app
+    And I should find "Donkey" in the app
```
dpalou, line 55: Wouldn't it be better to do a PTR in the block and check that now it displays "Donkey"?

src/core/services/sites.ts
```diff
@@ -1409,8 +1409,11 @@ export class CoreSitesProvider {
      * @returns Site.
      */
     protected async getSiteFromDB(siteId: string): Promise<CoreSite> {
+        const db = CoreAppDB.getDB();
+
         try {
-            const record = await this.sitesTable.getOneByPrimaryKey({ id: siteId });
+            // For some unknown reason we cannot rely on sitesTable.getOneByPrimaryKey directly here.
```
dpalou, line 1415: I would change the comments. First, in the function JSDoc, I would put something like:

```
Get a site directly from the database, without using any optimizations.
This function is used to read the site from DB during the app initialization, to avoid deadlocks.
```

And then, in this line, something like:

`Do not use sitesTable.getOneByPrimaryKey in here, it can cause a deadlock if this function is used during the app initialization.`

dpalou (changes requested):

dpalou (approved):

PR #4435, opened 2025-05-21
Mobile 4653

crazyserver:


src/core/directives/reading-mode.ts
```diff
@@ -140,34 +137,40 @@ export class CoreReadingModeDirective implements AfterViewInit, OnDestroy {
...
-        // Rename style attributes on DOM elements.
-        this.renamedStyles = Array.from(this.element.querySelectorAll('*[style]'));
-        this.renamedStyles.forEach((element: HTMLElement) => {
-            this.renamedStyles.push(element);
-            element.setAttribute('data-original-style', element.getAttribute('style') || '');
-            element.removeAttribute('style');
-        });
+        elements.forEach((element: HTMLElement) => {
+            // Disable all styles in element.
+            this.disabledStyles = Array.from(element.querySelectorAll('style:not(disabled)'));
```
dpalou, line 144: `elements` is an array, in theory you can have more than one element. But with this code, you will only store the `disabledStyles` of the last element (and the same happens for `renamedStyles`). You should probably create an empty array before the forEach, and then concatenate the styles found for each element.

dpalou (changes requested):

dpalou (approved):

PR #4492, opened 2025-07-08
Mobile 4739 - Update to Angular 20

crazyserver:


UPGRADE.md
```diff
@@ -2,6 +2,11 @@ This file describes API changes in the Moodle App that affect site plugins, info
 
 For more information about upgrading, read the official documentation: https://moodledev.io/general/app/upgrading/
 
+5.1.0
+=====
+
+ - Support for two-way data binding has been removed for site plugins because modelWithInjectionContext is not compatible on Angular 18.
```
dpalou: Here I would be more specific and say that "model" was removed, otherwise site plugin devs might wonder what does this mean in practice. Also, even if it's true that modelWithInjectionContext is not compatible with Angular 18, the main reason to remove it is that site plugins won't need it. So I would say something like:

"model" has been removed for site plugins because it isn't needed and it's not compatible with Angular 18. "model" is meant to support 2-way data binding in custom components, and site plugins cannot create components.

src/addons/filter/activitynames/activitynames.module.ts
```diff
@@ -23,11 +23,9 @@ import { AddonFilterActivityNamesHandler } from './services/handlers/activitynam
...
     ],
     providers: [
-        {
-            provide: APP_INITIALIZER,
-            multi: true,
-            useValue: () => CoreFilterDelegate.registerHandler(AddonFilterActivityNamesHandler.instance),
-        },
+        provideAppInitializer(() => {
+            CoreFilterDelegate.registerHandler(AddonFilterActivityNamesHandler.instance);
+    }),
```
dpalou: Wrong indentation here, and also in the filters multilang, multilang2, tex.

package.json
```diff
@@ -170,12 +170,12 @@
         "terser-webpack-plugin": "^5.3.14",
         "ts-jest": "^29.2.6",
         "ts-node": "^8.10.2",
-        "typescript": "~5.5.4",
+        "typescript": "~5.8.3",
```
dpalou, line 172: We should update typescript in cordova-plugin-moodleapp/package.json too.

src/core/components/tabs-outlet/tabs-outlet.ts
```diff
@@ -254,3 +256,36 @@ export type CoreTabsOutletTab = CoreTabBase & {
 };
 
 export type CoreTabsOutletTabWithId = Omit<CoreTabsOutletTab, 'id'> & { id: string };
+
+/**
+ * Imported from: @ionic/angular/common/directives/navigation/stack-utils
```
dpalou: I would say "Copied" instead of "Imported". We want to import, but we can't.

src/core/services/navigator.ts
```diff
@@ -842,3 +849,18 @@ export class CoreNavigatorService {
 }
 
 export const CoreNavigator = makeSingleton(CoreNavigatorService);
+
+/**
+ * Imported from: @ionic/angular/common/providers/nav-controller
+ * because the import of AnimationOptions was not working.
```
dpalou: I understand that the import that didn't work was "NavigationOptions", not "AnimationOptions", right?

src/addons/mod/forum/components/index/index.ts
```diff
@@ -86,12 +84,14 @@ import { CoreCourseModuleInfoComponent } from '@features/course/components/modul
...
 
+    route = inject(ActivatedRoute);
+
     @ViewChild(CoreSplitViewComponent) splitView!: CoreSplitViewComponent;
 
     component = ADDON_MOD_FORUM_COMPONENT_LEGACY;
     pluginName = 'forum';
     descriptionNote?: string;
-    promisedDiscussions: CorePromisedValue<AddonModForumDiscussionsManager>;
+    promisedDiscussions: CorePromisedValue<AddonModForumDiscussionsManager> = new CorePromisedValue();
```
dpalou: IMO it would be better to use:

`promisedDiscussions = new CorePromisedValue<AddonModForumDiscussionsManager>();`

This way the type is inferred from the value.

src/addons/mod/forum/pages/discussion/discussion.ts
```diff
@@ -85,6 +85,11 @@ type Post = AddonModForumPost & { children?: Post[] };
 })
 export default class AddonModForumDiscussionPage implements OnInit, AfterViewInit, OnDestroy, CanLeave {
 
+    protected splitView = inject(CoreSplitViewComponent, { optional: true });
+    protected element: HTMLElement = inject(ElementRef).nativeElement;
+    protected route = inject(ActivatedRoute);
+    protected courseContentsPage? = inject(CoreCourseContentsPage, { optional: true });
```
dpalou: I think the "?" at the end of the name isn't needed, right?

It also happens in the new-discussion page and others, please search "courseContentsPage?".

src/core/components/style/style.ts
```diff
@@ -33,21 +33,21 @@ import { CoreDom } from '@singletons/dom';
 })
 export class CoreStyleComponent implements OnChanges {
 
-    @Input() css = ''; // CSS rules.
-    @Input() prefix = ''; // Prefix to add to CSS rules.
+    protected element: HTMLElement = inject(ElementRef).nativeElement;
```
dpalou: I see you added the new inject at the beginning of components. For components I usually prefer putting Input and Output at the beginning because it's the "public API", it's what affects things outside of the component, while in this case the "element" is something internal. What do you think?

src/core/components/swipe-slides/swipe-slides.ts
```diff
@@ -52,6 +53,8 @@ import { CoreBaseModule } from '@/core/base.module';
 })
 export class CoreSwipeSlidesComponent<Item = unknown> implements OnChanges, OnDestroy, AsyncDirective {
 
+    protected content? = inject(IonContent);
```
dpalou: I think the "?" isn't needed, right?

src/core/features/compile/components/compile-html/compile-html.ts
```diff
@@ -96,12 +100,10 @@ export class CoreCompileHtmlComponent implements OnChanges, OnDestroy, DoCheck {
...
 
-    constructor(
-        protected changeDetector: ChangeDetectorRef,
-        protected injector: Injector,
-        element: ElementRef,
-        differs: KeyValueDiffers,
-    ) {
+    constructor() {
+        const element = inject(ElementRef);
+        const differs = inject(KeyValueDiffers);
```
dpalou: In most places you're using directly:

protected element: HTMLElement = inject(ElementRef).nativeElement;

Why aren't you doing the same here? "element" isn't used for anything else. And maybe you can do the same for the differs.

src/core/features/editor/components/rich-text-editor/rich-text-editor.ts
```diff
@@ -101,13 +104,10 @@ export class CoreEditorRichTextEditorComponent implements AfterViewInit, OnDestr
...
 
-    constructor(
-        @Optional() protected content: IonContent,
-        elementRef: ElementRef<HTMLElement>,
-    ) {
+    constructor() {
          // Generate a "unique" ID based on timestamp.
         this.pageInstance = `app_${Date.now()}`;
-        this.element = elementRef.nativeElement;
+        this.element = inject(ElementRef).nativeElement;
```
dpalou: In most places you put the inject when declaring the variable, I would do the same here for consistency.

src/core/features/fileuploader/components/audio-histogram/audio-histogram.ts
```diff
@@ -35,16 +35,12 @@ export class CoreFileUploaderAudioHistogramComponent implements AfterViewInit, O
     @Input({ transform: toBoolean }) paused = false;
     @ViewChild('canvas') canvasRef?: ElementRef<HTMLCanvasElement>;
 
-    private element: HTMLElement;
+    private element: HTMLElement = inject<ElementRef<HTMLElement>>(ElementRef).nativeElement;
```
dpalou: Since you're specifying the type of ElementRef on the right side, there's no need to manually specify that the element has type HTMLElement on the left side. Just one of them is enough, the one you prefer.

I usually prefer specifying the type on the right side if possible, but it's more verbose in this case and in most places you do it in the left side for ElementRef, so I'm fine with it too.

src/core/features/h5p/components/h5p-player/h5p-player.ts
```diff
@@ -68,10 +68,8 @@ export class CoreH5PPlayerComponent implements OnInit, OnChanges, OnDestroy {
     protected logger: CoreLogger;
     protected nativeElement: HTMLElement;
 
-    constructor(
-        elementRef: ElementRef,
-    ) {
-        this.nativeElement = elementRef.nativeElement;
+    constructor() {
+        this.nativeElement = inject(ElementRef).nativeElement;
```
dpalou: As mentioned in another comment, usually you put this when declaring the variable.

src/core/features/question/classes/base-question-component.ts
```diff
@@ -58,9 +58,10 @@ export class CoreQuestionBaseComponent<T extends AddonModQuizQuestion = AddonMod
     protected hostElement: HTMLElement;
     protected onReadyPromise = new CorePromisedValue<void>();
 
-    constructor(@Optional() @Inject('') logName: string, elementRef: ElementRef) {
-        this.logger = CoreLogger.getInstance(logName);
-        this.hostElement = elementRef.nativeElement;
+    constructor() {
+        this.logger = CoreLogger.getInstance(this.constructor.name);
+
+        this.hostElement = inject(ElementRef).nativeElement;
```
dpalou: As mentioned in another comment, you usually put this when declaring the variable.

src/core/features/siteplugins/classes/call-ws-directive.ts
```diff
@@ -40,11 +42,8 @@ export class CoreSitePluginsCallWSBaseDirective implements OnInit, OnDestroy {
     protected element: HTMLElement;
     protected invalidateObserver?: Subscription;
 
-    constructor(
-        element: ElementRef,
-        protected parentContent: CoreSitePluginsPluginContentComponent | null,
-    ) {
-        this.element = element.nativeElement || element;
+    constructor() {
+        this.element = inject(ElementRef).nativeElement;
```
dpalou: As mentioned in another comment, you usually put this when declaring the variable.

src/core/features/siteplugins/components/module-index/module-index.ts
```diff
@@ -79,9 +81,7 @@ export class CoreSitePluginsModuleIndexComponent implements OnInit, OnDestroy, C
     ptrEnabled = true;
     isDestroyed = false;
 
-    jsData?: Record<string, unknown>; // Data to pass to the component.
```
dpalou, line 83: I guess you deleted this comment by mistake.

src/core/features/siteplugins/directives/new-content.ts
```diff
@@ -67,11 +69,8 @@ export class CoreSitePluginsNewContentDirective implements OnInit {
 
     protected element: HTMLElement;
 
-    constructor(
-        element: ElementRef,
-        @Optional() protected parentContent: CoreSitePluginsPluginContentComponent,
-    ) {
-        this.element = element.nativeElement || element;
+    constructor() {
+        this.element = inject(ElementRef).nativeElement;
```
dpalou: As mentioned in another comment, you usually put this when declaring the variable.

src/core/features/usertours/components/user-tour/user-tour.ts
```diff
@@ -85,10 +86,10 @@ export class CoreUserToursUserTourComponent implements AfterViewInit, OnDestroy
     protected content?: HTMLIonContentElement | null;
     protected lastActivatedTime = 0;
 
-    constructor({ nativeElement: element }: ElementRef<HTMLElement>) {
-        this.element = element;
+    constructor() {
+        this.element = inject(ElementRef).nativeElement;
```
dpalou: As mentioned in another comment, you usually put this when declaring the variable.

package.json
```diff
@@ -170,12 +170,12 @@
...
         "typescript-eslint": "^8.34.1",
         "vinyl": "^3.0.0"
     },
     "engines": {
-        "node": ">=v20.18 <21"
+        "node": ">=v22.17 <25"
```
dpalou: Should this be <25 or <23?

src/core/directives/tests/format-text.test.ts
```diff
@@ -33,7 +33,10 @@ describe('CoreFormatTextDirective', () => {
     let config: Partial<RenderConfig>;
 
     beforeEach(() => {
-        mockSingleton(CoreSites, { getSite: () => Promise.reject() });
+        mockSingleton(CoreSites, {
+            getSite: () => Promise.reject(),
+            getCurrentSiteId: () => '',
```
dpalou: I think this change wasn't needed in the end, right?

src/core/singletons/path.ts
```diff
@@ -117,4 +117,45 @@ export class CorePath {
...
+        let currentSegments = currentPath.split('/');
+        // Remove empty segments from the current path.
+        currentSegments = currentSegments.filter((segment) => segment !== '');
+
+        let relativeSegments = relativePath.split('/');
+        relativeSegments = relativeSegments.filter((segment) => segment !== '' && segment !== '.');
+
+        relativeSegments.forEach((segment) => {
+            if (segment === '..') {
+                currentSegments.pop();
```
dpalou, line 149: If the relativePath has too many "..", at some point they will be ignored. If this is ok, then IMO we should put a case in the unit test to make it clear this is expected.

E.g. resolveRelativePath('/foo', '../../../bar') will return just '/bar'.

src/core/features/emulator/components/capture-media/capture-media.html
```diff
@@ -20,17 +20,14 @@ <h1>{{ title | translate }}</h1>
...
-            }
-
-            <!-- Canvas to treat the image and an img to show the result. -->
-            @if (isImage) {
+            } @else if (isImage) {
+                <!-- Canvas to treat the image and an img to show the result. -->
                 <canvas class="core-webcam-image-canvas" #imgCanvas></canvas>
-            }
-            @if (isImage) {
+            } @else if (isImage) {
```
dpalou: You have 2 ifs checking the same in a row.

src/addons/mod/quiz/pages/player/player.ts
```diff
@@ -117,13 +130,7 @@ export default class AddonModQuizPlayerPage implements OnInit, OnDestroy, CanLea
     protected autoSave!: AddonModQuizAutoSave; // Class to auto-save answers every certain time.
     protected autoSaveErrorSubscription?: Subscription; // To be notified when an error happens in auto-save.
     protected forceLeave = false; // If true, don't perform any check when leaving the view.
-    protected reloadNavigation = false; // Whether navigation needs to be reloaded because some data was sent to server.
```
dpalou, line 121: I guess you deleted this comment by mistake.

src/core/features/course/pages/contents/contents.ts
```diff
@@ -90,9 +92,7 @@ export default class CoreCourseContentsPage implements OnInit, OnDestroy, CoreRe
     protected syncObserver?: CoreEventObserver;
     protected isDestroyed = false;
     protected modulesHaveCompletion = false;
-    protected debouncedUpdateCachedCompletion?: () => void; // Update the cached completion after a certain time.
```
dpalou, line 94: I guess you deleted this comment by mistake.

src/core/classes/delegate.ts
```diff
@@ -80,10 +81,11 @@ export class CoreDelegate<HandlerType extends CoreDelegateHandler> {
     /**
      * Constructor of the Delegate.
      *
-     * @param delegateName Delegate name used for logging purposes.
+     * @param delegateName Deprecated since 5.1. Unused, remove parameter.
      */
-    constructor(delegateName: string) {
-        this.logger = CoreLogger.getInstance(delegateName);
+    // eslint-disable-next-line @typescript-eslint/no-unused-vars
+    constructor(delegateName?: string) {
```
dpalou: If we're not going to use delegateName IMO we can just remove it. If a plugin creates an instance passing a name but we don't accept one nothing will happen, it's not a breaking change for plugins. It will only affect TypeScript (custom apps), and they will need to adapt their code.

src/addons/block/sitemainmenu/components/sitemainmenu/addon-block-sitemainmenu.html
```diff
@@ -14,8 +14,8 @@ <h2>{{ 'addon.block_sitemainmenu.pluginname' | translate }}</h2>
                     </ion-label>
                 </ion-item>
             }
-
-            @for (modOrSubsection of mainMenuBlock.contents; track modOrSubsection) {
+            @for (modOrSubsection of mainMenuBlock.contents;
+                track isModule(modOrSubsection) ? `m${modOrSubsection.id}` : `s${modOrSubsection.id}`) {
```
dpalou: Since subsections are actually modules in Moodle, I think the id should always be unique, it probably isn't needed to differentiate between sections and modules. I think that tracking by id should be enough.

This happens in other places of the app too.

src/addons/calendar/components/calendar/addon-calendar-calendar.html
```diff
@@ -107,51 +107,57 @@ <h2 id="addon-calendar-monthname">
...
+                                        <p class="ion-hide-md-up addon-calendar-dot-types">
+                                            @for (type of day.calendareventtypes; track type) {
+                                                <span class="calendar_event_type calendar_event_{{type}}"></span>
+                                            }
+                                        </p>
                                         <!-- In tablet, display list of events. -->
                                         @if (day.filteredEvents) {
                                             <div class="ion-hide-md-down addon-calendar-day-events">
-                                                @for (event of day.filteredEvents | slice:0:4; track event; let index = $index) {
+                                                @for (event of day.filteredEvents | slice:0:4; track $index; let index = $index) {
```
dpalou: Here it's more correct to track by event.id. Although $index will probably work too taking into account how we use it in the app.

src/addons/calendar/components/calendar/addon-calendar-calendar.html
```diff
@@ -76,76 +76,86 @@ <h2 id="addon-calendar-monthname">
...
-                            <ion-col *ngFor="let value of week.prepadding" class="dayblank addon-calendar-day" role="cell" />
-                            <ion-col *ngFor="let day of week.days" class="addon-calendar-day ion-text-center" [ngClass]='{
+                        @for (week of month.weeks; track week) {
+                            <ion-row class="addon-calendar-week" role="row">
+                                <!-- Empty slots (first week). -->
+                                @for (value of week.prepadding; track value) {
+                                    <ion-col class="dayblank addon-calendar-day" role="cell" />
+                                }
+                                @for (day of week.days; track day) {
+<ion-col class="addon-calendar-day ion-text-center" [ngClass]='{
```
dpalou: The indentation of the ngClass looks weird now (maybe it already did before, I'm not sure). Please look at the whole file to see it.

src/addons/calendar/components/calendar/addon-calendar-calendar.html
```diff
@@ -107,51 +107,57 @@ <h2 id="addon-calendar-monthname">
...
-                                                            (ariaButtonClick)="eventClicked(event, $event)" [tabindex]="activeView ? 0 : -1">
-                                                            <span class="calendar_event_type calendar_event_{{event.formattedType}}"></span>
-                                                            @if (event.offline && !event.deleted) {
+                                                        <div class="addon-calendar-event"
+                                                            [class.addon-calendar-event-past]="$any(event).ispast"
+                                                            (ariaButtonClick)="eventClicked($any(event), $event)"
+                                                            [tabindex]="activeView ? 0 : -1">
+                                                            <span class="calendar_event_type calendar_event_{{$any(event).formattedType}}">
+                                                            </span>
+                                                            @if ($any(event).offline && !$any(event).deleted) {
```
dpalou: I don't like these `$any` at all. I don't like that ng-template doesn't know the types and they are `any`, but that's how the app currently works so it was something that we could improve later. But now we're manually casting it to `any` in some cases, and that is quite ugly.

I was looking for information and I saw that [Angular lets you create a directive with ngTemplateContextGuard](https://angular.dev/guide/directives/structural-directives#typing-the-directives-context) to specify the types for ng-template, but looking at the Angular docs it means we would need to create one directive per each ng-template we have in the app, which is quite bothersome. There is another example [in here](https://medium.com/@thomas.laforge/ngtemplateoutlet-type-checking-5d2dcb07a2c6).

In [this SO answer](https://stackoverflow.com/a/68318283/1455909), they create a generic directive to specify the type. But I don't understand how it works, because it does: 

In the template: `[typedTemplate]="typeToken"`
In the component: `typeToken: { param: string };`

But typeToken is a type, and AFAIK you cannot assign types to component properties. So I don't know how it's supposed to be used, maybe I'm missing something obvious.

Another solution I saw was creating our own functions to cast the variable to the right type in template (e.g. asEvent(event)). But it means using a new function in the template, which performs way worse than a variable. I compiled your code and $any was removed from the template for JS, the compile knows it needs to be removed, but our functions wouldn't be removed.

I think we should dedicate some time to see if we can find a way to add type checking to ng-template.

If we want to keep using type `any` for now, another possibility to avoid having to put $any everywhere is to cast the type right after the for. But it's mostly the same:

```
@for (eventUnknown of day.filteredEvents...
@let event = $any(eventUnknown);
```

src/addons/mod/assign/components/index/addon-mod-assign-index.html
```diff
@@ -19,7 +19,7 @@
         [courseId]="courseId" [hasDataToSync]="hasOffline" (completionChanged)="onCompletionChange()">
         @if (assign && assign.introattachments?.length && !assign.submissionattachments) {
             <div description>
-                @for (file of assign.introattachments; track file) {
+                @for (file of assign.introattachments; track $index) {
```
dpalou: When the type is CoreWSExternalFile[] like in this case, we can use fileurl to track the file. I would change it in that case.

But in some other cases the type is CoreWSFile[], which is either CoreWSExternalFile or CoreWSStoredFile, and the latter doesn't have fileurl, it has url. So in that case it's not so straightforward to change the track and can decrease readability. We can chose to leave $index in that case if we think it's better, as long as the array cannot be mutated.

src/addons/notifications/pages/settings/settings.html
```diff
@@ -106,15 +105,16 @@ <h1>{{ 'addon.notifications.notifications' | translate }}</h1>
                             </ion-col>
                             @for (state of ['loggedin', 'loggedoff']; track state) {
                                 <ion-col size="2" class="ion-text-center ion-no-padding">
-                                    @if (preferences!.enableall && notification.processorsByName[currentProcessorName][state]) {
+                                    @if (preferences!.enableall && notification.processorsByName![currentProcessorName][state]) {
```
dpalou: I see you added some "!" in templates, but we try to avoid "!" when possible. It's usually better to use "?" or previously checking that it exists.

src/core/components/attachments/core-attachments.html
```diff
@@ -43,13 +43,13 @@
         }
 
         @for (file of files; track file; let index = $index) {
```
dpalou: Here you're still tracking by an object, and in this case the list can be mutated because users can add or delete attachments, it's important to use a proper track here (not just $index).

PR #4496, opened 2025-07-14
Mobile 4842 - Additional signals

crazyserver:


src/core/services/network.ts
```diff
@@ -284,18 +289,10 @@ export class CoreNetworkService extends Network {
      * Check if device uses a limited connection.
      *
      * @returns Whether the device uses a limited connection.
+     * @deprecated since 5.0. Use isCellular instead.
```
dpalou: This should be 5.1.

src/core/services/network.ts
```diff
@@ -91,18 +97,6 @@ export class CoreNetworkService extends Network {
...
-            (<any> window).Connection = {
-                UNKNOWN: CoreNetworkConnection.UNKNOWN, // eslint-disable-line @typescript-eslint/naming-convention
-                ETHERNET: CoreNetworkConnection.ETHERNET, // eslint-disable-line @typescript-eslint/naming-convention
-                WIFI: CoreNetworkConnection.WIFI, // eslint-disable-line @typescript-eslint/naming-convention
-                CELL_2G: CoreNetworkConnection.CELL_2G, // eslint-disable-line @typescript-eslint/naming-convention
-                CELL_3G: CoreNetworkConnection.CELL_3G, // eslint-disable-line @typescript-eslint/naming-convention
-                CELL_4G: CoreNetworkConnection.CELL_4G, // eslint-disable-line @typescript-eslint/naming-convention
-                CELL: CoreNetworkConnection.CELL, // eslint-disable-line @typescript-eslint/naming-convention
-                NONE: CoreNetworkConnection.NONE, // eslint-disable-line @typescript-eslint/naming-convention
-            };
```
dpalou: This can be a breaking change for plugins using the webapp, since this property of the Cordova plugin API will be available in device but not in webapp.

We can decide to remove it if we think it's better, but then the change should be properly documented.

PR #4502, opened 2025-07-17
Mobile 4820 Avatar crop

crazyserver:


src/core/features/user/pages/about/about.ts
```diff
@@ -183,7 +186,28 @@ export default class CoreUserAboutPage implements OnInit, OnDestroy {
...
+                cssClass: 'core-modal-fullscreen',
+                componentProps: {
+                    image: fileObject,
+                },
+            });
+
+            if (editedImageBlob) {
+                // Override the file entry with the edited image.
+                fileEntry = await CoreFile.writeFile(fileEntry.fullPath, editedImageBlob);
+            } else {
```
dpalou, line 204: IMO it would be nice to detect if the file has been edited or not if it's possible. If the user didn't do any changes then there's no need to override the file.

src/core/features/user/pages/about/about.ts
```diff
@@ -183,7 +186,28 @@ export default class CoreUserAboutPage implements OnInit, OnDestroy {
...
+                return;
+            }
+
+            const result =
+                await CoreFileUploaderHelper.uploadFileEntry(fileEntry, true, maxSize, true, false) as CoreWSUploadFileResult;
```
dpalou: Instead of having to cast the type manually, it would be nice to create function overrides to correctly infer the type based on the "upload" parameter. This means creating overrides for "uploadFileEntry", "uploadFileObject" and "copyAndUploadFile".

This is just a nice to have, we can do it later.

package.json
```diff
@@ -117,6 +117,7 @@
         "jszip": "^3.10.1",
         "mathjax": "3.2.2",
         "mp3-mediarecorder": "4.0.5",
+        "ngx-image-cropper": "^9.1.5",
```
dpalou, line 120: I see that the package-lock has 10k additions and 16k deletions. That seems a lot just to install a library. Can you verify it's ok? Sometimes npm changes the format for no reason, and I cannot see the diff in here.

crazyserver: you're right!

src/core/features/user/pages/about/about.ts
crazyserver: Now it's very difficult it does not change, because the croping is set to square and most images won't be that proportion from the begining. I change inputs to signals.

src/core/features/fileuploader/services/fileuploader-helper.ts
```diff
@@ -698,10 +701,34 @@ export class CoreFileUploaderHelperProvider {
...
+        name?: string,
+    ): Promise<FileEntry>;
+    async uploadFileEntry(
+        fileEntry: FileEntry,
+        deleteAfter: boolean,
+        maxSize: number | undefined,
         upload?: boolean,
         allowOffline?: boolean,
         name?: string,
+    ): Promise<FileEntry>;
```
dpalou: If I'm not mistaken, this one should be:

```suggestion
    ): Promise<CoreWSUploadFileResult | FileEntry>;
```

src/core/features/fileuploader/services/fileuploader-helper.ts
```diff
@@ -727,7 +754,28 @@ export class CoreFileUploaderHelperProvider {
...
+        allowOffline?: boolean,
+        name?: string,
+    ): Promise<FileEntry>;
+        async uploadFileObject(
+        file: IFile | File,
+        maxSize: undefined | number,
+        upload?: boolean,
+        allowOffline?: boolean,
+        name?: string,
+    ): Promise<FileEntry>;
```
dpalou: As above, this one should be CoreWSUploadFileResult | FileEntry. Also the indentation of one of the lines is wrong.

src/core/features/viewer/components/image-edit/image-edit.ts
```diff
@@ -0,0 +1,114 @@
...
+    readonly image = input<IFile>();
+    transform: ImageTransform = {
+        scale: 1,
+        rotate: 0,
+        flipH: false,
+        flipV: false,
+    };
+
+    canvasRotation = 0;
+    changed = false;
```
dpalou: This variable is no longer needed since you cannot detect if the image has changed.

src/core/features/viewer/components/image-edit/image-edit.ts
```diff
@@ -0,0 +1,114 @@
...
+    standalone: true,
+    imports: [
+        CoreSharedModule,
+        ImageCropperComponent,
+    ],
+})
+export class CoreViewerImageEditComponent {
+
+    readonly editTool = viewChild(ImageCropperComponent);
+    readonly image = input<IFile>();
```
dpalou: Initially you defined this input as required, but not anymore. Just mentioning it in case you want to declare it as required (input.required).

src/core/features/viewer/components/image-edit/image-edit.ts
crazyserver: I can check if the image has changed:
I have to check if the original ratio is squared. If it's then the crop can be unchanged, otherwise the image will be cropped to maintain AR 1:1

dpalou (dismissed):

PR #4504, opened 2025-07-18
Mobile 4819 QR code improvements

crazyserver:


src/core/features/viewer/components/qr-scanner/qr-scanner.ts
```diff
@@ -32,15 +32,30 @@ import { ModalController, Translate } from '@singletons';
...
 export class CoreViewerQRScannerComponent implements OnInit, OnDestroy {
 
-    @Input() title?: string; // Page title.
+    readonly title = input(Translate.instant('core.viewer.qrscannertitle'), { transform: (value?: string) => {
+        if (value === undefined || value === null) {
+            return Translate.instant('core.viewer.qrscannertitle');
+        }
+
+        return value;
+    } });
```
dpalou: Having to create a transform just to set a default value is quite ugly and bothersome. I guess this happens because it's an input of a modal, hopefully this will be improved in the future in Angular/Ionic.

Instead of doing this, IMO it would be better to set the default when opening the modal in the "scanQR" function:

componentProps: {
    title || Translate.instant('core.viewer.qrscannertitle'),
},

If someone opens the modal from another code and doesn't supply a title then it won't be displayed, but I guess it shouldn't happen.

src/core/features/viewer/lang.json
```diff
@@ -5,6 +5,10 @@
     "exitreadingmode": "Exit reading mode",
     "increasetextsize": "Increase text size",
     "openreadingmodesettings": "Open reading mode settings",
+    "qrscannertitle": "Scan QR code",
+    "qrscannerdisabletorch": "Disable torch",
+    "qrscannerenabletorch": "Enable torch",
+    "qrscannerswitchcamera": "Switch camera",
```
dpalou: Reading the issue's comments, I see that Júlia suggested using "Turn on torch", "Turn off torch", “Switch to front camera” and “Switch to rear camera”.

src/core/services/qrscan.ts
```diff
@@ -180,6 +183,55 @@ export class CoreQRScanService {
         delete this.qrScanData;
     }
 
+    /**
+     * Check if the QR scanner camera light can be enabled.
+     *
+     * @returns Whether the QR scanner camera light can be enabled.
+     */
+    async canEnableLight(): Promise<boolean> {
```
dpalou, line 228: Since we're checking "this.canScanQR()" in `startScanQR`, maybe we should do the same in all these methods too, for consistency,

dpalou (changes requested):

src/core/services/qrscan.ts
crazyserver: I preffer doing it in the ngInit of the page

src/core/features/viewer/lang.json
crazyserver: Hmm IDK where my changes went... sorry!

src/core/features/viewer/lang.json
```diff
@@ -3,8 +3,16 @@
     "default": "Default",
     "enterreadingmode": "Enter reading mode",
     "exitreadingmode": "Exit reading mode",
+    "gotosettings": "Go to settings",
```
dpalou: We have a "Open settings" string already, can't we use that one instead?

src/core/features/viewer/components/qr-scanner/qr-scanner.ts
```diff
@@ -25,21 +26,41 @@ import { ModalController, Translate } from '@singletons';
...
         CoreSharedModule,
     ],
 })
 export class CoreViewerQRScannerComponent implements OnInit, OnDestroy {
 
-    @Input() title?: string; // Page title.
+    readonly title = input(Translate.instant('core.viewer.qrscannertitle'));
+
+    canEnableTorch = false;
+    canSwitchCamera = false;
```
dpalou: I recommend converting these to signal too, that way in the template we'll only use signals so the component should work fine in a zoneless app.

src/core/services/qrscan.ts
```diff
@@ -139,13 +145,37 @@ export class CoreQRScanService {
...
+                return;
+            }
             // eslint-disable-next-line @typescript-eslint/naming-convention
             error.message = error.message || (error as { _message?: string })._message;
 
             throw error;
         }
     }
 
+    protected async askForPermissionWhenDenied(): Promise<void> {
```
dpalou, line 171: This function lacks JSDoc.

dpalou (changes requested):

dpalou (approved):

PR #4509, opened 2025-07-21
MOBILE-4854 messages: Remove support for messages on LMS 3.5

crazyserver:


src/addons/messages/services/messages.ts
```diff
@@ -1528,26 +1163,29 @@ export class AddonMessagesProvider {
...
         limitFromRead: number = 0,
-        notUsed: boolean = false, // eslint-disable-line @typescript-eslint/no-unused-vars
         siteId?: string,
     ): Promise<AddonMessagesGetMessagesMessage[]> {
         limitFromUnread = limitFromUnread || 0;
         limitFromRead = limitFromRead || 0;
 
         params.read = false;
         params.limitfrom = limitFromUnread;
+        params.limitnum = ADDON_MESSAGES_LIMIT_MESSAGES;
```
dpalou: Instead of overriding the value, maybe it would be better to do:

`params.limitnum = params.limitnum || ADDON_MESSAGES_LIMIT_MESSAGES;`

This way the function allows specifying a limitnum, since it's what the parameter implies. If we don't want to allow defining limitnum then the type of params should specify it to avoid confusion.

src/addons/messages/services/messages.ts
```diff
@@ -1528,26 +1163,29 @@ export class AddonMessagesProvider {
...
-     * @param preSets Set of presets for the WS.
      * @param limitFromUnread Number of read messages already fetched, so fetch will be done from this number.
      * @param limitFromRead Number of unread messages already fetched, so fetch will be done from this number.
-     * @param notUsed // Deprecated 3.9.5
      * @param siteId Site ID. If not defined, use current site.
      * @returns Promise resolved with the data.
      */
     async getRecentMessages(
         params: AddonMessagesGetMessagesWSParams,
-        preSets: CoreSiteWSPreSets,
```
dpalou, line 1540: About removing preSets, IMO it would be nice to convert the `siteId` parameter to an `options` parameter with type `CoreSitesCommonWSOptions`, that way we don't lose the functionality to fetch data from cache if needed (even if we aren't using it right now). And by default it could use CoreSitesReadingStrategy.ONLY_NETWORK to avoid using cache. What do you think?

dpalou (changes requested):

src/addons/messages/services/messages.ts
crazyserver: Thanks! I've also removed the first 2 lines (the defaults were already set) and changed || by ??

src/addons/messages/services/messages.ts
```diff
@@ -1528,28 +1163,28 @@ export class AddonMessagesProvider {
...
-        limitFromRead = limitFromRead || 0;
-
         params.read = false;
         params.limitfrom = limitFromUnread;
+        params.limitnum = params.limitnum ?? ADDON_MESSAGES_LIMIT_MESSAGES;
+
+        const preSets: CoreSiteWSPreSets = {
+            cacheKey: AddonMessages.getCacheKeyForDiscussion(params.useridto),
+            getFromCache: false,
+            emergencyCache: false,
```
dpalou:
```suggestion
            ...CoreSites.getReadingStrategyPreSets(options.readingStrategy ?? CoreSitesReadingStrategy.ONLY_CACHE),
```

dpalou (approved):

dpalou (approved):

PR #4510, opened 2025-07-21
Mobile 4515 module-summary: Check offline use config to enable download

alfonso-salces:


src/core/features/course/components/module-summary/module-summary.html
```diff
@@ -198,7 +198,7 @@ <h2>
             </ion-card>
         }
 
-        @if ((canPrefetch && displayOptions.displayPrefetch) || (size && displayOptions.displaySize)) {
+        @if (canPrefetch && (displayOptions.displayPrefetch || (size && displayOptions.displaySize))) {
```
dpalou: This change is wrong IMO. _canPrefetch_ will be false if offline usage is **enabled** in the site but the activity is marked as NOW_DOWNLOADABLE. But in that case the activity can have downloaded data, and we want to display the size and allow deleting it, that's why the if was implemented this way.

Even if a module is NOT_DOWNLOADABLE (the whole activity cannot be downloaded for some reason, e.g. an activity that has been closed already), the app can still store information if the activity is opened: store WebService requests responses, download embedded files, etc. That's why we want to display the size even if the module is NOT_DOWNLOADABLE.

IMO it's ok to hide the whole block if the NoDelegate_CoreOffline feature is disabled, but we should keep the previous logic otherwise.

src/core/features/course/tests/behat/disabled_features.feature
```diff
@@ -290,3 +290,15 @@ Feature: Test disabled text is shown when opening a disabled activity.
...
+    Given the following "activities" exist:
+      | activity   | course | name            | intro                 | content | showdescription |
+      | page       | C1     | Test page name  | Test page description | CONTENT | 1               |
+    And the following config values are set as admin:
+      | disabledfeatures | NoDelegate_CoreOffline | tool_mobile |
+    Then I entered the course "Course 1" as "student1" in the app
+    And I should find "Test page name" in the app
+    And I press "Test page name" in the app
+    When I press "Information" "ion-button" in the app
+    Then I should not find "Download" in the app
```
dpalou: The purpose of this test is to test the "Information" menu, not to test that the activity has been created by the generator, so IMO you can skip the test to see that the activity is in the course (if it's not, the "press" step will fail anyway). So IMO the test can be simplified like this:

```suggestion
    When I entered the course "Course 1" as "student1" in the app
    And I press "Test page name" in the app
    And I press "Information" "ion-button" in the app
    Then I should not find "Download" in the app
```

dpalou (changes requested):

src/core/features/course/components/module-summary/module-summary.html
```diff
@@ -198,7 +198,7 @@ <h2>
             </ion-card>
         }
 
-        @if ((canPrefetch && displayOptions.displayPrefetch) || (size && displayOptions.displaySize)) {
+        @if (isOfflineUseDisabled && ((canPrefetch && displayOptions.displayPrefetch) || (size && displayOptions.displaySize))) {
```
dpalou: The isOfflineUseDisabled check should be the opposite. You want to hide the information if offline is disabled, but you're only showing it if it's disabled. That's why your behat failed.

PR #4511, opened 2025-07-21
MOBILE-4855 bootstrap: Support Bootstrap 4 and 5 tabs

crazyserver:


src/core/singletons/bootstrap.ts
```diff
@@ -194,6 +196,127 @@ export class CoreBootstrap {
...
+        elements.forEach((element) => {
+            if (!element.hasAttribute('role')) {
+                element.setAttribute('role', 'tablist');
+            }
+            const children = Array.from(element.querySelectorAll<HTMLElement>('.nav-link:not(.dropdown-toggle),\
+                .list-group-item:not(.dropdown-toggle),\
+                [role="tab"]:not(.dropdown-toggle),\
+                [data-bs-toggle="tab"],[data-toggle="tab"],\
+                [data-bs-toggle="pill"],[data-toggle="pill"],\
+                [data-bs-toggle="list"],[data-toggle="list"]'));
```
dpalou: This code doesn't follow our coding guidelines in terms of wrapping. Also, I guess the string will end up with several blank spaces between each option (because of the indentation), which probably don't cause any problem but it can look weird. It might be better to use an array with all the options and then do `.join(,)`.

In this same function you have a couple more cases where you're using a multiline string in querySelector. Those are ok in terms of coding guidelines, so they can stay as they are if they work, but they can also be converted to array for consistency if you prefer. As you wish.

If you prefer to keep this one as a multiline string, then it should be like this:


```suggestion
            const children = Array.from(element.querySelectorAll<HTMLElement>(
                '.nav-link:not(.dropdown-toggle),\
                .list-group-item:not(.dropdown-toggle),\
                [role="tab"]:not(.dropdown-toggle),\
                [data-bs-toggle="tab"],[data-toggle="tab"],\
                [data-bs-toggle="pill"],[data-toggle="pill"],\
                [data-bs-toggle="list"],[data-toggle="list"]',
            ));
```

src/core/singletons/bootstrap.ts
```diff
@@ -194,6 +196,127 @@ export class CoreBootstrap {
...
+            // Initialize the accordion.
+            element.addEventListener('click', async (ev: Event) => {
+                if (element.classList.contains('active')) {
+                    return;
+                }
+
+                ev.preventDefault();
+                ev.stopPropagation();
+
+                // On format text div root moved to core-format-text element.
```
dpalou: This comment is a bit cryptic. I understand what it means because I have a deep knowledge of format-text, but I would make it more friendly.

dpalou (changes requested):

src/core/singletons/bootstrap.ts
crazyserver: I've changed all of them a bit not to repeat them so much.

PR #4512, opened 2025-07-22
MOBILE-4798 edit-entry: Show or hide publish state select

alfonso-salces:


src/addons/blog/pages/edit-entry/edit-entry.ts
```diff
@@ -90,6 +90,8 @@ export default class AddonBlogEditEntryPage implements CanLeave, OnInit, OnDestr
     siteHomeId?: number;
     forceLeave = false;
     isOfflineEntry = false;
+    readonly bloglevel = signal(CoreSiteBlogLevel.BLOG_SITE_LEVEL);
```
dpalou: This variable doesn't follow camel case, it should be "blogLevel".

src/addons/blog/pages/edit-entry/edit-entry.ts
```diff
@@ -132,6 +134,13 @@ export default class AddonBlogEditEntryPage implements CanLeave, OnInit, OnDestr
             return CoreNavigator.back();
         }
 
+        const bloglevel = parseInt(await site.getConfig('bloglevel'));
+        this.bloglevel.set(bloglevel ?? CoreSiteBlogLevel.BLOG_SITE_LEVEL);
```
dpalou: This "??" won't work as expected because of the "parseInt". parseInt will always return a number or NaN, and "??" operator doesn't work with NaN (it will always keep NaN), so this "??" operator won't do anything.

src/addons/blog/tests/behat/edit-entry.feature
```diff
@@ -67,3 +67,23 @@ Feature: Edit blog entries
...
+
+  Scenario: Changing blog visibility to Users can only see their own blog
+    Given the following config values are set as admin:
+      | bloglevel | 1 |
+    Then I entered the app as "testuser"
+    When I press the user menu button in the app
+    And I press "Blog entries" in the app
+    And I press "Add a new entry" in the app
+    And I press "Publish to" in the app
+    And I should not find "Anyone on this site" in the app
```
dpalou:
```suggestion
    And I entered the app as "testuser"
    When I press the user menu button in the app
    And I press "Blog entries" in the app
    And I press "Add a new entry" in the app
    And I press "Publish to" in the app
    Then I should not find "Anyone on this site" in the app
```

src/addons/blog/tests/behat/edit-entry.feature
```diff
@@ -67,3 +67,23 @@ Feature: Edit blog entries
...
+
+  Scenario: Changing blog visibility to All site users can see all blog entries
+    Given the following config values are set as admin:
+      | bloglevel | 4 |
+    Then I entered the app as "testuser"
+    When I press the user menu button in the app
+    And I press "Blog entries" in the app
+    And I press "Add a new entry" in the app
+    And I press "Publish to" in the app
+    And I should find "Anyone on this site" in the app
```
dpalou:
```suggestion
    And I entered the app as "testuser"
    When I press the user menu button in the app
    And I press "Blog entries" in the app
    And I press "Add a new entry" in the app
    And I press "Publish to" in the app
    Then I should find "Anyone on this site" in the app
```

src/addons/blog/pages/edit-entry/edit-entry.html
```diff
@@ -30,9 +30,11 @@ <h1>{{ entry ? entry.subject : 'addon.blog.addnewentry' | translate }}</h1>
...
                         {{ 'addon.blog.publishtonoone' | translate }}
                     </ion-select-option>
-                    <ion-select-option class="core-select-option-title" [value]="publishState.site">
-                        {{ 'addon.blog.publishtosite' | translate }}
-                    </ion-select-option>
+                    @if (!isUserLevel()) {
+                        <ion-select-option class="core-select-option-title" [value]="publishState.site">
+                            {{ 'addon.blog.publishtosite' | translate }}
+                        </ion-select-option>
+                    }
```
dpalou, line 37: This code is correct, but looking at Moodle's code I saw that they also display an option 'publishtoworld' when bloglevel >= BLOG_GLOBAL_LEVEL. We should probably do the same in the app if the WebServices allow it. You can do it in this issue (in a new commit) or in a new issue, as you prefer.

src/addons/blog/pages/edit-entry/edit-entry.ts
```diff
@@ -519,3 +528,14 @@ type AddonBlogEditEntrySaveEntryParams = {
 type AddonBlogEditEntryFormattedOfflinePost = Omit<
     AddonBlogEditEntryPost, | 'attachment' | 'attachmentfiles' | 'rating' | 'format' | 'usermodified' | 'module'
 > & { attachmentfiles?: CoreFileEntry[] };
+
+/**
+ * Restriction level of user blog visualization.
+ */
+const enum CoreSiteBlogLevel {
```
dpalou: This constant should be in a constants file so it can be used in other places.

PR #4518, opened 2025-07-28
MOBILE-4817 forgotten-password: Beautify invalidparameter error

alfonso-salces:


src/core/features/login/pages/forgotten-password/forgotten-password.ts
```diff
@@ -125,7 +126,11 @@ export default class CoreLoginForgottenPasswordPage implements OnInit {
                 await CoreLoginHelper.passwordResetRequested(this.site.getURL());
             }
         } catch (error) {
-            CoreAlerts.showError(error);
+            if (error.debug.code === 'invalidparameter') {
```
crazyserver: debug is not avaiable if debug mode is not enabled, I think you should check error.errorcode instead. Test it to ensure this is correct.

crazyserver (changes requested):

src/core/features/login/pages/forgotten-password/forgotten-password.ts
```diff
@@ -125,7 +126,11 @@ export default class CoreLoginForgottenPasswordPage implements OnInit {
                 await CoreLoginHelper.passwordResetRequested(this.site.getURL());
             }
         } catch (error) {
-            CoreAlerts.showError(error);
+            if (error.errorcode === 'invalidparameter') {
+                CoreAlerts.showError(new CoreError(Translate.instant('core.invalidusername')));
```
crazyserver: We should discuss if this message will be shown on both username and email

alfonso-salces: We finally decided to use `core.login.invalidemail` if the field is an email, otherwise we'll use `core.invalidusername`

PR #4519, opened 2025-07-29
MOBILE-4796 screen: Watch full screen to hide statusbars

crazyserver:


src/core/services/screen.ts
```diff
@@ -128,6 +128,21 @@ export class CoreScreenService {
         });
     }
 
+    /**
+     * Watch fullscreen changes.
+     */
+    async watchFullscreen(): Promise<void> {
```
crazyserver, line 134: I have doubts between placing this function here or in CoreAppProvider

albertgasset (dismissed):

PR #4523, opened 2025-07-29
Mobile 4770: Activities overview

dpalou:


src/core/features/course/pages/overview/overview.ts
```diff
@@ -0,0 +1,180 @@
...
+export default class CoreCourseOverviewPage implements OnInit {
+
+    loaded = signal(false);
+    modTypes = signal<OverviewModType[]>([]);
+
+    protected courseId!: number;
+    protected logView: () => void;
+
+    constructor() {
+        this.logView = CoreTime.once(async () => {
```
crazyserver, line 70: This is just a comment: once function should be placed on CoreUtils IMHO, with debounce....

src/core/features/course/pages/overview/overview.ts
```diff
@@ -0,0 +1,180 @@
...
+    protected async loadModTypes(): Promise<void> {
+        try {
+            const sections = await CoreCourse.getSections(this.courseId, false, true);
+
+            const archetypes: Record<string, number> = {};
+            const modIcons: Record<string, string> = {};
+            let modFullNames: Record<string, string> = {};
+            const brandedIcons: Record<string, boolean|undefined> = {};
+
+            const modules = CoreCourse.getSectionsModules(sections, {
```
crazyserver, line 114: I'd suggest to move all this code to a helper since it's being used by list-mod-types and activity modules block. But they will be deprecated soon so, no problem at all.

src/core/features/course/services/course-helper.ts
```diff
@@ -1115,21 +1115,40 @@ export class CoreCourseHelperProvider {
...
+
+    /**
+     * Given a completion info, load the offline completion any and return the completion with the offline data added.
+     *
+     * @param cmId The module ID.
+     * @param completiondata The completion data.
+     * @param siteId Site ID. If not defined, current site.
+     * @returns Promise resolved when done.
+     */
+    async loadOfflineCompletionData(
```
crazyserver, line 1141: Suggestion: I would name this getOfflineCompletionData since it's returning. Also return JSDoc can be improved

src/core/features/course/pages/overview/overview.ts
```diff
@@ -133,7 +143,7 @@ export default class CoreCourseOverviewPage implements OnInit {
             modFullNames = CoreObject.sortValues(modFullNames);
 
             const modTypes = await Promise.all(Object.keys(modFullNames).map(async (modName): Promise<OverviewModType> => {
-                const iconModName = modName === 'resources' ? 'page' : modName;
+                const iconModName = modName === 'resource' ? 'page' : modName;
```
crazyserver: Should we use ADDON_MOD_RESOURCE_MODNAME ?
I think there are more place where we use 'resource' directy.

src/core/features/course/classes/module-base-handler.ts
```diff
@@ -90,4 +92,49 @@ export class CoreModuleHandlerBase implements Partial<CoreCourseModuleHandler> {
...
+            };
+        }
+
+        if (item.contenttype === 'core_calendar\\output\\humandate' && Number(item.parsedData.timestamp)) {
+            return {
+                content: CoreTime.userDate(Number(item.parsedData.timestamp) * 1000),
+            };
+        }
+
+        if (item.contenttype === 'basic') {
```
crazyserver: I would extract this basic into a constant.

src/core/features/course/pages/overview/overview.html
```diff
@@ -0,0 +1,44 @@
...
+        </ion-title>
+    </ion-toolbar>
+</ion-header>
+<ion-content class="limited-width">
+    <ion-refresher slot="fixed" [disabled]="!loaded()" (ionRefresh)="refreshData($event.target)">
+        <ion-refresher-content pullingText="{{ 'core.pulltorefresh' | translate }}" />
+    </ion-refresher>
+    <core-loading [hideUntil]="loaded()">
+        <ion-item class="ion-text-wrap">
+            <ion-label>
```
crazyserver, line 17: I cannot test the issue right now but I place the comment here not to forget: This should be an item-heading?

src/core/features/user/services/user-preferences.ts
```diff
@@ -0,0 +1,299 @@
...
+    }
+
+    /**
+     * Old cache key for user preferences.
+     *
+     * @param name Preference name.
+     * @returns Cache key.
+     * @deprecated since 5.1. In the future it can be removed because the cache should already use the new cache key.
+     */
+    protected getOldPreferenceCacheKey(name: string): string {
```
crazyserver, line 135: If this is only used to invalidate old cache, would not be better to invalidate on an upgrade method instead of calling the invalidate of old data everytime?

The alternative is to also read from getOldPreferenceCacheKey also... An improvement (future) should be to have some internal cache migration methods.

src/core/features/course/components/overview-item-action-link/overview-item-action-link.scss
```diff
@@ -0,0 +1,12 @@
+:host {
+    ion-button {
+        margin-left: 0;
```
crazyserver: Not to be changed but I prefer always adding the units (px) because if instead of a property we use a var : --margin-left: 0; and this is used to do some calc, it won't work properly.

src/core/features/course/pages/overview/overview.scss
```diff
@@ -1,6 +1,23 @@
...
+    ion-accordion ion-item core-mod-icon {
+        --margin-vertical: 0;
+        --margin-end: 0;
+    }
+
+    --rotate-expandable: rotate(180deg);
+
+    .expandable-status-icon {
+        font-size: var(--text-size);
+        @include margin-horizontal(0, 2px);
```
crazyserver: Units missing

src/core/features/course/pages/overview/overview.scss
```diff
@@ -1,6 +1,23 @@
 @use "theme/globals" as *;
 
-ion-accordion ion-item core-mod-icon {
-    --margin-vertical: 0;
-    --margin-end: 0;
+:host {
+    ion-accordion ion-item core-mod-icon {
+        --margin-vertical: 0;
```
crazyserver: Units missing, it can make other calculations to fail, please add them for safety reasons.

src/core/features/course/classes/module-base-handler.ts
```diff
@@ -140,6 +140,15 @@ export class CoreModuleHandlerBase implements Partial<CoreCourseModuleHandler> {
             };
         }
 
+        if (item.contenttype === 'core_courseformat\\output\\local\\overview\\overviewdialog') {
```
crazyserver: After reviewing all that code, I'm starting to think that it should be better to encapsulate this values somewhere (enum or separate constants). An enum will be incomplete I think. In any case you can add some comment about where to get this value from and why you are using it.

crazyserver (changes requested):

src/core/features/course/pages/overview/overview.ts
dpalou: As we talked, both list-mod-types and the activity modules should be deprecated in 5.1, so I prefer to keep the logic separated because the implementations are a bit different, IMO it's not worth trying to abstract it.

src/core/features/course/services/course-helper.ts
dpalou: IMO getOfflineCompletionData makes you think that it will only return the offline data, and if there isn't offline data it will return undefined or throw an error, but that's not the case. I'm not a huge fan of the "load" name, but IMO it's better than get. I will fix the JSDoc.

src/core/features/course/pages/overview/overview.html
dpalou: IMO it shouldn't, it's just a "description" of the page.

src/core/features/course/pages/overview/overview.ts
dpalou: ADDON_MOD_RESOURCE_MODNAME refers to the activity mod_resource, but in this case we're referring to a collection of all the activities with resource archetype, it's not the same. I changed it to use a local constant.

src/core/features/course/pages/overview/overview.ts
dpalou: We can do it later, in a separate issue.

src/core/features/course/classes/module-base-handler.ts
dpalou: I created an enum.

src/core/features/user/services/user-preferences.ts
dpalou: I wanted to avoid migrating the info because it's more risky, if something goes wrong migrating we could lose offline info. The invalidatePreference method isn't used much so IMO it's ok to keep it like this. What do you think? We can discuss it if you want.

src/core/features/course/classes/module-base-handler.ts
dpalou: I mentioned it another comment, I created an enum, yes. The enum will contain the ones supported by default in the app.

src/core/features/user/services/user-preferences.ts
crazyserver: It's ok! :-)

src/core/features/course/pages/overview/overview.ts
```diff
@@ -0,0 +1,402 @@
...
+                        activity,
+                        item,
+                    },
+                    content: 'content' in content ? content.content : null,
+                    classes: header.classes.concat(content.classes ?? []),
+                };
+            }));
+
+            const nameItemToRender = itemsToRender.find(item => item.key === 'name') ??
+                itemsToRender.find(item => item.contenttype === 'core_courseformat\\output\\local\\overview\\activityname');
```
crazyserver: You should use the enum here as well

src/core/features/course/tests/behat/overview.feature
```diff
@@ -0,0 +1,287 @@
...
+    And I should find "To do" within "Terra baixa" "tr" in the app
+
+    When I press "Mark as done" within "Mar i Cel" "tr" in the app
+    Then I should find "Done" within "Mar i Cel" "tr" in the app
+
+    When I press "To do" within "Terra baixa" "tr" in the app
+    Then I should find "To do:" in the app
+    And I should find "View" in the app
+
+  # Scenario: Activities overview provide grade information to the student
```
crazyserver: Why is this scenario disabled?

src/core/features/course/pages/overview/overview.html
```diff
@@ -114,9 +114,9 @@ <h1>{{ 'core.activities' | translate }}</h1>
 
         @for (activity of overview.activities; track activity.cmid) {
             <ion-card>
-                @if (activity.haserror) {
+                @if (activity.haserror || !activity.hasItemsBesidesName) {
                     <!-- If activity has an error, display just the header without being expandable. -->
-                    <ion-item class="ion-text-wrap core-course-overview-activity-has-error">
+                    <ion-item class="ion-text-wrap" [ngClass]="{'core-course-overview-activity-has-error': activity.haserror}">
```
crazyserver: Totally optional but I preffer [class.className] syntax, I think is crearer.

src/core/features/course/pages/overview/overview.ts
dpalou: You're right, I changed it.

src/core/features/course/tests/behat/overview.feature
dpalou: The scenario is in LMS tests, but I think this should be tested in each activity so I commented just to be sure that, in case I forgot to remove it, we noticed it like you did xD I just removed it.

src/core/features/course/pages/overview/overview.html
dpalou: Changed.

crazyserver (approved):

crazyserver (approved):

PR #4531, opened 2025-08-01
Mobile 4860

crazyserver:


src/core/features/user/services/handlers/participants-link.ts
```diff
@@ -0,0 +1,58 @@
...
+        siteIds: string[],
+        url: string,
+        params: Record<string, string>,
+        courseId?: number,
+    ): CoreContentLinksAction[] | Promise<CoreContentLinksAction[]> {
+        const courseIdentifier = courseId || Number(params.id);
+
+        return [{
+            action: async (siteId): Promise<void> => {
+                await CoreCourseHelper.getAndOpenCourse(courseIdentifier, { selectedTab: 'CoreUserParticipants' }, siteId);
```
dpalou: The course to open might already be the current page, in that case there's no need to call the WebServices to get the course info. What we usually do is:

```
if (
    (!siteId || siteId === CoreSites.getCurrentSiteId()) &&
    CoreCourse.currentViewIsCourse(courseId)
) {
    CoreCourse.selectCourseTab('', pageParams);
} else {
    await CoreCourseHelper.getAndOpenCourse(courseId, pageParams, siteId);
}
```

We should probably abstract this logic somewhere, or make getAndOpenCourse handle this logic first.

src/core/components/course-image/course-image.ts
```diff
@@ -65,15 +60,33 @@ export class CoreCourseImageComponent {
     protected async setCourseColor(): Promise<void> {
         const course = this.course();
 
-        await CoreCoursesHelper.loadCourseColorAndImage(course);
+        // Moodle 4.1 downwards geopatterns are embedded in b64 in only some WS, remote them to keep it coherent.
```
dpalou: Typo: I guess you mean "remove", not "remote".

src/core/features/course/components/course-index/course-index.ts
```diff
@@ -63,7 +63,7 @@ export class CoreCourseCourseIndexComponent implements OnInit {
             return;
         }
 
-        let completionEnabled = CoreCoursesHelper.isCompletionEnabledInCourse(this.course);
+        let completionEnabled = AddonCourseCompletion.isCompletionEnabledInCourse(this.course);
```
dpalou: Code from core shouldn't use code from an addon. I know we're already doing it in course helper, but with this change we're doing in more places. IMO we should either:

- Move this logic to core (either couse helper or another service), and make the completion addon use the core logic to check if it's enabled.
- Move the completion addon (or part of it) to core if we think it's important enough to be in core.

src/core/features/courses/pages/dashboard/dashboard.ts
```diff
@@ -135,16 +137,17 @@ export default class CoreCoursesDashboardPage implements OnInit, OnDestroy {
     protected loadFallbackBlocks(): void {
         this.blocks = [
             {
-                name: 'myoverview',
+                name: ADDON_BLOCK_MYOVERVIEW_BLOCK_NAME,
```
dpalou: Here we're using "code" from an addon in core, which we shouldn't do, even if it's just a constant. Core shouldn't be aware of which addons are in the app.

The "cleaner" solution would probably be to modify the delegate to allow blocks specify if they're fallback blocks or similar, but that means any block could be added as fallback and I'm not sure we want that.

An easier solution is to list all the "core" blocks in core/features/block constants, with a constant for each block included in LMS. That way we can use the constants in core, and each addon can import its constant from core.

PR #4534, opened 2025-08-04
MOBILE-4821 rte: New resizing algorithm 

albertgasset:


src/core/directives/collapsible-header.ts
```diff
@@ -514,6 +528,9 @@ export class CoreCollapsibleHeaderDirective implements OnInit, OnChanges, OnDest
         }
 
         this.content = content;
+        if (content.nextElementSibling?.tagName === 'ION-ROUTER-OUTLET') {
+            this.hasSplitView = content.nextElementSibling.querySelector('ion-content:not(.disable-scroll-y)') !== null;
+        }
```
dpalou: I don't like this way to check if it has split view, it can easily be broken by mistake. The main problem is that collapsible-header is not a parent or a child of split view, so we cannot use inject or ViewChild.

One possible alternative is to first get the split-view element (if any):

`const splitViewEl = this.content.closest('core-split-view');`

We could use the "outlet-activated" class to check if it's active or not, but that can easily be broken too by mistake, it's better to directly call the `outletActivated` getter of the component instance.

This means you will need to obtain the component instance from the DOM element, and you can use CoreDirectivesRegistry to achieve that. In the split view constructor you should register the instance:

`CoreDirectivesRegistry.register(this.element, this);`

And then in collapsible-header you can obtain the instance:
```

const splitView = splitViewEl ? CoreDirectivesRegistry.resolve(splitViewEl, CoreSplitViewComponent) : null;
this.hasSplitView = !!splitView?.outletActivated;
```

What do you think about this approach?

src/core/features/editor/components/rich-text-editor/rich-text-editor.ts
```diff
@@ -382,50 +375,39 @@ export class CoreEditorRichTextEditorComponent implements AfterViewInit, OnDestr
...
-            while (scrollContentHeight === 0 && content?.children) {
-                const children = Array.from(content.children)
-                    .filter((element) => element.slot !== 'fixed' && !element.classList.contains('core-loading-container'));
+        // Make sure there is enough space to render the editor.
+        this.height = Math.max(this.height, CoreEditorRichTextEditorComponent.MIN_HEIGHT);
 
-                scrollContentHeight = children
-                    .map((element) => element.getBoundingClientRect().height)
-                    .reduce((a, b) => a + b, 0);
+        this.element.style.setProperty('height', `${this.height}px`);
```
dpalou: Instead of setting the style "manually", we can use the Component decorator's host property for that, that way every time the height changes the style will automatically change too. I recommend using a signal to improve Angular's change detection. You need to add this in the Component decorator of the RTE component:

```
    host: {
        '[style.height]': 'height() + "px"',
    },
```

dpalou (changes requested):

PR #4535, opened 2025-08-04
MOBILE-4820 image-edit: Use base64 format to aoiv incompatibilities

crazyserver:


src/core/services/file.ts
```diff
@@ -443,6 +443,29 @@ export class CoreFileProvider {
...
+                    reject('Error reading blob as base64.');
+
+                    return;
+                }
+                resolve(reader.result.toString());
+            };
+
+            reader.readAsDataURL(blob);
+    });
+    }
```
dpalou: I don't like this solution, working with base64 is not as efficient because it requires extra encode/decode and the image size can be up to 33% bigger when using base64.

The reason your code wasn't working is that fileObject has type "IFile", not "Blob". They share a lot of properties, but they're not exactly the same. Blob has an _arrayBuffer_ function, but IFile doesn't, that's why the app is throwing an error in the console with your code (it tries to use the arrayBuffer function but it doesn't exist).

Please notice in that webapp it worked because OUR implementation of the file API returns a Blob, but the Cordova implementation doesn't.

To make it work you need to convert the fileObject to Blob, and the implementation is quite similar to the blobToBase64 function you created, something like:

```
const reader = new FileReader();
reader.onload = (event) => {
    if (!reader.result) {
        reject(new Error('FileReader returned no result.'));

        return;
    }
    resolve(new Blob([new Uint8Array(reader.result as ArrayBuffer)], { type: 'image/png' }));
};
reader.onerror = reject;
reader.readAsArrayBuffer(fileEntry);
```

Please notice I hardcoded 'image/png', you should probably get the type from the fileEntry or fileObject.

dpalou: Also, this function sounds like a "utils" function, I can't think of any case where we might want to override this. So it might be better to create a File singleton (not a service) and put the function there.

dpalou (changes requested):

PR #4538, opened 2025-08-06
Mobile 4842

crazyserver:


src/core/features/block/constants.ts
```diff
@@ -0,0 +1,27 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+export enum CoreBlocksRegion {
+    CONTENT = 'content',
+    MAIN = 'main',
+    SIDE = 'side', // Used as a prefix for side blocks. (site-pre, side-post).
```
dpalou: Typo: site-pre shoud be side-pre I guess.

src/core/features/courses/pages/dashboard/dashboard.ts
```diff
@@ -133,18 +134,13 @@ export default class CoreCoursesDashboardPage implements OnInit, OnDestroy {
...
-            },
-        ];
-
-        this.hasMainBlocks = CoreBlockDelegate.isBlockSupported('myoverview') || CoreBlockDelegate.isBlockSupported('timeline');
+        this.blocks = CORE_BLOCKS_DASHBOARD_FALLBACK_BLOCKS.map((blockName) => ({
+            name: blockName,
+            visible: true,
+        }));
+
+        this.hasSideBlocks = CORE_BLOCKS_DASHBOARD_FALLBACK_BLOCKS.some((blockName) =>
```
dpalou: Why did you change `hasMainBlocks` with `hasSideBlocks`? Looking at the code, the fallback blocks are added as main blocks, not side blocks.

src/core/features/courses/services/courses-helper.ts
```diff
@@ -436,14 +437,15 @@ export class CoreCoursesHelperProvider {
 
     /**
      * Check whether completion is available in a certain course.
-     * This is a temporary function to be used until we move AddonCourseCompletion to core folder (MOBILE-4537).
+     * This is a temporary function to be used until we move CoreCourseCompletionHelper to core folder (MOBILE-4537).
```
dpalou: This comment is a bit confusing, since CoreCourseCompletionHelper is now in core folder. Is MOBILE-4537 still needed after these changes?

src/core/features/course/services/course-completion-helper.ts
```diff
@@ -0,0 +1,66 @@
...
+import { CoreSite } from '@classes/sites/site';
+import { CoreCourseAnyCourseData } from '@features/courses/services/courses';
+import { CoreSites } from '@services/sites';
+import { makeSingleton } from '@singletons';
+
+/**
+ * Service that provides some features regarding a course completion.
+ */
+@Injectable({ providedIn: 'root' })
+export class CoreCourseCompletionHelperService {
```
dpalou: In this case IMO we can remove the "Helper" word and just name it `CoreCourseCompletionService`. AFAIK the "Helper" services were created mainly for 2 reasons:

1. To avoid having too much code in a single service.
2. To put some "advanced" logic that could look weird in the base service. E.g. prefetching courses.

In this case we don't have any advanced logic and this service is the only completion service, so I wouldn't use the Helper word.

Lately I have some mixed feelings about our helpers, because the "Helper" word doesn't provide much info on what it does (it's not easy to understand the difference between CourseHelperService and CourseService without looking at their implementations). If we continue the trend of splitting code into different services maybe in the future we can get rid of our Helper services and have better naming.

src/core/features/course/services/course-completion-helper.ts
```diff
@@ -0,0 +1,66 @@
...
+// limitations under the License.
+
+import { Injectable } from '@angular/core';
+import { CoreSite } from '@classes/sites/site';
+import { CoreCourseAnyCourseData } from '@features/courses/services/courses';
+import { CoreSites } from '@services/sites';
+import { makeSingleton } from '@singletons';
+
+/**
+ * Service that provides some features regarding a course completion.
```
dpalou: Maybe we can add a comment indicating that this service is only for course completion, not activity completion in a course. Since both things can be confused.

E.g. in our Course service we have the type CoreCourseCompletionActivityStatus, but that type doesn't belong here because it's for activity completion, not course completion.

src/addons/block/timeline/classes/section.ts
```diff
@@ -14,76 +14,55 @@
...
      */
     async loadMore(): Promise<void> {
-        this.dataSubject$.next({
-            ...this.dataSubject$.value,
-            loadingMore: true,
-        });
+        this.loadingMore.set(true);
 
-        const lastEventId = this.dataSubject$.value.lastEventId;
         const { events, canLoadMore } = this.course
```
dpalou: This isn't new, but the fact that "canLoadMore" is an event ID is really confusing, I expect a variable named "canLoadMore" to be a boolean. The timeline service is not exported to services, so this can be a good opportunity to rename that variable to something that makes more sense.

src/addons/block/timeline/components/events/addon-block-timeline-events.html
```diff
@@ -1,89 +1,93 @@
...
-                    </div>
-                }
-            </ion-label>
-        </ion-item>
-    </ng-container>
-</ion-item-group>
+                            @if (event.activitystr) {
+                                <p>
+                                    <span>
+                                        @if (event.activitystr) {
```
dpalou: This if is duplicated, it's already a couple of lines above.

src/addons/block/timeline/components/events/events.ts
```diff
@@ -32,22 +32,21 @@ import { CoreContentLinksHelper } from '@features/contentlinks/services/contentl
...
 export class AddonBlockTimelineEventsComponent implements OnInit {
 
-    @Input() events: AddonBlockTimelineDayEvents[] = []; // The events to render.
-    @Input() course?: CoreEnrolledCourseDataWithOptions; // Whether to show the course name.
-    @Input({ transform: toBoolean }) showInlineCourse = true; // Whether to show the course name within event items.
-    @Input({ transform: toBoolean }) canLoadMore = false; // Whether more events can be loaded.
-    @Input({ transform: toBoolean }) loadingMore = false; // Whether loading is ongoing.
-    @Output() loadMore = new EventEmitter(); // Notify that more events should be loaded.
+    readonly events = input.required<AddonBlockTimelineDayEvents[]>(); // The events to render.
+    readonly course = input.required<CoreEnrolledCourseDataWithOptions>(); // Whether to show the course name.
```
dpalou: This was already wrong, but the comment is wrong, it says "Whether to show the course name."

src/addons/block/timeline/components/events/events.ts
```diff
@@ -32,22 +32,21 @@ import { CoreContentLinksHelper } from '@features/contentlinks/services/contentl
...
+    readonly showInlineCourse = input(true, { transform: toBoolean }); // Whether to show the course name within event items.
+    readonly canLoadMore = input(false, { transform: toBoolean }); // Whether more events can be loaded.
+    readonly loadingMore = input(false, { transform: toBoolean }); // Whether loading is ongoing.
+    readonly loadMore = output(); // Notify that more events should be loaded.
 
     colorizeIcons = false;
```
dpalou: colorizeIcons is used in the template, so IMO it would be better if it was a signal. I guess it would work as it is because it's only set in OnInit, but that way it's harder to introduce bugs later if at some point we need to recalculate the value.

src/addons/block/timeline/components/timeline/timeline.ts
```diff
@@ -47,29 +46,70 @@ import { Translate } from '@singletons';
...
+export class AddonBlockTimelineComponent extends CoreBlockBaseComponent implements OnInit {
+
+    sortFormControl = new FormControl(AddonBlockTimelineSort.ByDates, { nonNullable: true });
+    readonly sort = signal<AddonBlockTimelineSort>(this.sortFormControl.value);
+    sortOptions: AddonBlockTimelineOption<AddonBlockTimelineSort>[];
+
+    filterFormControl = new FormControl(AddonBlockTimelineFilter.Next30Days, { nonNullable: true });
+    readonly filter = signal<AddonBlockTimelineFilter>(this.filterFormControl.value);
+    dateFilterOptions: AddonBlockTimelineOption<AddonBlockTimelineFilter>[];
+    statusFilterOptions: AddonBlockTimelineOption<AddonBlockTimelineFilter>[] = [
```
dpalou: As mentioned in another comment, it's probably safer that all variables used in the template are signals, unless they are constants. E.g. sortFormControl, filterFormControl, dateFilterOptions, etc.

I guess another valid option would be to convert them to readonly, that way we are sure they cannot be changed later.

src/addons/block/timeline/components/timeline/timeline.ts
```diff
@@ -47,29 +46,70 @@ import { Translate } from '@singletons';
...
+        all: { from: -14 },
+        overdue: { from: -14, to: 1 },
+        next7days: { from: 0, to: 7 },
+        next30days: { from: 0, to: 30 },
+        next3months: { from: 0, to: 90 },
+        next6months: { from: 0, to: 180 },
+    };
+
+    readonly init = signal(false);
     loaded = false;
```
dpalou: "loaded" value changes during the execution of the component and it is used in the template, so in this case it should be a signal.

src/addons/block/timeline/components/timeline/timeline.ts
```diff
@@ -116,117 +161,70 @@ export class AddonBlockTimelineComponent implements OnInit, ICoreBlockComponent
...
-                value,
-                name: `addon.block_timeline.${value}`,
-            }));
-    }
+    async loadSections(
+        filter: AddonBlockTimelineFilter,
+        sort: AddonBlockTimelineSort,
+        search: string,
+    ): Promise<void> {
+        const firstLoad = !this.loaded;
```
dpalou: This will work now, but since the "loaded" variable can change it's a bit fragile. If at some point this component sets `loaded` to false **before** calling this function, then it will say "firstLoad=true" but that wouldn't be true, it wouldn't be the first load of the component.

If we really want to detect that it's the first load, IMO it would be better to have a separate property for that.

src/addons/block/timeline/components/timeline/timeline.ts
```diff
@@ -346,3 +346,6 @@ export interface AddonBlockTimelineOption<Value> {
     value: Value;
     name: string;
 }
+
+const SORT_CONFIG_KEY = 'AddonBlockTimelineSort';
+const FILTER_CONFIG_KEY = 'AddonBlockTimelineFilter';
```
dpalou: These constants are only used in the component, so according to our coding guidelines they should be "protected static readonly" in the class.

src/addons/messages/pages/search/search.ts
```diff
@@ -135,6 +135,12 @@ export default class AddonMessagesSearchPage implements OnDestroy {
      * @returns Resolved when done.
      */
     async search(query: string, loadMore?: 'contacts' | 'noncontacts' | 'messages', infiniteComplete?: () => void): Promise<void> {
+        if (!query || query.trim() === '') {
```
dpalou: In some places you check these 2 conditions, but in others like the forum search page you're only checking the second one. Both codes work, but it would be nice to be consistent.

If `search` should always be a string, checking `query.trim() === ''` is already checking `!query` so I guess having only the second condition would be enough.

However, please check my comments in the search component first because they can affect this change here.

src/core/features/search/components/search-box/search-box.ts
```diff
@@ -42,44 +54,65 @@ import { CoreSharedModule } from '@/core/shared.module';
...
+    readonly searchLabel = input(Translate.instant('core.search')); // Label to be used on action button.
+    readonly placeholder = input(Translate.instant('core.search')); // Placeholder text for search text input.
+    readonly autocorrect = input<'on'|'off'>('on'); // Enables/disable Autocorrection on search text input.
+    readonly spellcheck = input(true, { transform: toBoolean }); // Enables/disable Spellchecker on search text input.
+    readonly autoFocus = input(false, { transform: toBoolean }); // Enables/disable Autofocus when entering view.
+    readonly lengthCheck = input(3); // Check value length before submit. If 0, any string will be submitted.
+    readonly showClear = input(true, { transform: toBoolean }); // Show/hide clear button.
+    readonly disabled = input(false, { transform: toBoolean }); // Disables the input text.
+    /**
+     * @deprecated since 5.1. Use (searchText) instead.
```
dpalou: In this comment I think it should be [searchText] instead of (searchText), right?

Another thing to take into account is that [searchText] isn't equivalent to the previous input. Before, if you changed `initialSearch` after the component was initialized then nothing happened. But with [searchText] if you change the value later then it will affect the component. I'm not saying this is wrong, just that it's something that should probably be documented here.

src/core/features/search/components/search-box/search-box.ts
```diff
@@ -42,44 +54,65 @@ import { CoreSharedModule } from '@/core/shared.module';
...
+     */
+    readonly initialSearch = input(''); // Initial search text.
 
+    // eslint-disable-next-line @typescript-eslint/no-deprecated
+    readonly searchText = model(this.initialSearch());
```
dpalou, line 75: Having a `searchText` and a `textInput` might be confusing for people using this component. If I see a component named search-box with a `searchText` and a `searchTextChanged`, I would think that the `searchTextChanged` is triggered every time the user enters a character. But that's not true in this case because you only want to trigger the searchTextChanged when the user clicks submit or clear.

IMO we should make this clear in the documentation, to avoid confusion. Another option would be to remove `textInput` and keep the `onSubmit` and `onClear`, that way `searchTextChanged` will be triggered every time the user types something, and we should still use onSubmit and onClear. But I guess there'll be no use case where we want to automatically detect every letter typed.

src/core/features/block/components/block/block.ts
```diff
@@ -113,12 +113,18 @@ export class CoreBlockComponent implements OnChanges, OnDestroy {
...
 
     /**
-     * Invalidate some data.
-     *
-     * @returns Promise resolved when done.
+     * Invalidate block data.
      */
     async invalidate(): Promise<void> {
         await this.dynamicComponent()?.callComponentMethod('invalidateContent');
+        this.loaded = false;
```
dpalou: IMO It's weird that an invalidate function sets `loaded` to false, specially because you can't be sure that the `loadContent` or `reloadContent` functions will be called after that. E.g. you could have a page using blocks that only invalidates them so they're updated the next time you open the page, but it doesn't reload the data immediately. In that case you would end up with a spinner that would never disappear.

Putting it at the start of `reload` isn't enough? The spinner will appear later in that case because it will wait until the page's fetchContent is finished, but since we have the PTR spinner at the top maybe that's ok.

src/core/components/loading/loading.ts
```diff
@@ -61,54 +57,58 @@ import { CoreTimesPipe } from '@pipes/times';
...
+            if (CorePlatform.isIOS()) {
+                this.mutationObserver.observe(this.element, { childList: true });
+            }
+        } else {
+            this.lastScrollPosition = this.getScrollPosition();
+            this.mutationObserver.disconnect();
+        }
+
+        return loaded;
+    });
```
dpalou: core-loading is one of the components that I migrated to signals in my branch, but I had to remove from my branch because it caused errors with unit tests. The problem was with getScrollElement, which I see you changed so maybe that's what I was missing in my code.

One of the things I did in my code was remove this "loaded" variable, because using "hideUntil" should be enough. And the login about readyPromise, lastScroll, etc. I put it in an effect. You can see it in this commit:

https://github.com/moodlehq/moodleapp/commit/36c6f391359dbfb98bb7c3c7326271fafd0a37d9

src/core/features/reminders/components/date/date.ts
```diff
@@ -33,14 +33,14 @@ import { CoreRemindersSetButtonComponent } from '../set-button/set-button';
...
+    readonly instanceId = input<number>();
+    readonly type = input<string>();
+    readonly label = input('');
+    readonly time = input(0);
+    readonly relativeTo = input(0);
+    readonly title = input('');
+    readonly url = input('');
 
     showReminderButton = false;
     timebefore?: number; // Undefined means no reminder has been set.
```
dpalou: I mentioned it in another component too, I think it's safer if all variables used in the template are signals, unless they're constants. This also incudes readableTime, which isn't displayed here.

`readableTime` and `showReminderButton` can be computed signals and this way they will be recalculated if the inputs change, so we're adding reactivity and making the component better.

timebefore has the problem that it requires an async call, so in order to recalculate it when inputs change we would need to use `resource`, which is still in beta. But you can use an `effect` in the meantime.

src/core/features/reminders/components/set-button/set-button.ts
```diff
@@ -33,26 +33,24 @@ import { CoreSharedModule } from '@/core/shared.module';
...
+    readonly label = input('');
+    readonly initialTimebefore = input<number>();
+    readonly time = input(-1);
+    readonly title = input('');
+    readonly url = input('');
 
-    labelClean = '';
+    readonly labelClean = computed(() => this.label().replace(':', ''));
     timebefore?: number;
     reminderMessage?: string;
```
dpalou: timebefore and reminderMessage should be signals. In this case they must be signals because they change dynamically in the setTimebefore function, which is called when saving a reminder.

src/core/features/reminders/components/set-reminder-custom/set-reminder-custom.ts
```diff
@@ -28,8 +28,8 @@ import { PopoverController } from '@singletons';
 })
 export class CoreRemindersSetReminderCustomComponent {
 
-    @Input() customValue = 10;
-    @Input() customUnits = CoreRemindersUnits.MINUTE;
+    readonly customValue = input(10);
+    readonly customUnits = input(CoreRemindersUnits.MINUTE);
 
     customUnitsOptions = [
```
dpalou: This `customUnitOptions` seems to be a constant, so according to our coding style it should be a `protected static readonly`, but AFAIK we cannot use static properties from a template. One possibility is to declare it just readonly, similar to what we do with the signals, to make it clear it's a constant (even though in this case the array can be mutated even if it's readonly). 

If we decide to do this, we should modify our coding styles to allow this case. We might want to discuss this with the team.

src/core/features/reminders/components/set-reminder-menu/set-reminder-menu.ts
```diff
@@ -36,9 +36,9 @@ import { CoreRemindersUnits, REMINDERS_DISABLED } from '@features/reminders/cons
...
 
-    @Input() initialValue?: number;
-    @Input() eventTime?: number;
-    @Input() noReminderLabel = '';
+    readonly initialValue = input<number>();
+    readonly eventTime = input<number>();
+    readonly noReminderLabel = input('');
 
     currentValue = '0m';
     customLabel = '';
```
dpalou: This component also has properties that should either be signals (currentValue, customLabel) or constants (presetOptions).

src/addons/calendar/components/calendar/calendar.ts
```diff
@@ -67,35 +62,34 @@ import { CoreUserPreferences } from '@features/user/services/user-preferences';
...
     ],
+    host: {
+        '[attr.hidden]': 'hidden() ? "hidden" : null',
+    },
 })
-export class AddonCalendarCalendarComponent implements OnInit, DoCheck, OnDestroy {
+export class AddonCalendarCalendarComponent implements OnInit, OnDestroy {
 
-    @ViewChild(CoreSwipeSlidesComponent) swipeSlidesComponent?: CoreSwipeSlidesComponent<PreloadedMonth>;
+    readonly swipeSlidesComponent = viewChild(CoreSwipeSlidesComponent);
```
dpalou: With this change we've lost some type checking. With your code I can call 

swipeSlidesComponent.slideToItem({ foo: 'bar' });

And it will accept it because the Item inside the slidesComponent is unknown. Before, the slidesComponent knew the type of the item so this code above would complain. So IMO you should add the PreloadedMonth type.

src/addons/calendar/components/calendar/calendar.ts
```diff
@@ -67,35 +62,34 @@ import { CoreUserPreferences } from '@features/user/services/user-preferences';
...
-    @Input() initialMonth?: number; // Initial month to load.
-    @Input() filter?: AddonCalendarFilter; // Filter to apply.
-    @Input({ transform: toBoolean }) hidden = false; // Whether the component is hidden.
-    @Input({ transform: toBoolean }) canNavigate = true; // Whether to include arrows to change the month
-    @Input({ transform: toBoolean }) displayNavButtons = true; // Whether to display nav buttons created by this component.
-    @Output() onEventClicked = new EventEmitter<number>();
-    @Output() onDayClicked = new EventEmitter<{day: number; month: number; year: number}>();
+    readonly initialYear = input<number>(); // Initial year to load.
+    readonly initialMonth = input<number>(); // Initial month to load.
+    readonly filter = input.required<AddonCalendarFilter>(); // Filter to apply.
```
dpalou: Why is `filter` required now?

src/addons/calendar/components/calendar/calendar.ts
```diff
@@ -167,33 +171,6 @@ export class AddonCalendarCalendarComponent implements OnInit, DoCheck, OnDestro
         this.fetchData();
     }
 
-    /**
-     * @inheritdoc
-     */
-    ngDoCheck(): void {
```
dpalou, line 173: Removing the `ngDoCheck` is a breaking change. I'm not saying we shouldn't do it, on the opposite, we should probably try to remove them all, but we need to be careful because we can introduce bugs in our code and in plugins (if the component can be used by plugins, but in this case I think it's not).

For example, before your change if the code outside the component did:

`filter.courseId = 1;`

This component would detect the change and react to it. With your code, that will not happen anymore, the only way to detect changes is to create a new object, like it happens with signals. 

In this specific case, [the page is re-assigning the filter variable](https://github.com/moodlehq/moodleapp/blob/main/src/addons/calendar/pages/index/index.ts#L162), which looks good. However, this `filterData` received is the property `this.filter` from `AddonCalendarFilterComponent`, which is an input received by the modal and the value passed to it is [this.filter from the page](https://github.com/moodlehq/moodleapp/blob/main/src/addons/calendar/pages/index/index.ts#L369). So unless Angular/Ionic create a copy for inputs, there's a possibility that the object you're assigning to the filter property is the same that was already there, so the calendar component will not detect the change in that case.

We should make sure the object is different, e.g. by doing in the page:

`this.filter = { ...filterData };`

Also, I think it's important to document this change somewhere, like in the upgrade notes. Even if it doesn't affect plugins, IMO it's an important change that can cause confusion so it's better to document it.

src/addons/calendar/pages/day/day.ts
```diff
@@ -74,7 +74,7 @@ import { CoreUserPreferences } from '@features/user/services/user-preferences';
 })
 export default class AddonCalendarDayPage implements OnInit, OnDestroy {
 
-    @ViewChild(CoreSwipeSlidesComponent) swipeSlidesComponent?: CoreSwipeSlidesComponent<PreloadedDay>;
+    readonly swipeSlidesComponent = viewChild(CoreSwipeSlidesComponent);
```
dpalou: We lost type checking here, please add the PreloadedDay type.

src/addons/messages/pages/group-conversations/group-conversations.ts
```diff
@@ -72,9 +72,9 @@ const enum AddonMessagesGroupConversationOptionNames {
 })
 export default class AddonMessagesGroupConversationsPage implements OnInit, OnDestroy {
 
-    @ViewChild(CoreSplitViewComponent) splitView!: CoreSplitViewComponent;
-    @ViewChild(IonContent) content?: IonContent;
-    @ViewChild('accordionGroup', { static: true }) accordionGroup!: IonAccordionGroup;
+    readonly splitView = viewChild.required(CoreSplitViewComponent);
+    readonly content = viewChild(IonContent);
+    readonly accordionGroup = viewChild.required<IonAccordionGroup>('accordionGroup');
```
dpalou, line 77: This viewChild used "static: true", I don't know why. Did you double check that it's working as expected now with this change?

src/addons/messages/pages/group-conversations/group-conversations.ts
```diff
@@ -812,12 +814,13 @@ export default class AddonMessagesGroupConversationsPage implements OnInit, OnDe
     protected async expandOption(option: AddonMessagesGroupConversationOption, getCounts = false): Promise<void> {
         // Collapse all and expand the right one.
         option.loading = true;
-        this.accordionGroup.value = option.optionName;
+        const accordionGroup = this.accordionGroup();
+        accordionGroup.value = option.optionName;
```
dpalou: Do we really need to store `accordionGroup` in a variable in this case? Can't you just do this?

`this.accordionGroup().value = option.optionName;`

src/addons/mod/assign/components/feedback-plugin/feedback-plugin.ts
```diff
@@ -40,7 +40,7 @@ import { CoreSharedModule } from '@/core/shared.module';
 })
 export class AddonModAssignFeedbackPluginComponent implements OnInit {
 
-    @ViewChild(CoreDynamicComponent) dynamicComponent!: CoreDynamicComponent<IAddonModAssignFeedbackPluginComponent>;
+    readonly dynamicComponent = viewChild.required(CoreDynamicComponent);
```
dpalou: We lost some type checking here too. E.g. with your code using:

this.dynamicComponent().callComponentMethod('whatever');

Will work, but it shouldn't because 'whatever' doesn't exist in the feedback plugin component. Please add the IAddonModAssignFeedbackPluginComponent type.

src/addons/mod/assign/components/submission-plugin/submission-plugin.ts
```diff
@@ -40,7 +40,7 @@ import { CoreSharedModule } from '@/core/shared.module';
 })
 export class AddonModAssignSubmissionPluginComponent implements OnChanges {
 
-    @ViewChild(CoreDynamicComponent) dynamicComponent!: CoreDynamicComponent<AddonModAssignSubmissionPluginBaseComponent>;
+    readonly dynamicComponent = viewChild.required(CoreDynamicComponent);
```
dpalou: Again, we lost some type checking here too.

src/addons/mod/assign/pages/submission-review/submission-review.ts
```diff
@@ -174,7 +174,8 @@ export default class AddonModAssignSubmissionReviewPage implements OnInit, OnDes
         try {
             await Promise.all(promises);
         } finally {
-            this.submissionComponent && this.submissionComponent.invalidateAndRefresh(true);
+            const submissionComponent = this.submissionComponent();
+            submissionComponent && submissionComponent.invalidateAndRefresh(true);
```
dpalou: Here you can just do this:

`this.submissionComponent()?.invalidateAndRefresh(true);`

src/addons/mod/book/pages/contents/contents.ts
```diff
@@ -55,7 +55,7 @@ import { CoreSharedModule } from '@/core/shared.module';
 })
 export default class AddonModBookContentsPage implements OnInit, OnDestroy {
 
-    @ViewChild(CoreSwipeSlidesComponent) swipeSlidesComponent?: CoreSwipeSlidesComponent;
+    readonly swipeSlidesComponent = viewChild(CoreSwipeSlidesComponent);
```
dpalou: Here we didn't have type checking before, but IMO it's important to have it. In this case the type I think it's LoadedChapter.

src/addons/mod/scorm/pages/online-player/online-player.ts
```diff
@@ -64,26 +63,17 @@ export default class AddonModScormOnlinePlayerPage implements OnInit, OnDestroy
...
-                }
-            });
+        effect(() => {
+            const isOnline = CoreNetwork.isOnline();
+            if (!isOnline) {
+                // User lost connection while playing an online package. Show an error.
+                CoreAlerts.showError(new CoreError(Translate.instant('core.course.changesofflinemaybelost'), {
+                    title: Translate.instant('core.youreoffline'),
+                }));
+            }
```
dpalou: The functionality has changed with this new code. If a user opens the page in offline, before your change it would NOT display the warning because `wasOnline` was false. In that case the app already displays a network error while loading the data.

With your new code, in this case the app will display both the network error message and this message. And I think it can be confusing, IMO we should only display this warning if the user was online and then he lost the connection.

src/app/app.component.ts
```diff
@@ -39,7 +39,7 @@ register();
 })
 export class AppComponent implements OnInit, AfterViewInit {
 
-    @ViewChild(IonRouterOutlet) outlet?: IonRouterOutlet;
+    readonly outlet = viewChild.required(IonRouterOutlet);
```
dpalou, line 42: This view child wasn't required before. Does it need to be required now?

If we want to make it required then you can probably remove the code to check if it exists or not.

src/core/features/course/classes/main-activity-page.ts
```diff
@@ -26,7 +26,7 @@ import { CoreAlerts } from '@services/overlays/alerts';
 })
 export class CoreCourseModuleMainActivityPage<ActivityType extends CoreCourseModuleMainResourceComponent> implements OnInit {
 
-    activityComponent?: ActivityType;
+    readonly activityComponent = signal<ActivityType | undefined>(undefined).asReadonly();
```
dpalou, line 33: I would add a comment here explaining that this should be overridden with a viewChild in the child classes, it looks weird if you don't know that.

src/core/features/editor/components/rich-text-editor/rich-text-editor.ts
```diff
@@ -92,7 +92,7 @@ export class CoreEditorRichTextEditorComponent implements AfterViewInit, OnDestr
     @Input() draftExtraParams?: Record<string, unknown>; // Extra params to identify the draft.
     @Output() contentChanged: EventEmitter<string | undefined | null> = new EventEmitter();
 
-    @ViewChild(CoreDynamicComponent) dynamicComponent!: CoreDynamicComponent<CoreEditorBaseComponent>;
+    readonly dynamicComponent = viewChild.required(CoreDynamicComponent);
```
dpalou: Lost type checking here, please add CoreEditorBaseComponent.

src/core/features/block/components/side-blocks/side-blocks.ts
```diff
@@ -118,7 +119,12 @@ export class CoreBlockSideBlocksComponent implements OnInit {
     async doRefresh(refresher?: HTMLIonRefresherElement): Promise<void> {
         await CorePromiseUtils.ignoreErrors(this.invalidateBlocks());
 
-        await this.loadContent().finally(() => {
+        await this.loadContent().finally(async () => {
+            await CorePromiseUtils.allPromisesIgnoringErrors(
+                this.blocksComponents()?.map((blockComponent) =>
+                    blockComponent.reload()),
+            );
```
dpalou, line 126: If I understood it right, this change was needed because, due to the new `for` syntax, the block components aren't re-created.

If that's the case, site plugins blocks might not work as expected now. Site plugins don't fetch any data in the block itself, they render the core-site-plugins-plugin-content component and it's this component the one fetching the info. This means the `reload` function will do nothing for site plugins.

I did a couple of tests in Dashboard and it worked fine for me: the `core-block` component is not re-created, but all the dynamic components are re-created due to our `ngOnChanges` code in core-block, so the site plugin worked fine. But this made me think that the `fetchContent` function might be triggered twice when doing PTR: once when the component is re-created, and once due to reload function.

I tried this and in webapp the `fetchContent` function was only called once (at least in Dashboard) because, when calling `reload`, `this.dynamicComponent()` is already undefined so the reload function did nothing in that case. This means the new reload code in Dashboard is useless (at least in my webapp, but maybe it depends on race conditions or something).

This behaviour is confusing and seems quite fragile. If the `core-block` component isn't re-created and the block class to render hasn't changed, then we probably shouldn't re-create the component inside core-block, to make it consistent. In that case I guess the sub-components wouldn't be re-created and the reload function would be the one responsible to update the data (and we would need to implement the reload function of site plugins blocks to make them refetch the data too).

Please take a look at this :)

src/core/classes/tabs.ts
```diff
@@ -120,6 +102,22 @@ export class CoreTabsBaseComponent<T extends CoreTabBase> implements AfterViewIn
...
+            const swiper = CoreSwiper.initSwiperIfAvailable(this.swiper(), this.swiperRef(), this.swiperOpts);
+            if (!swiper) {
+                return;
+            }
+
+            this.swiper.set(swiper);
+
+            // Subscribe to changes.
+            swiper.on('slideChangeTransitionEnd', () => NgZone.run(() => this.slideChanged()));
+        });
```
dpalou: This effect is reading and setting the same signal (`this.swiper`), which is something that isn't recommended at all. We probably should re-think this code and check if some of this code is still needed.

E.g. if we're getting a new swiperRef, can we still have an old swiper that is still valid? I guess it was possible before, otherwise the code probably wouldn't be there, but maybe with the new signals it can no longer happen, I don't know.

Please look at the next comment too.

src/core/classes/tabs.ts
```diff
@@ -120,6 +102,22 @@ export class CoreTabsBaseComponent<T extends CoreTabBase> implements AfterViewIn
...
+
+        effect(() => {
+            this.init();
+        });
```
dpalou: We should probably re-think / re-evaluate this part too. E.g. we have an ngAfterViewInit that calls "init", but maybe now we no longer need it if we're using signals to react to changes. It also looks weird that we keep calling a function called "init" when the component has already been initialized (this also happened before, it's not new).

Taking this comment and the previous one into account, IMO it's better to leave the component as it was for now, and in a future release try to migrate it to signals completely and refactor some code.

src/core/components/iframe/iframe.ts
```diff
@@ -64,11 +65,13 @@ export class CoreIframeComponent implements OnChanges, OnDestroy {
 
     static loadingTimeout = 15000;
 
-    @ViewChild('iframe') set iframeElement(iframeRef: ElementRef | undefined) {
-        this.iframe = iframeRef?.nativeElement;
+    readonly iframeRef = viewChild<ElementRef<HTMLIFrameElement>>('iframe');
+    readonly iframe = computed(() => {
+        const iframe = this.iframeRef()?.nativeElement;
+        this.initIframeElement(iframe);
```
dpalou: I was wondering whether the "initFrameElement" should be part of the computation or not in terms of code organization. E.g. instead of calling it here, the computed could just do:

`readonly iframe = computed(() => this.iframeRef()?.nativeElement);`

And then you can have an `effect` listening to `this.iframe()` and calling `initIframeElement` when it changes. This way the computation function is simpler and only calculates the variable itself, and other code that depends on the variable is placed in an effect. What do you think about this?

src/core/components/swipe-slides/swipe-slides.ts
```diff
@@ -97,6 +77,31 @@ export class CoreSwipeSlidesComponent<Item = unknown> implements OnChanges, OnDe
         this.resizeListener = CoreDom.onWindowResize(() => {
             this.updateSlidesComponent();
         });
+
+        effect(() => {
+            const swiper = CoreSwiper.initSwiperIfAvailable(this.swiper(), this.swiperRef(), this.options);
+            if (!swiper) {
+                return;
+            }
+            this.swiper.set(swiper);
```
dpalou: Here we have the same problem I mentioned in core-tabs (reading and writing same signal), maybe it's better to leave this component unchanged for now.

src/core/features/editor/components/classic-editor/classic-editor.ts
```diff
@@ -130,6 +116,19 @@ export class CoreEditorClassicEditorComponent extends CoreEditorBaseComponent im
...
+    constructor() {
+        super();
+
+        effect(() => {
+            const swiper = CoreSwiper.initSwiperIfAvailable(this.toolbarSlides(), this.swiperRef(), this.swiperOpts);
+            if (!swiper) {
+                return;
+            }
+
+            this.toolbarSlides.set(swiper);
```
dpalou: Reading and setting the same signal here too.

src/core/features/login/pages/reconnect/reconnect.ts
```diff
@@ -108,6 +98,19 @@ export default class CoreLoginReconnectPage implements OnInit, OnDestroy {
...
+            if (!loginMethods) {
+                return;
+            }
+
+            this.currentLogin = await CorePromiseUtils.ignoreErrors(loginMethods.getCurrentLogin());
+
+            if (this.currentLogin) {
+                effectRef.destroy();
+            }
+        },{ manualCleanup:true });
```
dpalou: I'm a bit confused about this manualCleanup. According to the documentation:

> Whether the effect should require manual cleanup.
> 
> If this is false (the default) the effect will automatically register itself to be cleaned up with the current DestroyRef.
> 
> If this is true and you want to use the effect outside an injection context, you still need to provide an Injector to the effect.

The second line makes me think that, if it's true, then the effect will NOT be destroyed automatically, and we don't want that. In a video tutorial I found (not official), they said the same thing, using manualCleanup=true means the effect will not be destroyed automatically.

So I decided to try it. I created an effect with manualCleanup=true in a page that listened to a signal in a service. I entered the page, I left it and the effect stoped reacting to changes in my service signal. So it was automatically destroyed even if manualCleanup was true, which contradicts what was said before.

However, I also tried using effectRef.destroy() in an effect that used manualCleanup=false and it worked fine too. So I'm not sure what this `manualCleanup` is for. 

IMO we should either not use manualCleanup if it's not needed to destroy the effect, OR use manualCleanup=true but also call effectRef.destroy in the component's OnDestroy to be sure it has been destroyed. The second one is probably safer.

src/core/features/viewer/components/image/image.ts
```diff
@@ -75,6 +60,19 @@ export class CoreViewerImageComponent implements OnInit {
...
     };
 
+    constructor() {
+        effect(() => {
+            const swiper = CoreSwiper.initSwiperIfAvailable(this.swiper(), this.swiperRef(), this.swiperOpts);
+            if (!swiper) {
+                return;
+            }
+
+            this.swiper.set(swiper);
```
dpalou: This effect also reads and writes the same signal.

dpalou (changes requested):

src/core/features/courses/pages/dashboard/dashboard.ts
crazyserver: Ooops nice catch, just an error on my mind.

src/core/features/courses/services/courses-helper.ts
crazyserver: Comment removed. Work will continue in MOBILE-4537

src/core/features/course/services/course-completion-helper.ts
crazyserver: I agree! Helper is not needed if we split the services wisely.

src/addons/block/timeline/components/timeline/timeline.ts
crazyserver: I agree with you but this is a largest change that implies changing the block base component to change loaded an so on.

I added the readonly modifier since signal based forms are still experimental and we can change that in the future.

src/addons/block/timeline/components/timeline/timeline.ts
crazyserver: As mentioned above: I agree with you but this is a largest change that implies changing the block base component to change loaded an so on.

src/addons/block/timeline/components/timeline/timeline.ts
crazyserver: I changed it to showUpdateToast property.

src/addons/calendar/components/calendar/calendar.ts
crazyserver: true...

src/addons/messages/pages/group-conversations/group-conversations.ts
crazyserver: Yes, it works correctly.

src/core/features/login/pages/reconnect/reconnect.ts
crazyserver: After reading https://angular.dev/guide/signals#destroying-effects
I can ensure effects can be destroyed without manualCleanup

dpalou (approved):

PR #4541, opened 2025-08-07
MOBILE-4862 opener.ts: Handle error when open file

alfonso-salces:


src/core/singletons/opener.ts
```diff
@@ -124,18 +125,26 @@ export class CoreOpener {
...
                 if (CoreOpener.shouldOpenWithDialog(options)) {
                     await FileOpener.showOpenWithDialog(path, mimetype || '');
                 } else {
                     await FileOpener.open(path, mimetype || '');
                 }
             } catch (error) {
+                if (
+                    hasFailed ||
+                    (error.status !== CoreNativeCordovaPluginResultStatus.ERROR && error.message.includes('Activity not found'))
+                ) {
```
dpalou: I think this if condition is wrong. If I understood it right when we talked about this, we want to encode the path in case the status is ERROR and the message isn't "Activity not found", which would be (pseudo code):

status === ERROR && !message.includes('Activity not found')

Since you're using an if guard, you want to negate the previous condition. Which would be:

status !== ERROR || message.includes('Activity not found')

So the whole if condition would be:

hasFailed || status !== ERROR || message.includes('Activity not found')

dpalou (changes requested):

dpalou (approved):

PR #4553, opened 2025-09-08
Mobile 4868

dpalou:


src/addons/mod/h5pactivity/services/handlers/report-link.ts
```diff
@@ -77,6 +71,31 @@ export class AddonModH5PActivityReportLinkHandlerService extends CoreContentLink
...
 
+    /**
+     * Get the module basic info for the activity.
+     *
+     * @param cmId Course module ID, NaN if not valid.
+     * @param instanceId Activity instance ID, NaN if not valid.
+     * @param siteId Site ID.
+     * @returns Module basic info.
+     */
+    protected getModuleBasicInfo(cmId: number, instanceId: number, siteId: string): Promise<CoreCourseModuleBasicInfo> {
```
crazyserver, line 82: I'd create a general function with the module name as parameter because I think this is not the only case like this.

src/core/features/course/pages/overview/overview.scss
```diff
@@ -31,4 +31,9 @@
         --background: var(--danger-tint);
         --color: var(--danger-shade);
     }
+
+    table.core-table {
+        margin-top: 0;
```
crazyserver: This is correct but I prefer always adding units to avoid calcs failing when they are used in vars (this is not our case).

crazyserver (changes requested):

src/addons/mod/h5pactivity/services/handlers/report-link.ts
dpalou: I looked at the code and I think the only other places where something similar happens is in some view.php links, and that is handled in the `getActions` function of `CoreContentLinksModuleIndexHandler`. In that case it uses `navigateToModule` or `navigateToModuleByInstance` instead of `getModuleBasicInfo`, it's not the same usage. Since right now this is the only case like this IMO we can leave it and if we find other places in the future we can create a shared function.

crazyserver (approved):

PR #4554, opened 2025-09-09
Mobile 4842

crazyserver:


src/core/features/course/classes/base-course-format-component.ts
```diff
@@ -0,0 +1,43 @@
...
+import { CoreCourseSection, CoreCourseModuleCompletionData } from '../services/course-helper';
+import { CoreCourseSectionToDisplay } from '../components/course-section/course-section';
+
+/**
+ * Template class to easily create components for course formats.
+ */
+@Component({
+    template: '',
+})
+export abstract class CoreCourseFormatBaseComponent {
```
dpalou: This name can be confusing IMO, because it made me think that `CoreCourseFormatComponent` should also extend this class because it's a component about course format. But this class is only for the "subcomponents" or "dynamic components" used inside `CoreCourseFormatComponent`.

The "subcomponents" used in `CoreCourseFormatComponent` which should implement this class are `courseFormatComponent`, `singleSectionComponent` and `allSectionsComponent`.

Maybe we should name this class `CoreCourseFormatDynamicComponent` or something like this. What do you think? You can include the "Base" word if you want to.

dpalou: Also, if we expect all the subcomponents to implement this class then it might be interesting to specify it in `CoreCourseFormatHandler`(getCourseFormatComponent, getSingleSectionComponent, getAllSectionsComponent) and the corresponding functions in `CoreCourseFormatDelegateService`. This is done for example in the assign submission delegate or in the database field delegate.

src/addons/badges/pages/badge-class/badge-class.html
```diff
@@ -4,77 +4,78 @@
...
 </ion-header>
 <ion-content class="limited-width">
-    <ion-refresher slot="fixed" [disabled]="!badgeLoaded" (ionRefresh)="refreshBadgeClass($event.target)">
+    <ion-refresher slot="fixed" [disabled]="!loaded()" (ionRefresh)="refreshBadgeClass($event.target)">
         <ion-refresher-content pullingText="{{ 'core.pulltorefresh' | translate }}" />
     </ion-refresher>
-    <core-loading [hideUntil]="badgeLoaded" placeholderType="imageandboxes">
-        @if (badge) {
+    <core-loading [hideUntil]="loaded()" placeholderType="imageandboxes">
+        @if (badgeToShow) {
```
dpalou, line 21: I guess it works, but it looks a bit weird that we use a variable which is declared in a different DOM tree. We might want to re-declare the `badgeToShow` variable here, or move the let to the top of the file so it's on the same tree as both usages.

src/addons/mod/feedback/pages/form/form.ts
```diff
@@ -93,13 +92,9 @@ export default class AddonModFeedbackFormPage implements OnInit, OnDestroy, CanL
...
-        // Refresh online status when changes.
-        this.onlineObserver = CoreNetwork.onChange().subscribe(() => {
-            // Execute the callback in the Angular zone, so change detection doesn't stop working.
-            NgZone.run(() => {
-                this.offline = !CoreNetwork.isOnline();
-            });
-        });
+        effect((() => {
+            this.offline.set(!this.isOnline());
+        }));
```
dpalou: Instead of an effect, you can use a `linkedSignal`. linkedSignal is like a computed signal, but then you can override the value.

src/addons/mod/h5pactivity/components/index/index.ts
```diff
@@ -126,47 +124,6 @@ export class AddonModH5PActivityIndexComponent extends CoreCourseModuleMainActiv
...
-            return;
-        }
-
-        if (this.isOnline && this.triedToPlay) {
-            // User couldn't play the package because he was offline, but he reconnected. Try again.
-            this.triedToPlay = false;
-            this.play();
-
-            return;
-        }
```
dpalou, line 169: Why did you delete all this code? It was important to improve the UX for users that lose connection or re-connect.

src/core/services/network.ts
```diff
@@ -326,6 +326,24 @@ export class CoreNetworkService extends Network {
...
+    }
+
+    /**
+     * Returns a signal to watch if the device uses a wifi connection.
+     *
+     * @returns Signal.
+     */
+    isWifiSignal(): Signal<boolean> {
+        return computed<boolean>(() => this.connectionTypeSignal()() === CoreNetworkConnectionType.WIFI);
+    }
```
dpalou, line 347: Related to this and some existing code in this service, some weeks ago we talked about signals in CoreKeyboard and we decided to use a getter to retrieve a signal value, because we introduced a bug by mistake because we only used one parenthesis instead of 2. E.g. we used `keyboardShownSignal()` instead of `keyboardShownSignal()()`. With the getter, to obtain the value you only need to do: `keyboardShownSignal()`.

I guess this is something we should do in this service (and other services where we are using signals, if any) for concistency. So if `onlineSignal` is a getter, then in the components where you want to obtain the signal you need to do `CoreNetwork.onlineSignal` (instead of `CoreNetwork.onlineSignal()`), and if you want to obtain the value then you can read it with `CoreNetwork.onlineSignal()` instead of `CoreNetwork.onlineSignal()()`. The same applies to `connectionTypeSignal`.

dpalou, line 347: Also in this code, every time `isCellularSignal` or `isWifiSignal` are called you're creating a new computed signal. Wouldn't it be better to have these computed signals as properties in the service and then just return them? With a getter as I mention in the previous comment.

This change means we'll always be calculating these values even if they aren't being used, but they're very simple so IMO it's not a big deal. And that allows you to use these functions anywhere, not just in a constructor or injection context like it happens with your implementation.

src/core/services/screen.ts
```diff
@@ -66,6 +66,7 @@ export class CoreScreenService {
 
     protected breakpointsSubject: BehaviorSubject<Record<Breakpoint, boolean>>;
     private _layoutObservable: Observable<CoreScreenLayout>;
+    readonly orientationSignal = signal<CoreScreenOrientation>(this.orientation);
```
dpalou: Do we want this signal to be public? We usually make signals in services protected or private and then we expose as readonly using a getter/function. Otherwise anyone can change the value of this signal.

src/addons/coursecompletion/pages/report/report.html
```diff
@@ -86,24 +92,25 @@ <h2>{{ 'addon.coursecompletion.requiredcriteria' | translate }}</h2>
...
-                                <ion-col>
+                            <ion-col>
+                                @if (criteria.complete) {
+                                    {{ 'core.yes' | translate }}
+                                } @else {
+                                    {{ 'core.no' | translate }}
+                                }
+                            </ion-col>
+                            <ion-col>
+                                @if (criteria.timecompleted) {
```
dpalou, line 103: Before this change, if there was no timecompleted, the ion-col wasn't displayed. Now it will be displayed but empty. Is that ok? Just to double check it's a deliberate decision.

src/addons/coursecompletion/pages/report/report.ts
```diff
@@ -39,26 +40,56 @@ import { CoreSharedModule } from '@/core/shared.module';
...
     constructor() {
+        try {
+            this.courseId = CoreNavigator.getRequiredRouteNumberParam('courseId');
+            this.userId.set(CoreNavigator.getRouteNumberParam('userId') || CoreSites.getCurrentSiteUserId());
+        } catch (error) {
+            CoreAlerts.showError(error);
+            CoreNavigator.back();
+
+            return;
+        }
```
dpalou: AFAIK we always read parameters in _ngOnInit_, since that's when the Angular component is considered "ready to be initialized". I'm not sure if we have any case where we read route params in constructor, or if this can cause problems (e.g. parameters not ready in the constructor).

I would keep the code in _ngOnInit_ just in case, and also to keep it consistent. A third reason is that we usually try not to call services in constructor, because that can cause "true" circular dependencies at runtime, breaking the app.

I guess you moved this code in the constructor to be able to define courseId as a constant. I would just define the courseId as a signal and that's it, even if it doesn't change later.

src/addons/mod/h5pactivity/components/index/index.ts
crazyserver: My fault, IDK why I decided it was not necessary... :'(

src/core/services/network.ts
crazyserver: OK! I added a new commit for that

src/core/services/screen.ts
crazyserver: Done.
I've also added the deprecation of the event and usage of the signal in the same commit.

src/addons/coursecompletion/pages/report/report.html
crazyserver: Hmmm you're wrong there was an else with an empty ion-col. But in any case it's done this way to make the number of ion-cols match the header row.

src/addons/coursecompletion/pages/report/report.ts
crazyserver: OK, changed. FTR, we use CoreNavigator to get params from the constructor in many places.

src/core/components/iframe/iframe.ts
```diff
@@ -89,7 +88,7 @@ export class CoreIframeComponent implements OnChanges, OnDestroy {
 
     protected fullScreenInitialized = false;
     protected style?: HTMLStyleElement;
-    protected orientationObs?: CoreEventObserver;
+    protected watchOrientation = false;
```
dpalou: This variable doesn't seem to be used.

src/core/services/screen.ts
```diff
@@ -66,6 +66,7 @@ export class CoreScreenService {
 
     protected breakpointsSubject: BehaviorSubject<Record<Breakpoint, boolean>>;
     private _layoutObservable: Observable<CoreScreenLayout>;
+    protected readonly orientationSignalInternal = signal<CoreScreenOrientation>(this.orientation);
```
dpalou: i'm not a big fan of naming it "Internal". Maybe it would be better to change the variable to private and name it "`_orientationSignal`" (this is a common convention in JS code).

The downside of this change is that, if we ever want to create a subclass of this service, we won't be able to use `_orientationSignal` directly, but I don't think we'll need that.

dpalou (changes requested):

PR #4558, opened 2025-09-17
Mobile 4891

crazyserver:


src/core/singletons/bootstrap.ts
```diff
@@ -420,6 +421,160 @@ export class CoreBootstrap {
...
+            });
+
+            // Resume on mouse leave.
+            element.addEventListener('mouseleave', () => {
+                isPaused = false;
+            });
+
+            // Start the interval to change slides.
+            if (interval > 0) {
+                setInterval(() => {
```
albertgasset: Bootstrap does not enable autoplay by default, it is only enabled when `data-bs-ride="carousel"`

src/core/singletons/bootstrap.ts
```diff
@@ -420,6 +421,180 @@ export class CoreBootstrap {
...
+                        }
+                    };
+
+                    // Listen for first interaction: click, touch, or keyboard.
+                    element.addEventListener('click', startAutoplay, { once: true });
+                    element.addEventListener('touchstart', startAutoplay, { once: true });
+                    element.addEventListener('keydown', startAutoplay, { once: true });
+                }
+
+                setInterval(() => {
```
albertgasset: The setInterval is never cleared. We could check in the callback if the element is still connected to the document and clear the interval if it is not.

crazyserver: Thanks Albert! Nice catch!

albertgasset (approved):

PR #4575, opened 2025-10-01
MOBILE-4902 format-text: Use signals on format-text to fix render

crazyserver:
There was a problem with rendering emoji on iOS on message discussions. Those emojis were not rendered until something change in the page.

### DIFF
--- a/src/core/directives/format-text.ts
+++ b/src/core/directives/format-text.ts
@@ -476,8 +476,19 @@ export class CoreFormatTextDirective implements OnDestroy, AsyncDirective {
         // Emit the afterRender output.
         this.afterRender.emit();
         if (triggerFilterRender) {
-            this.filterContentRenderingComplete.emit();
+            this.finishFilterContentRendering();
         }
+    }
+
+    /**
+     * Finish the rendering once filters have been applied.
+     */
+    protected async finishFilterContentRendering(): Promise<void> {
+        // Force redraw to make sure emojis are displayed on iOS.
+        if (CorePlatform.isIOS() && CoreText.containsEmoji(this.element.innerText)) {
+            await CoreDom.forceElementRedraw(this.element);
+        }
+        this.filterContentRenderingComplete.emit();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/core/directives/format-text.ts
```diff
@@ -124,25 +219,29 @@ export class CoreFormatTextDirective implements OnChanges, OnDestroy, AsyncDirec
...
+        effect(() => {
+            // Refresh the contents if any of the relevant inputs change.
+            this.textComputed();
+            this.filter();
+            this.contextLevel();
+            this.contextInstanceIdComputed();
 
-        if ('disabled' in changes) {
-            const disabled = changes['disabled'].currentValue;
+            this.formatAndRenderContents();
```
albertgasset: The signals read inside this function are tracked by the effect, including the signals read in `formatContents`. The only signals that are not tracked are those read after the first `await` is executed. 

To preserve the exact previous behaviour, this function must be called in a non-tracking context using  `untracked`:
```
untracked(() => this.formatAndRenderContents());
```

src/core/directives/format-text.ts
```diff
@@ -73,38 +74,132 @@ import { CoreBootstrap } from '@singletons/bootstrap';
...
+
+        return text ? text.trim() : '';
+    });
+
+    /**
+     * Site ID to use. If not defined, current site.
+     */
+    readonly siteId = input<string>();
+    protected readonly siteIdComputed = computed(() => this.site()?.getId());
+    protected readonly site = signal<CoreSite | undefined>(undefined);
```
albertgasset: Signals `site` and `siteIdComputed` are not updated in sync with the other signals. This can cause issues: reading the signal with an old value, and repeated execution of an effect (with the old and new value).
In this case, I think it's better to just use helper async functions instead of signals.

crazyserver:
Thanks Albert for your review, I've changed what you said and now is working perfectly.

However, I've added a 3rd commit because an error was showing on the console. This is because dynamic-component now sets the inputs of the components but it cannot check if it's an output.

This is the error message type:

dynamic-component.ts:222 NG0303: Can't set value of the 'onAbort' input on the 'AddonQtypeDdImageOrTextComponent' component. Make sure that the 'onAbort' property is declared as an input using the input() or model() function or the @Input() decorator.
 Could you review also this part?

albertgasset (approved):

PR #4710, opened 2026-02-04
Mobile 4948 Test on non pure pipes

crazyserver:


src/core/pipes/tests/bytes-to-size.pipe.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+        expect(pipe.transform(500)).toContain('bytes');
+        // Simulate language change
+        setTestTranslations({
+            'core.sizeb': 'grepe',
+            'core.sizegb': 'GB',
+            'core.sizekb': 'KB',
+            'core.sizemb': 'MB',
+            'core.sizetb': 'TB',
+        });
```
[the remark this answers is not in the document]
crazyserver: This is merging translations, so it's not needed

src/core/pipes/tests/date-day-or-time.pipe.test.ts
```diff
@@ -0,0 +1,47 @@
...
+            'core.strftimetime': 'NOW',
+        });
+    });
+
+    it('should return correct day or time in default language', () => {
+        const pipe = new CoreDateDayOrTimePipe();
+        const now = Math.floor(Date.now() / 1000);
+
+        expect(pipe.transform(now)).toEqual('NOW');
+    });
```
dpalou, line 40: Is it hard to test other values? Since we're writing unit tests for the pipe it would be nice to test the different cases.

src/core/pipes/tests/duration.pipe.test.ts
```diff
@@ -0,0 +1,72 @@
...
+import { setTestTranslations } from '@/testing/utils';
+import { CoreLang } from '@services/lang';
+import { CoreTime } from '@static/time';
+
+describe('CoreDurationPipe', () => {
+    beforeEach(async () => {
+        await CoreTime.initialize();
+
+        setTestTranslations({
+            'core.now': 'now',
```
dpalou: Is the 'core.now' string used in this pipe?

src/core/pipes/tests/bytes-to-size.pipe.test.ts
dpalou: Since we're only using the 'core.sizeb' string in this test case, maybe we should just set that value in 'af' language instead of all the strings to decrease the number of lines of the test?

The same happens in other pipes. If you prefer it like this it's ok for me too.

Please read the next comments before changing this.

src/core/pipes/tests/bytes-to-size.pipe.test.ts
```diff
@@ -0,0 +1,55 @@
...
+describe('CoreBytesToSizePipe', () => {
+    beforeEach(() => {
+        setTestTranslations({
+            'core.sizeb': 'bytes',
+            'core.sizekb': 'KB',
+            'core.sizemb': 'MB',
+            'core.sizegb': 'GB',
+            'core.sizetb': 'TB',
+            'core.humanreadablesize': '{{size}} {{unit}}',
+        });
```
dpalou: I was just thinking that it seems a bit annoying to have to define all the strings for the test, specially if you want to change language and set the translations again. IMO it makes the test feel less realistic (but more isolated). Wouldn't it be better to use the real strings of the app?

dpalou (changes requested):

dpalou (approved):

dpalou (approved):

PR #4714, opened 2026-02-09
Mobile 5002

crazyserver:


src/core/components/alert-card/alert-card.ts
```diff
@@ -0,0 +1,103 @@
...
+    @Output() itemAction = new EventEmitter<Event>(); // Will emit an event when the item clicked.
+
+    readonly hasAction = signal(false);
+
+    /**
+     * @inheritdoc
+     */
+    ngOnInit(): void {
+        this.hasAction.set(this.itemAction.observed);
+    }
```
dpalou: ngOnInit should be below all the computed signals, since computed signals are properties and this is a function.

src/core/components/alert-card/alert-card.ts
```diff
@@ -0,0 +1,103 @@
...
+
+    readonly type = input<CoreAlertCardType | string>(CoreAlertCardType.WARNING);
+    readonly icon = input<string | undefined>();
+    readonly title = input('');
+    readonly message = input('');
+    readonly dismissible = input(false, { transform: toBoolean });
+
+    // New output syntax doesn't have the 'observed' property, keep EventEmitter for now.
+    // See https://github.com/angular/angular/issues/54837
+    @Output() itemAction = new EventEmitter<Event>(); // Will emit an event when the item clicked.
```
dpalou: I was now reviewing the linked Angular issue and they closed it as "not planned". And their recommendation is to use `outputFromObservable`.

Can you please check if that works for us? If so, we should use it instead of EventEmitter.

src/core/components/alert-card/alert-card.ts
```diff
@@ -0,0 +1,103 @@
...
+            case CoreAlertCardType.DANGER:
+                return 'fas-circle-exclamation';
+            case CoreAlertCardType.INFO:
+                return 'fas-circle-info';
+            case CoreAlertCardType.SUCCESS:
+                return 'fas-circle-check';
+            case CoreAlertCardType.WARNING:
+            default:
+                return 'fas-triangle-exclamation';
+        }
```
dpalou, line 69: This is just a personal opinion, but I prefer adding blank lines between different if / for / switch. But you can leave it like this if you prefer.

src/core/components/alert-card/alert-card.ts
dpalou: Also, I'm not a fan of this name. You call it "itemAction", but:

1. This is not an item, it's a card/alert.
2. What is the action? Is it click? Is it something else?

I suggest renaming it. The first thing that came into my mind to make it simple was "clicked", but it can be confused with Angular's (click). So I was now thinking to name it "alertClick" even if it's a bit redundant. What do you think?

dpalou (changes requested):

src/core/components/alert-card/alert-card.ts
crazyserver: I removed this action it won't be needed

src/core/components/alert-card/alert-card.ts
crazyserver: Removed

src/core/components/alert-card/alert-card.ts
crazyserver: This has been removed but I applied the same pattern in the other places to have them resolved

src/core/directives/supress-events.ts
```diff
@@ -84,12 +87,12 @@ export class CoreSupressEventsDirective implements OnInit {
...
         this.element.addEventListener('mouseup', (event) => { // Triggered in Android & iOS.
-            this.onClick.emit(event);
+            this.onClickEmitter.next(event);
         });
 
         this.element.addEventListener('touchend', (event) => { // Triggered desktop & browser.
             this.stopBubble(event);
-            this.onClick.emit(event);
+            this.onClickEmitter.next(event);
         });
```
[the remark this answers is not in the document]
crazyserver: I don't think this is necessary

dpalou (approved):

PR #4719, opened 2026-02-17
MOBILE-5006: Add customisations for custom menu items and settings handler

crazyserver:
This pull request addresses [MOBILE-5006](https://moodle.atlassian.net/browse/MOBILE-5006).

Jira Issue Summary:
Add customisations for custom menu items and settings handler

Jira Issue Description:
Add customisations for custom menu items and settings handler

Assignee: Pau Ferrer
Status: Development in progress

Summary of changes (last 3 commits):
- settings: Add the possibility to show a badge on settings
- customitems: Custom menu item can be overridden
- modals: Add border radius to modals

Please review the changes related to the settings badge, custom menu item override, and modal border radius improvements.

[MOBILE-5006]: https://moodle.atlassian.net/browse/MOBILE-5006?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/core/features/mainmenu/services/custommenu.ts
```diff
@@ -207,6 +207,16 @@ export class CoreCustomMenuService {
             }));
     }
 
+    /**
+     * Get the class of the component to render for custom menu items. If not defined, the default component will be used.
+     * This method can be overridden by other services to provide a custom component to render the custom menu items.
```
dpalou: The only way "other services" can override this is if they're child classes. So IMO we should improve this wording a bit because it can be confusing, because if I don't know about the app and I read this I would think that any service can override this.

src/core/features/settings/services/settings-delegate.ts
```diff
@@ -61,12 +61,42 @@ type CoreSettingsPageHandlerData = CoreSettingsHandlerBaseData & {
     /**
      * Name of the page to load for the handler.
      */
-    page: string;
+    page: string | undefined;
```
dpalou: Since `page` is optional now and you can provide an `action`, IMO we should improve the types to make the possible combinations clear, as I did in user handlers for example. As I understand there are 3 possible cases: toggle, page or action. In that case we should have 3 types, and each one of them would have the right property mandatory.

dpalou (changes requested):

dpalou (approved):

PR #4739, opened 2026-03-04
Mobile 4931 Disable collapsible header when there is an active split view

crazyserver:


src/core/directives/collapsible-header.ts
```diff
@@ -75,53 +85,76 @@ export const COLLAPSIBLE_HEADER_UPDATED = 'collapsible_header_updated';
...
 
-        this.init();
+            untracked(() => {
+                this.init();
+                this.setEnabled();
+            });
+        });
+
+        effect(() => {
+            this.calculateContentWidth();
```
dpalou: This works, but it looks a bit weird because the signals reading is done inside the function so it's not easy to see what is this effect reacting to. I would suggest doing:


```suggestion
            this.calculateContentWidth(this.content(), this.splitViewMode());
```

dpalou (changes requested):

dpalou (approved):

PR #4743, opened 2026-03-05
Mobile 5019

crazyserver:


src/addons/mod/bigbluebuttonbn/components/index/index.ts
```diff
@@ -123,6 +124,35 @@ export class AddonModBBBIndexComponent extends CoreCourseModuleMainActivityCompo
                     this.meetingInfo.statusmessage = Translate.instant('addon.mod_bigbluebuttonbn.userlimitreached');
                 }
             }
+
+            // If the module doesn't include activity dates, populate them from meetingInfo.
+            // As of LMS v5.1.0, these dates are normally provided by the module.
+            if (this.module && (!this.module.dates || !this.module.dates.length)) {
```
dpalou, line 130: dates can be empty if a module doesn't have an open or close date. IMO it would be better to check the LMS version.

dpalou (changes requested):

dpalou (approved):

PR #4758, opened 2026-03-26
Mobile 4974 update

crazyserver:


package.json
```diff
@@ -71,7 +71,7 @@
         "@awesome-cordova-plugins/sqlite": "^8.1.0",
         "@awesome-cordova-plugins/status-bar": "^8.1.0",
         "@awesome-cordova-plugins/web-intent": "^8.1.0",
-        "@ionic/angular": "^8.7.17",
+        "@ionic/angular": "8.8.1",
```
dpalou: Why are you fixing the version? IMO we should keep it with "^" until we freeze them.

package.json
```diff
@@ -148,22 +148,22 @@
...
+        "fs-extra": "^11.3.4",
         "globals": "^16.4.0",
         "gulp": "^5.0.1",
         "gulp-clip-empty-files": "^0.1.2",
         "gulp-slash": "^1.1.3",
-        "jest": "^30.2.0",
+        "jest": "^30.3.0",
         "jest-environment-jsdom": "29.7.0",
-        "jest-preset-angular": "^16.0.0",
+        "jest-preset-angular": "16.1.1",
```
dpalou: Same here, you're fixing the version.

dpalou (changes requested):

dpalou (approved):

PR #4765, opened 2026-04-07
MOBILE-5022 calendar: Erase local data and reminders when deleting event

crazyserver:


src/addons/calendar/services/calendar.ts
```diff
@@ -184,6 +184,9 @@ export class AddonCalendarProvider {
         try {
             await this.deleteEventOnline(eventId, deleteAll, siteId);
 
+             // Delete the event from local DB and cancel reminders/notifications.
+            await CorePromiseUtils.ignoreErrors(this.deleteLocalEvent(eventId, siteId));
```
dpalou: I guess you should do the same in calendar-sync, after calling `AddonCalendar.deleteEventOnline`. 

EDIT: I was now thinking that we have the risk to forget calling `deleteLocalEvent` after `deleteEventOnline`. I guess it would be better to put this line of code inside `deleteEventOnline`, that way we won't forget it.

dpalou (changes requested):

src/addons/calendar/services/calendar.ts
```diff
@@ -871,6 +871,25 @@ export class AddonCalendarProvider {
...
+     * @param initialTime Timestamp when the first fetch was done. If not defined, current time.
+     * @param daysToStart Number of days from now to start getting events.
+     * @param daysInterval Number of days between timestart and timeend.
+     * @param options Options for the request.
+     * @returns Promise to be resolved when the events are retrieved.
+     */
+    async getEventsList(
+        initialTime?: number,
+        daysToStart?: number,
+        daysInterval?: number,
```
dpalou: Since all parameters are optional, in the new signature IMO it would be better to put them all inside the options object. To avoid having to do: `getEventsList(undefined, undefined, undefined, ...)`

dpalou (changes requested):

dpalou (approved):

PR #4772, opened 2026-04-09
Mobile 4974

crazyserver:


dpalou (dismissed):

src/core/directives/collapsible-footer.ts
```diff
@@ -84,6 +91,15 @@ export class CoreCollapsibleFooterDirective implements OnInit, OnDestroy {
         this.listenEvents();
     }
 
+    /**
+     * Get the height of the collapsible footer.
+     *
+     * @returns The height of the collapsible footer.
+     */
+    getHeight(): number {
```
dpalou: If I understand it correctly, this always returns the "expanded" or "full" height, not the current height (e.g. if it's collapsed it still returns the height of the footer when it was expanded). IMO a `getHeight` function should return the current height, so I suggest renaming the function to better indicate what it does (e.g. `getExpandedHeight`).

dpalou (changes requested):

dpalou (approved):

PR #4773, opened 2026-04-14
MOBILE-5030 webapp: Migrate from wasm-sqlite to wa-sqlite

dpalou:


albertgasset (approved):
The code looks good. I've done some tests and it works without issues.

PR #4781, opened 2026-04-21
Mobile 4974

crazyserver:


src/core/directives/collapsible-header.ts
```diff
@@ -135,8 +135,13 @@ export class CoreCollapsibleHeaderDirective implements OnDestroy, AsyncDirective
...
 
+    protected readonly isEnabled = computed(() =>
+        this.collapsible() && !this.forceDisabled() && !this.manuallyDisabled());
+
+    protected readonly manuallyDisabled = signal(false);
+
     get enabled(): boolean {
-        return !!this.expandedHeader && this.collapsible() && !this.forceDisabled();
+        return this.isEnabled();
     }
```
dpalou, line 156: Since now this getter only checks a signal, IMO it's better to use the signal directly because it's important to know when is a signal used or not (e.g. for tracking dependencies). If in the code I see "this.enabled" I might not know that this uses a signal behind the scenes. IMO this would only make sense for other classes, e.g. if an external class wants to know if this header is enabled, internally I would always use the signal.

I think right now this getter is not used outside of the class. But I leave it up to you if you prefer deprecating it or not, but in both cases I would stop using this.enabled inside this class.

src/core/directives/collapsible-header.ts
```diff
@@ -171,6 +176,28 @@ export class CoreCollapsibleHeaderDirective implements OnDestroy, AsyncDirective
             this.page.classList.toggle('collapsible-header-page-is-not-within-content', !isWithinContent);
         });
 
+        effect(() => {
+            const frozen = this.isFrozen();
+            this.page?.classList.toggle('collapsible-header-page-is-frozen', frozen);
```
dpalou: I see that now the class reacts to the signal. But in line 695, in the ionScrollEnd callback, you're changing the class manually too. Why is that?

src/core/directives/collapsible-header.ts
```diff
@@ -732,6 +734,8 @@ export class CoreCollapsibleHeaderDirective implements OnDestroy, AsyncDirective
             frozen = maxScrollTop + collapsedHeight <= 2 * expandedHeaderHeight;
         }
 
+        this.isFrozen.set(frozen);
```
dpalou, line 754: IMO it's a bit confusing that checkIsFrozen calculates the isFrozen value, but checkEnabled doesn't calculate the isEnabled value. IMO we need better naming.

Instead of `checkIsFrozen`, I suggest `calculateIsFrozen`. `checkEnabled` is a bit trickier to find a good name, since our plan is to refactor the class we can leave that one as it is if we don't find a better name.

dpalou (changes requested):

dpalou (approved):

PR #4813, opened 2026-06-03
Mobile 5053 System reports

crazyserver:


src/core/features/reportbuilder/pages/report/report.ts
```diff
@@ -30,14 +30,21 @@ import { CoreReportBuilderReportDetailComponent } from '../../components/report-
...
     constructor() {
         try {
-            this.reportId.set(CoreNavigator.getRequiredRouteParam('id'));
+            const reportId = CoreNavigator.getRequiredRouteNumberParam('id');
+            this.reportId.set(reportId);
         } catch {
             // No id, it should be a system report.
+            const params = CoreNavigator.getRequiredRouteParam<CoreReportbuilderSystemReportParams>('params');
+
+            this.reportParams.set(params);
```
dpalou: I reportId is optional now, we shouldn't use `getRequiredNumberParam`, because it's not required.

src/core/features/reportbuilder/services/reportbuilder.ts
```diff
@@ -146,18 +146,28 @@ export class CoreReportBuilderService {
...
     async getSystemReport(
-        params: CoreReportbuilderCanViewSystemReportWSParams,
+        reportParams: CoreReportbuilderSystemReportParams,
         page?: number,
         perpage?: number,
-    ): Promise<CoreReportbuilderRetrieveSystemReportWSResponse> {
-        const report = await CoreReportBuilderWS.getSystemReport(params, { page, perpage });
+    ): Promise<CoreReportBuilderRetrieveReportMapped> {
+        const { name, context, ...rest } = reportParams;
+        void name; // Avoid unused variable error.
```
dpalou: If name isn't used in the function at all, then IMO the reportParams type should not include name.

Also, passing the parameters exactly as you receive them to the WS params object can easily cause bugs.  Because for TS, passing more parameters than expected is not a problem when using extended types, but for WS it is. If you pass a parameter that the WS doesn't expect, it breaks. So IMO it's better to manually copy each parameter to `params` instead of using the spread operator.

src/core/features/reportbuilder/components/report-detail/report-detail.ts
```diff
@@ -57,37 +59,58 @@ export class CoreReportBuilderReportDetailComponent implements OnInit {
...
+
+    readonly cardviewShowFirstTitle = computed(() => this.report()?.details?.settingsdata.cardviewShowFirstTitle ?? false);
+    readonly cardVisibleColumns = computed(() => {
+        const reportDetails = this.report()?.details;
+
+        return reportDetails?.settingsdata.cardviewVisibleColumns ?? 1;
+    });
+
+    readonly contextLevel = computed(() => this.reportParamsObject()?.context?.contextlevel ?? ContextLevel.SYSTEM);
+    readonly contextInstanceId = computed(() => this.reportParamsObject()?.context?.instanceid ?? 0);
```
dpalou, line 81: Before this change, all reports had a "contextLevel" and "contextInstanceId". Now only the system report will have them, because if I'm not mistaken only system reports uses this.reportParamsObject(). Is that ok?

src/core/features/reportbuilder/components/report-detail/report-detail.ts
```diff
@@ -155,7 +161,7 @@ export class CoreReportBuilderReportDetailComponent implements OnInit {
                         role: 'confirm',
                         handler: async () => {
                             const site = CoreSites.getRequiredCurrentSite();
-                            const href = `${site.getURL()}/reportbuilder/view.php?id=${this.reportId}`;
+                            const href = `${site.getURL()}/reportbuilder/view.php?id=${this.reportId()}`;
```
dpalou: Here, you're passing reportId() but for system reports this will be undefined. I guess the href should be different depending on whether you're seeing a system report or not?

src/core/features/reportbuilder/services/reportbuilder.ts
```diff
@@ -202,6 +212,7 @@ export type CoreReportBuilderReportDetailSettingsData = {
 
 export type CoreReportBuilderReport = CoreReportBuilderReportWSResponse;
 
-export type CoreReportbuilderSystemReportParams = {
+export type CoreReportbuilderSystemReportParams = Omit<Partial<CoreReportbuilderCanViewSystemReportWSParams>, 'source'> & {
```
dpalou: This should be CoreReportBuilder (b uppercase), not CoreReportbuilder.

src/core/features/reportbuilder/pages/list/list.ts
```diff
@@ -33,42 +30,33 @@ import { CoreSharedModule } from '@/core/shared.module';
...
-
-    readonly state = signal<CoreReportBuilderListState>({
-        page: 1,
-        perpage: REPORTS_LIST_LIMIT,
-        loaded: false,
-        loadMoreError: false,
-    });
+    readonly reports = signal<CoreReportBuilderReport[]>([]);
+    readonly page = signal(0);
+    readonly perpage = signal(REPORTS_LIST_LIMIT);
```
dpalou: IMO it should be perPage instead of perpage.

src/core/features/reportbuilder/pages/list/list.ts
```diff
@@ -33,42 +30,33 @@ import { CoreSharedModule } from '@/core/shared.module';
...
-    reports!: CoreListItemsManager<CoreReportBuilderReport, CoreReportBuilderReportsSource>;
-
-    readonly state = signal<CoreReportBuilderListState>({
-        page: 1,
-        perpage: REPORTS_LIST_LIMIT,
-        loaded: false,
-        loadMoreError: false,
-    });
+    readonly reports = signal<CoreReportBuilderReport[]>([]);
+    readonly page = signal(0);
```
dpalou: Before this change, the initial value for page was 1, now it's 0. Is this intended?

src/core/features/reportbuilder/pages/list/list.ts
```diff
@@ -77,25 +65,31 @@ export default class CoreReportBuilderListPage implements AfterViewInit, OnDestr
...
-    protected updateState(state: Partial<CoreReportBuilderListState>): void {
-        this.state.update((previousState) => ({ ...previousState, ...state }));
+            reports.sort((a, b) => a.timemodified < b.timemodified ? 1 : -1);
+            this.reports.set(reports);
+            this.hasMoreItems.set(reports.length >= this.perpage());
+        } catch (error) {
+            CoreAlerts.showError(error, { default: 'Error loading reports' });
+
+            this.hasMoreItems.set(false);
+        } finally {
```
dpalou: Here you're capturing the error. This means that the catch in `fetchMoreReports` will never trigger, so `this.loadMoreError` will never be true.

src/core/features/reportbuilder/pages/list/list.ts
dpalou: `page` and `perPage` are not used in the template or ir any computed signal, so there's no need for them to be signals, they could be standard variables. But I guess it doesn't hurt, so if you prefer to leave them as signals then it's ok.

src/core/features/reportbuilder/pages/list/list.ts
```diff
@@ -104,17 +98,27 @@ export default class CoreReportBuilderListPage implements AfterViewInit, OnDestr
      * @param complete Completion callback.
      */
     async fetchMoreReports(complete: () => void): Promise<void> {
+        this.page.update((value) => value + 1);
```
dpalou: Here you're updating the `page` before the actual action. If the action fails, the value of page will be inconsistent because it will point to a page that hasn't been loaded. It won't cause problems right now because hasMoreItems will be set to false so the page won't be used again, but leaving variables in an inconsistent state is dangerous because it can introduce bugs in future refactors.

src/addons/badges/services/handlers/badge-systemreport-link.ts
```diff
@@ -0,0 +1,107 @@
...
+        const pageParams: CoreReportbuilderSystemReportParams = {
+            name: Translate.instant(
+                `addon.badges.${reportParams.context.contextlevel === ContextLevel.COURSE ? 'coursebadges' : 'sitebadges'}`,
+            ),
+            ...reportParams,
+        };
+
+        return [{
+            action: async (siteId: string): Promise<void> => {
+                await CoreNavigator.navigateToSitePath('/reportbuilder/system', { siteId, params: { params: pageParams } });
```
dpalou, line 78: The `params` part is duplicated, so I guess it won't work, right? Shouldn't it be just `params: pageParams`?

dpalou (changes requested):

src/core/features/reportbuilder/components/report-detail/report-detail.ts
crazyserver: Yep, you are not mistaken, but it was taking "source" that is not contextLevel, and contextId which is not contextInstanceId...

src/addons/badges/services/handlers/badge-systemreport-link.ts
crazyserver: No because the param of the page is named params

src/core/features/reportbuilder/pages/list/list.ts
crazyserver: This is intended, the way the list manager works needs page 1 but this is not the case.

src/core/features/reportbuilder/components/report-detail/report-detail.ts
```diff
@@ -102,74 +120,81 @@ export class CoreReportBuilderReportDetailComponent implements OnInit {
...
                 return;
             }
 
-            this.updateState({
-                report,
-                cardVisibleColumns: report.details.settingsdata.cardviewVisibleColumns,
-                cardviewShowFirstTitle: report.details.settingsdata.cardviewShowFirstTitle,
-                canLoadMoreRows: report.data.totalrowcount > report.data.rows.length,
-            });
+            await CoreAlerts.showError(new CoreError(Translate.instant('core.reportbuilder.errorreportview')));
```
dpalou: For system reports IMO it's better to just show the original error, and default to errorreportview if no error message.

src/core/features/reportbuilder/components/report-detail/report-detail.ts
```diff
@@ -102,74 +120,81 @@ export class CoreReportBuilderReportDetailComponent implements OnInit {
...
-            const { page } = this.state$.getValue();
+            this.report.set(report);
 
-            const report = await CoreReportBuilder.loadReport(parseInt(this.reportId), page,this.perPage ?? REPORT_ROWS_LIMIT);
+            if (report.details) {
+                this.logView(report.details);
+                this.onReportLoaded.emit(report.details);
+            }
+        } catch {
+            if (this.reportId()) {
```
dpalou: This was wrong already, but IMO it's better to improve it now. I would only display this custom message if it's a WebService error, and maybe only for invalidresponse? At least check if it's a WS error, that's for sure because we don't want to display this for app offline.

dpalou (changes requested):

dpalou (approved):

PR #4828, opened 2026-07-03
Add host class binding for custom page CSS targeting

dgramm2025:
Add a class per custom page to make easier the CSS targeting


crazyserver:
Hi @dgramm2025,

Thanks for this PR, We've opened the issue in tracker: https://moodle.atlassian.net/browse/MOBILE-5084 where you can provide more information if you want. 

Thanks!

PR #4833, opened 2026-07-08
MOBILE-5047 android: Set AndroidInsecureFileModeEnabled to false

dpalou:


src/core/features/viewer/pages/iframe/iframe.ts
```diff
@@ -29,25 +32,38 @@ import { CoreSharedModule } from '@/core/shared.module';
 })
 export default class CoreViewerIframePage implements OnInit {
 
-    title?: string; // Page title.
-    url?: string; // Iframe URL.
-    autoLogin?: boolean; // Whether to try to use auto-login.
+    @Input() title?: string;
```
crazyserver: It would be great to use signals here. But up to you!

crazyserver (changes requested):

src/core/features/viewer/pages/iframe/iframe.ts
dpalou: Done.

crazyserver (approved):

PR #4835, opened 2026-07-10
Mobile 5049: Check if SCORM's eval can be removed

dpalou:


src/addons/mod/scorm/services/scorm.ts
```diff
@@ -323,59 +323,111 @@ export class AddonModScormProvider {
                         setElement = setElement.trim();
 
                         if (trackData[setElement] !== undefined &&
-                                (trackData[setElement].status == 'completed' || trackData[setElement].status == 'passed')) {
+                                (trackData[setElement].status === 'completed' || trackData[setElement].status === 'passed')) {
```
crazyserver: Do you think we can use an enum to type the statuses values?

crazyserver (changes requested):

src/addons/mod/scorm/services/scorm.ts
dpalou: Adding an enum here is not trivial, because the statuses are used in several ways:

1. To compare a value, like it's done here.
2. To have a valid list of statuses to determine if a status is valid (VALID_STATUSES constant).
3. To define a map of status -> value (STATUSES constant).

I could just create an enum and use it in every constant, but maybe there's a better way to avoid repeating the enum so much. Since it's not trivial, IMO it's better to do it in a separate issue.

crazyserver (approved):

PR #4838, opened 2026-07-10
Mobile 5072 changes on image viewer

crazyserver:


src/core/features/viewer/components/image/image.ts
```diff
@@ -53,12 +52,22 @@ export class CoreViewerImageComponent implements OnInit {
         });
     }
 
-    @Input() title = ''; // Modal title.
-    @Input() image = ''; // Image URL.
-    @Input() component?: string; // Component to use in external-content.
-    @Input() componentId?: string | number; // Component ID to use in external-content.
+    readonly title = input(Translate.instant('core.imageviewer')); // Modal title.
+    readonly image = input(''); // Image URL.
+    readonly imageHTML = input<HTMLImageElement | HTMLPictureElement>(); // Image HTML element.
```
dpalou: The name imageHTML makes me think that this variable has HTML code, but it's the element. I would call it imageElement or imageHTMLElement if you prefer. Both in the component and in the viewImageElement function.

src/core/features/viewer/services/viewer.ts
```diff
@@ -68,6 +68,29 @@ export class CoreViewerService {
...
+    ): Promise<void> {
+        if (!imageHTML) {
+            return;
+        }
+        const { CoreViewerImageComponent } = await import('@features/viewer/components/image/image');
+
+        await CoreModals.openModal({
+            component: CoreViewerImageComponent,
+            componentProps: {
+                imageHTML,
```
dpalou: As we talked in chat, maybe it would be better to pass a clone of the element, that way the component cannot modify the original image.

src/core/features/viewer/services/viewer.ts
```diff
@@ -68,6 +68,29 @@ export class CoreViewerService {
...
+     * View an image HTML element in a modal.
+     *
+     * @param imageHTML Image element of the image.
+     */
+    async viewImageElement(
+        imageHTML: HTMLImageElement | HTMLPictureElement,
+    ): Promise<void> {
+        if (!imageHTML) {
+            return;
+        }
```
dpalou: If we're worried that this can be used to pass elements that aren't images, you could improve this if to check that it's an image. And maybe do it in the component too, just in case.

src/core/features/viewer/components/image/image.html
```diff
@@ -1,13 +1,17 @@
...
         <swiper-slide>
-            <div class="swiper-zoom-container">
-                @if (dataUrl) {
-                    <img [src]="dataUrl" [alt]="title">
-                } @else {
-                    <img [url]="image" [alt]="title" core-external-content [component]="component" [componentId]="componentId">
-                }
-            </div>
+            @if (imageHTML()) {
+                <div class="swiper-zoom-container" [innerHTML]="imageHTML()?.outerHTML"></div>
```
dpalou: In chat you said that maybe the image has CSS that makes the viewer look wrong. Can we remove the inline styles and CSS classes of the image? I'm not sure if there can be some CSS classes or styles that are needed for the image to look ok here in the image viewer.

dpalou (changes requested):

crazyserver:
Thanks for your comments, I've added a new commit to complete this safety checks

dpalou (approved):

PR #4840, opened 2026-07-13
Mobile 5003 Forum mark as read

crazyserver:


src/addons/mod/forum/components/post-options-menu/post-options-menu.ts
```diff
@@ -31,70 +31,61 @@ import { CoreAlerts } from '@services/overlays/alerts';
     imports: [
         CoreSharedModule,
     ],
+    changeDetection: ChangeDetectionStrategy.OnPush,
```
dpalou: This component uses core-loading, and core-loading doesnt use OnPush. So IMO this could cause problems, I think it's better not to put OnPush here.

src/addons/mod/forum/components/post-options-menu/post-options-menu.ts
```diff
@@ -41,26 +40,28 @@ export class AddonModForumPostOptionsMenuComponent implements OnInit {
...
-            return site.createSiteUrl('/mod/forum/discuss.php', { d: post.discussionid.toString() }, `p${post.id}`);
-        }
-    });
 
     readonly isOfflinePost = computed(() => this.postCalculated().id < 0);
+    protected readonly isOnline = CoreNetwork.onlineSignal;
+
+    constructor() {
+        effect(() => {
+            // Display the open in browser button to prevent having an empty menu.
```
dpalou: This comment is wrong.

src/addons/mod/forum/components/post/post.ts
```diff
@@ -142,18 +151,29 @@ export class AddonModForumPostComponent implements OnInit, OnDestroy, OnChanges
...
                 AddonModForumHelper.canUpdatePost(this.post) ||
-                AddonModForumHelper.canSetReadState(this.post);
+                (AddonModForumHelper.canSetReadState(this.post) && isOnline);
         } else {
             // Cannot know if the user can edit/delete/setreadstate or not, display the menu if the WebServices are available.
             this.optionsMenuEnabled = AddonModForum.isGetDiscussionPostAvailable() && (
                 AddonModForumWS.isDeletePostAvailable() ||
                 AddonModForumWS.isUpdatePostAvailable() ||
-                AddonModForum.isSetReadStateAvailable()
+                (AddonModForum.isSetReadStateAvailable() && isOnline)
```
dpalou: Is it expected that the "isOnline" check only affects the isSetReadStateAvailable condition? If you're offline but the delete WS is available, the menu will be shown even if the post is online.

src/addons/mod/forum/tests/behat/read_state.feature
```diff
@@ -0,0 +1,74 @@
...
+    Then I should find "Unread" within "Test reply message 1" "ion-card" in the app
+
+  Scenario: Display options are not shown when forum_usermarksread is disabled
+    Given the following config values are set as admin:
+      | forum_usermarksread | 0 |
+    And I entered the forum activity "Test forum name" on course "Course 1" as "student1" in the app
+    When I press "Test discussion" in the app
+    Then I should not find "Display options" within "Test reply message 1" "ion-card" in the app
+
+  @test
```
dpalou: You forgot to remove this tag.

dpalou (changes requested):

src/addons/mod/forum/components/post-options-menu/post-options-menu.html
```diff
@@ -37,6 +37,32 @@
...
+                        </ion-label>
+                    </ion-item>
+                } @else {
+                    <ion-item lines="none" class="ion-text-wrap">
+                        <ion-icon name="fas-lock" slot="start" aria-hidden="true" />
+                        <ion-label>
+                            <p class="item-heading">{{ 'core.youreoffline' | translate }}</p>
+                            <p>{{ 'addon.mod_forum.reconnecttomanage' | translate }}</p>
+                        </ion-label>
+                    </ion-item>
```
dpalou: It took me a while to understand that this code was behaving as expected, at the beginning I thought the title "you're offline" was being displayed in the wrong case. Probably because of the double negation (if you consider "unknown" as a negation). I would recommend changing the code to:

```
if (unknownCapabilities()) {
} else {
}
```

src/addons/mod/forum/components/post-options-menu/post-options-menu.ts
```diff
@@ -32,64 +32,85 @@ import { AddonModForumHelper } from '../../services/forum-helper';
...
 
     protected readonly postCalculated = linkedSignal(() => this.post());
-    readonly canEdit = computed(() => this.isOfflinePost() ||
-        (AddonModForumHelper.canUpdatePost(this.postCalculated()) && this.isOnline()));
+    readonly canEdit = computed(() => this.isOfflinePost() || AddonModForumHelper.canUpdatePost(this.postCalculated()));
 
-    readonly canDelete = computed(() => this.isOfflinePost() ||
-        (AddonModForumHelper.canDeletePost(this.postCalculated()) && this.isOnline()));
+    readonly canDelete = computed(() => this.isOfflinePost() || AddonModForumHelper.canDeletePost(this.postCalculated()));
+    readonly editDeleteDisabled = computed(() => this.unknownCapabilities() || !this.isOfflinePost() && !this.isOnline());
```
dpalou: When mixing different operators like && and || I think it's a good idea to add parenthesis to make the execution order clear. Because at first glance I don't know which operator will apply first.

src/addons/mod/forum/components/post-options-menu/post-options-menu.ts
```diff
@@ -32,64 +32,85 @@ import { AddonModForumHelper } from '../../services/forum-helper';
...
 
     readonly canMarkAsRead = computed(() =>
-        !this.isOfflinePost() && AddonModForumHelper.canSetReadState(this.postCalculated()) && this.isOnline());
+        !this.isOfflinePost() && AddonModForumHelper.canSetReadState(this.postCalculated()));
+
+    readonly markAsReadDisabled = computed(() => {
+        const wsAvailable = AddonModForum.isSetReadStateAvailable();
```
dpalou: Do we need to check the WS here? If it's not available then the option won't be displayed, right?

src/addons/mod/forum/components/post-options-menu/post-options-menu.ts
```diff
@@ -32,64 +32,85 @@ import { AddonModForumHelper } from '../../services/forum-helper';
...
+        if (this.isOfflinePost()) {
+            return false;
+        }
+
+        // Only check this one because the others are always available if this one is.
+        const wsAvailable = AddonModForum.isDeletePostAvailable();
+        if (!wsAvailable) {
+            // If the WS is not available, we cannot do anything.
+            return false;
+        }
```
dpalou: Again, do we want to check wsAvailable here? If we want to know if capabilities are known or not we just need to check `post.capabilities`, whether the WS is there or not doesn't seem related.

src/addons/mod/forum/components/post-options-menu/post-options-menu.ts
```diff
@@ -32,64 +32,85 @@ import { AddonModForumHelper } from '../../services/forum-helper';
...
+    constructor() {
+        effect(() => {
+            const unknownCapabilities = this.unknownCapabilities();
+            const isOnline = this.isOnline();
 
-            return;
-        }
+            untracked(async () => {
+                if (!unknownCapabilities) {
+                    this.loaded.set(true);
```
dpalou: This loaded change (and the following ones) are set inside an `untracked`, so Angular probably won't notice the change if we switch to a Zoneless behaviour and it can cause bugs.

dpalou (changes requested):

src/addons/mod/forum/components/post-options-menu/post-options-menu.ts
crazyserver: In this case you are right

src/addons/mod/forum/components/post-options-menu/post-options-menu.ts
crazyserver: If WS is not available, capabilities are know, you cannot do the action whether the capabilites object are loaded or not (and we don't have to load them)

dpalou (approved):

PR #4843, opened 2026-07-20
Mobile 5096

crazyserver:


src/core/components/site-logo/site-logo.html
```diff
@@ -1,29 +1,32 @@
-@if (showLogo && logoLoaded) {
+@let siteIdValue = siteId();
+@let siteNameValue = siteName();
```
dpalou: Do you need to store these in variables? Looking at how they're used I guess you can just use siteId() and siteName() where needed, right?

src/core/components/site-logo/site-logo.ts
```diff
@@ -78,53 +126,48 @@ export class CoreSiteLogoComponent implements OnInit, OnDestroy {
...
-     * @returns Site.
+     * @param site The site to update the config from.
+     * @param refresh Whether to refresh the config.
      */
-    protected loadSite(): CoreSite | CoreUnauthenticatedSite {
-        this.site = this.site ?? CoreSites.getRequiredCurrentSite();
+    protected async updateSiteConfig(site: CoreSite | CoreUnauthenticatedSite, refresh = false): Promise<void> {
+        if (site instanceof CoreSite && this.logoType() === CoreSiteLogoType.TOP) {
+            if (refresh) {
+                await site.invalidateConfig();
```
dpalou: The getConfig function has a second parameter to ignore Cache, so you could just do getConfig(undefined, refresh). Read my other comments first though.

src/core/components/site-logo/site-logo.ts
```diff
@@ -39,37 +39,85 @@ import { CoreFormatTextDirective } from '@directives/format-text';
...
-            await this.loadInfo();
-        }, this.siteId);
-
-        this.fallbackLogo = this.logoType === 'top' ? 'assets/img/top_logo.png' : 'assets/img/login_logo.png';
-        this.showSiteName = this.logoType !== 'top';
+            const site = this.siteEffective();
 
-        await this.loadInfo();
+            untracked(() => {
+                void this.updateSiteConfig(site, true);
```
dpalou: Do we really need to pass refresh=true in SITE_UPDATED? This event is usually triggered in updateSiteInfo, and that function already updates the config.

src/core/components/site-logo/site-logo.ts
```diff
@@ -39,37 +39,85 @@ import { CoreFormatTextDirective } from '@directives/format-text';
...
+        const logoType = this.logoType();
+        const showLogo = this.showLogo();
+        const siteConfig = this.siteConfig();
+
+        return showLogo && (logoType !== CoreSiteLogoType.TOP || site.getShowTopLogo(siteConfig) !== 'hidden');
+    });
+
+    protected readonly siteConfig = signal<CoreSiteConfig | undefined>(undefined);
+
+    readonly showLogoEffective = computed(() =>  {
```
dpalou: I think the naming showLogo + shouldShowLogo + showLogoEffective is a bit confusing, it's hard to see the difference. If it was just showLogo + showLogoEffective it would be ok because that's the naming we usually use when applying changes or fallbacks to an input, but having a third one makes it confusing IMO.

Finding a good name for these 3 variables is hard, but I thought of a possible alternative. We can just remove the showLogoEffective one, and have a computed property that only checks the logoError and hideOnError part, something like shouldHideLogo. And in the places where you're currently using showLogoEffective you would need to do "shouldShowLogo() && !shouldHideLogo()". However, it's also confusing to have both a shouldShowLogo and a shouldHideLogo, so maybe the shouldShowLogo should be renamed to something else (maybe this one should be the showLogoEffective?).

dpalou (changes requested):

dpalou (approved):

PR #4848, opened 2026-07-27
Mobile 5089 Support the new Moodle LMS "disabled features" to control how the course navigation works

crazyserver:


src/core/features/course/components/course-format/course-format.html
```diff
@@ -45,24 +45,10 @@
...
-                    }
-                    @if (nextSection) {
-                        <ion-button (click)="sectionChanged(nextSection)" expand="block"
-                            [ariaLabel]="('core.next' | translate) + ': ' + nextSection.name" class="ion-text-nowrap">
-                            <core-format-text [text]="nextSection.name" contextLevel="course" [contextInstanceId]="course.id" />
-                            <ion-icon name="fas-arrow-right" slot="end" aria-hidden="true" />
-                        </ion-button>
-                    }
-                </div>
+                <core-course-section-nav-buttons [currentSectionId]="selectedSection?.id" [allSections]="sections"
```
dpalou, line 50: With the new component, it can happen that the collapsible-footer div doesn't contain anything, just an empty component. Please check that this case doesn't cause any problem in the app (e.g. displaying some small and empty footer, or causing problems with scroll or something).

dpalou (changes requested):

src/core/features/course/components/course-format/course-format.html
crazyserver: Double checking there is no problem with this empty div

dpalou (approved):

PR #4857, opened 2026-07-29
MOBILE-5078 format-text: Fix using pluginfile scripts in user content

dpalou:


src/core/directives/format-text.ts
```diff
@@ -669,7 +669,7 @@ export class CoreFormatTextDirective implements OnDestroy, AsyncDirective {
         const site = await this.getSite();
         const allowedScriptUrls = site?.getContentAllowedScriptUrls() ?? [];
 
-        scripts.forEach((script) => {
+        await CorePromiseUtils.ignoreErrors(Promise.all(scripts.map(async (script) => {
```
crazyserver: IIRC when using Promise.all if one promise fails  the global promise is rejected and the function will return.

In CorePromises we have allPromisesIgnoringErrors that will wait all promises to end before rejecting and I think it would work better here.

crazyserver (changes requested):

crazyserver (approved):

PR #4870, opened 2026-08-25
Mobile 5080

crazyserver:


src/addons/mod/assign/components/submission/addon-mod-assign-submission.html
```diff
@@ -321,9 +321,22 @@ <h2 class="big">{{'addon.mod_assign.feedback' | translate}}</h2>
                                 </ion-item>
                             }
                         }
-
-                        <addon-mod-assign-feedback-plugin *ngFor="let plugin of feedback.plugins" [assign]="assign"
-                            [submission]="userSubmission" [userId]="submitId" [plugin]="plugin" />
+                        @if (!hasMultipleMarkers()) {
+                            <addon-mod-assign-feedback-plugin *ngFor="let plugin of feedback.plugins" [assign]="assign"
+                                [submission]="userSubmission" [userId]="submitId" [plugin]="plugin" />
+                        } @else {
```
dpalou: Are we sure we don't want to show the overall feedback in this case?

src/addons/mod/assign/components/submission/addon-mod-assign-submission.html
```diff
@@ -294,7 +294,7 @@ <h2 class="big">{{'addon.mod_assign.feedback' | translate}}</h2>
                     @if (feedback) {
                         <!-- Only display grader if there's a gradeddate, otherwise it's displayed when anonymous and not revealed. -->
                         @if (feedback.gradeddate) {
-                            @if (!grader) {
+                            @if (!grader || hasMultipleMarkers()) {
```
dpalou: Something I just realized is that in Moodle 5.2 there can be multiple markers but no feedback for markers, so we support that case for students. In Moodle 5.3 is when we don't support it. So we should probably check the Moodle version too.

dpalou (changes requested):

src/addons/mod/assign/components/submission/addon-mod-assign-submission.html
crazyserver: Thanks Dani, after checking it with Denisse I've readded this info!

src/addons/mod/assign/components/submission/submission.ts
```diff
@@ -125,6 +125,20 @@ export class AddonModAssignSubmissionComponent implements OnInit, OnDestroy {
...
+    readonly hasMultipleMarkersWithFeedback = computed(() => {
+        if (!this.hasMultipleMarkers()) {
+            return false;
+        }
+
+        // In Moodle 5.2 there can be multiple markers but no feedback for markers, so we support that case for students.
+        // In Moodle 5.3 is when we don't support it.
+        const site = CoreSites.getRequiredCurrentSite();
+
+        return site.isVersionGreaterEqualThan('5.2');
```
dpalou: Shouldn't this be 5.3?

dpalou (changes requested):

src/addons/mod/assign/components/submission/submission.ts
crazyserver: Oops yes, my bad!

dpalou (approved):

PR #4877, opened 2026-08-28
MOBILE-4826 iframe: Create iframe from TS to fix allow attribute breaks

crazyserver:


src/core/components/iframe/iframe.ts
```diff
@@ -70,79 +70,147 @@ import { BackButtonPriority } from '@/core/constants';
...
 
     constructor() {
         // Listen for messages from the iframe.
         window.addEventListener('message', this.messageListenerFunction = (event) => this.onIframeMessage(event));
 
         effect(() => {
-            this.initIframeElement(this.iframe());
+            const host = this.iframeHostRef();
+            const iframe = this.iframeElement();
+            if (!host || (iframe && host.nativeElement.contains(iframe))) {
```
dpalou: If iframe exists but it's not in host, you're creating a new iframe. Shouldn't we just put the existing iframe inside the host?

src/core/components/iframe/iframe.ts
```diff
@@ -70,79 +70,147 @@ import { BackButtonPriority } from '@/core/constants';
...
+        effect(() => {
+            const iframe = this.iframeElement();
+            if (!iframe) {
+                return;
+            }
+
+            // The "allow" attribute is excluded on purpose: browsers only read it once, when the iframe is created.
+            iframe.style.width = this.formattedWidth();
+            iframe.style.height = this.formattedHeight();
+            iframe.toggleAttribute('allowfullscreen', this.allowFullscreen());
```
dpalou: The allowfullscreen attribute is also part of the "Unsafe bindings" Angular policy, see [NG0910](https://angular.dev/errors/NG0910). This means we should not change it dynamically, as it happens with "allow".

The difference is that when we implemented the allowfullscreen part we decided to allow changing it, and if a dev changed it we **recreated** the iframe. With "allow" we decided to make it static.

Your code will not recreate the iframe if allowFullscreen changes, it will just change the attribute but this might not affect the current page if already loaded.

src/core/components/iframe/iframe.ts
```diff
@@ -70,79 +70,147 @@ import { BackButtonPriority } from '@/core/constants';
...
+        });
+
+        effect(() => {
+            const iframe = this.iframeElement();
+            const url = this.safeUrl();
+            if (!iframe || !url) {
+                return;
+            }
+
+            iframe.src = url ? (DomSanitizer.sanitize(SecurityContext.RESOURCE_URL, url) ?? '') : '';
```
dpalou: Is there any reason why the src is treated different than other attributes? Width, height and id are handled in another effect, but this one only handles src. If the reason is that we don't want to set the src if safeUrl() hasn't changed, maybe we could compare if iframe.src is equal to the value we want to set. I'm saying this to have all the attributes that can be dynamically changed in a single effect, to keep code simpler.

src/core/components/iframe/iframe.ts
```diff
@@ -70,79 +70,147 @@ import { BackButtonPriority } from '@/core/constants';
...
 
-        this.initialized = true;
+        this.initialized.set(true);
 
         // Show loading only with external URLs.
-        this.loading = !this.src || !CoreUrl.isLocalFileUrl(this.src);
+        const src = this.src();
+        this.loading.set(!src || !CoreUrl.isLocalFileUrl(src));
+        this.setLoadingTimeout();
+    }
```
dpalou: Is this code still needed? init() is only called at the end of updateSrc, and with your changes the loading and setLoadingTimeout is already done at the beginning of the updateSrc. So maybe you can just do this.initialized.set(true) instead of this.init()

src/core/components/iframe/iframe.ts
```diff
@@ -244,55 +314,50 @@ export class CoreIframeComponent implements OnChanges, OnDestroy {
...
+        let url = src;
 
-            return;
-        }
-
-        let url = this.src;
+        this.safeUrl.set(undefined);
+        this.displayHelp.set(false);
+        this.loading.set(!src || !CoreUrl.isLocalFileUrl(src));
+        this.setLoadingTimeout();
```
dpalou: One thing that worries me here is that for local URLs you will set loading to false BEFORE setting the URL to the iframe. And this function is async. With the current implementation, local URLs won't await for anything, but this can change in the future. So in the future we might be setting loading to false to display the iframe, but then it takes a while to actually load set the URL. IMO it would be better to set loading to true here, and then when setting the src we can set this condition.

And maybe setLoadingTimeout should be set after setting the URL too, otherwise our async time is already counting for the timeout when this timeout is supposed to be iframe load timeout.

src/core/components/iframe/iframe.ts
```diff
@@ -212,25 +275,32 @@ export class CoreIframeComponent implements OnChanges, OnDestroy {
     /**
      * Initialize things related to the iframe element.
      *
-     * @param iframe Iframe element.
+     * @param host Host element to add the iframe.
      */
-    protected initIframeElement(iframe?: HTMLIFrameElement): void {
```
dpalou, line 217: Since this function is now creating the iframe, I would rename it to `createIframeElement`.

src/core/components/iframe/iframe.ts
```diff
@@ -244,55 +314,50 @@ export class CoreIframeComponent implements OnChanges, OnDestroy {
...
 
         if (url) {
             const { launchExternal, label } = CoreIframe.frameShouldLaunchExternal(url);
 
             if (launchExternal) {
-                this.launchExternalLabel = label;
-                this.loading = false;
+                this.launchExternalLabel.set(label);
+                this.loading.set(false);
+                this.setLoadingTimeout();
```
dpalou: The name setLoadingTimeout makes me think the timeout will be set, not unset. IMO it would be better to have a clearLoadingTimeout function to call it here, inside setLoadingTimeout, and in onDestroy.

dpalou (changes requested):

src/core/components/iframe/iframe.ts
crazyserver: I've changed the condition because if the iframe exists it should be on the host.

src/core/components/iframe/iframe.ts
crazyserver: Hmm and probably initialized is also not needed I'll recheck

src/core/components/iframe/iframe.ts
```diff
@@ -70,79 +70,165 @@ import { BackButtonPriority } from '@/core/constants';
...
+
+            const id = this.id();
+            if (id) {
+                iframe.id = id;
+            } else {
+                iframe.removeAttribute('id');
+            }
+
+            const url = this.safeUrl();
+            const src = url ? (DomSanitizer.sanitize(SecurityContext.RESOURCE_URL, url) ?? '') : '';
```
dpalou: When setting safeUrl, you're using DomSanitizer.bypassSecurityTrustResourceUrl. Why are you sanitizing it again?

src/core/components/iframe/iframe.ts
```diff
@@ -70,79 +70,165 @@ import { BackButtonPriority } from '@/core/constants';
...
+            if (src !== iframe.src) {
+                iframe.src = src;
+                untracked(() => {
+                    this.setLoading(!src || !CoreUrl.isLocalFileUrl(src));
```
dpalou: I think there are a couple of problems here:

1. isLocalFileUrl is called with the sanitized URL, so the check will probably say false.
2. When safeUrl is being calculated, src will be empty. And in that case it doesn't mean the src input is empty, it means we're calculating so we want to display the loading.

IMO this check should be done with `this.src()` as it happened before.

src/core/components/iframe/iframe.ts
```diff
@@ -244,55 +334,48 @@ export class CoreIframeComponent implements OnChanges, OnDestroy {
...
-        }
+    protected async updateSrc(src?: string): Promise<void> {
+        // Discard outdated resolutions if src changes again before this one finishes.
+        const updateId = ++this.srcUpdateId;
+        let url = src;
 
-        let url = this.src;
+        this.safeUrl.set(undefined);
+        this.displayHelp.set(false);
+        this.setLoading(true);
```
dpalou: In this particular case, since we haven't set the iframe URL, IMO we shouldn't set the timeout (we should clear previous timeouts though). E.g. imagine a case where the site is slow and getting the auto login URL takes more than 15 seconds. The timeout will trigger before we've set the iframe URL, the loading will disappear and the user won't see anything.

src/core/components/iframe/iframe.ts
```diff
@@ -305,15 +388,14 @@ export class CoreIframeComponent implements OnChanges, OnDestroy {
...
-        // Now that the URL has been set, initialize the iframe. Wait for the iframe to the added to the DOM.
-        setTimeout(() => {
-            this.init();
-            if (changes.showFullscreenOnToolbar || changes.autoFullscreenOnRotate) {
-                this.configureFullScreen();
-            }
-        });
+        this.safeUrl.set(url ? DomSanitizer.bypassSecurityTrustResourceUrl(CoreFile.convertFileSrc(url)) : undefined);
+
+        // Now that the URL has been set, initialize the iframe. Wait for the iframe to be added to the DOM.
```
dpalou: This comment is no longer true, we're not waiting for it to be in the DOM.

src/core/components/iframe/iframe.ts
```diff
@@ -305,15 +388,14 @@ export class CoreIframeComponent implements OnChanges, OnDestroy {
...
+        this.initialized.set(true);
```
dpalou: Based on your comment, I checked if initialized is needed and it might not be anymore. It's only used in the template. But it's inside a safeUrl() check. If safeUrl() is set it means this code has run, right?

dpalou (changes requested):

src/core/components/iframe/iframe.ts
```diff
@@ -70,95 +68,183 @@ import { BackButtonPriority } from '@/core/constants';
...
+    readonly autoFullscreenOnRotate = input(false, { transform: toBoolean });
+    readonly allowAutoLogin = input(true, { transform: toBoolean });
+    readonly addSiteReferer = input(false, { transform: toBoolean });
+    @Output() loaded = new EventEmitter<HTMLIFrameElement>();
 
-    loading?: boolean;
-    displayHelp = false;
-    fullscreen = false;
-    launchExternalLabel?: string; // Text to set to the button to launch external app.
+    readonly safeUrl = signal<SafeResourceUrl | undefined>(undefined);
```
dpalou: Now that we no longer use DomSanitizer.bypassSecurityTrustResourceUrl because the URL is set to the iframe using JS without involving Angular, we should change the name and type of this variable because it's no longer a SafeResourceUrl. And I guess the toString used later can also be removed:

` const src = url?.toString() ?? '';`

From a security POV, doing a JS set of src with no validation and using bypassSecurityTrustResourceUrl in Angular is the same: in both cases we're allowing any URL. So we haven't decreased the security of the app because of this change.

dpalou (changes requested):

dpalou (approved):

PR #4889, opened 2026-09-15
Mobile 5168

dpalou:


src/core/features/compile/services/compile.ts
```diff
@@ -422,6 +423,7 @@ export class CoreCompileProvider {
             getMainMenuExportedObjects(),
             getContentLinksExportedObjects(),
             getSitePluginsExportedObjects(),
+            [SplashScreen], // @todo Capacitor: Export it in another place instead of hardcoding it here.
```
crazyserver: I would deprecated this for site plugins.

crazyserver: And I think this can be placed in the getNativeServices() using async import in the function

crazyserver (changes requested):

src/core/features/compile/services/compile.ts
dpalou: I deprecated it. I decided to put it in injectLibraries because I think it makes it easier to see it's deprecated for plugins.

crazyserver (approved):

PR #4890, opened 2026-09-15
MOBILE-5154 device: Migrate cordova-plugin-device to @capacitor/device

crazyserver:


src/core/features/pushnotifications/services/pushnotifications.ts
```diff
@@ -334,12 +335,12 @@ export class CorePushNotificationsProvider {
...
-            model:      Device.model,
-            platform:   `${Device.platform}-fcm`,
-            version:    Device.version,
+            name:       NativeDevice.deviceInfo?.manufacturer ?? '',
+            model:      NativeDevice.deviceInfo?.model ?? '',
+            platform:   `${NativeDevice.deviceInfo?.platform}-fcm`,
+            version:    NativeDevice.deviceInfo?.osVersion ?? '',
             pushid:     this.pushID,
-            uuid:       Device.uuid,
+            uuid:       NativeDevice.uuid,
```
dpalou: I think we should use the getInfo() and getId() methods instead of using the public properties whenever it's possible. This is a protected function and it's safe and easy to change its signature to async.

I would only use NativeDevice.deviceInfo in places where we cannot change it to async, like in CorePlatform. Using getter functions is safer for properties that are initialized lazily like these ones, and it makes it easier to keep a stable API.

src/core/services/native/device.ts
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        this.deviceId = await DeviceService.getId();
+        this.uuid = this.deviceId.identifier;
+
+        return this.deviceId; // Add this line to return the device identifier.
+    }
+
+}
+export const NativeDevice = makeSingleton(Device);
```
dpalou: Should we name this CoreDevice or CoreNativeDevice instead? Other wrappers like Network or QRScan are named CoreNetwork and CoreQRScan.

src/core/features/settings/pages/deviceinfo/deviceinfo.html
```diff
@@ -156,42 +148,50 @@ <h1>
...
-                    <p class="item-heading">{{ 'core.settings.cordovadevicemodel' | translate}}</p>
+                    <p class="item-heading">{{ 'core.settings.devicemodel' | translate}}</p>
                     <p>{{ deviceInfo.model }}</p>
                 </ion-label>
             </ion-item>
         }
+        @if (deviceInfo.webViewVersion) {
+            <ion-item (longPress)="copyItemInfo($event)">
+                <ion-label class="ion-text-wrap">
+                    <p class="item-heading">{{ 'core.settings.webviewversion' | translate }}</p>
```
dpalou: Maybe we can name this just "WebView" and put it hardcoded to avoid adding a new string? To avoid generating more work for translators just to display some redundant data.

dpalou (changes requested):

dpalou (approved):

PR #4891, opened 2026-09-15
MOBILE-5151 clipboard: Migrate cordova-clipboard to @capacitor/clipboard

crazyserver:


src/core/services/clipboard.ts
```diff
@@ -0,0 +1,52 @@
...
+    async copy(text: string): Promise<void> {
+        try {
+            await Clipboard.write({ string: text }); // eslint-disable-line id-denylist
+        } catch {
+            // Use HTML Copy command.
+            const virtualInput = document.createElement('textarea');
+            virtualInput.innerHTML = text;
+            virtualInput.select();
+            virtualInput.setSelectionRange(0, 99999);
+            document.execCommand('copy'); // eslint-disable-line @typescript-eslint/no-deprecated
```
dpalou: IMO this is a good opportunity to check if this is still needed. If not I would just delete it.

src/core/services/clipboard.ts
```diff
@@ -0,0 +1,52 @@
...
+        }
+
+        // Show toast using ionicLoading.
+        CoreToasts.show({
+                message: 'core.copiedtoclipboard',
+                translateMessage: true,
+        });
```
dpalou, line 38: Wrong indentation here.

src/core/services/clipboard.ts
```diff
@@ -0,0 +1,52 @@
...
+    }
+
```
dpalou, line 40: Since the `paste` function is easy to implement, maybe we could add it here too? Otherwise if a plugin is using it there's no alternative. The `clear` one seems a lot more edge.

dpalou (changes requested):

dpalou (approved):

PR #4901, opened 2026-09-21
MOBILE-5172 login: Fix display QR scan on reconnect screen

crazyserver:


src/core/features/login/components/login-methods/login-methods.ts
```diff
@@ -94,21 +94,25 @@ export class CoreLoginMethodsComponent implements OnInit {
...
+                this.showScanQR = false;
 
-            if (this.showScanQR) {
                 return;
             }
         }
 
-        // If still false or credentials screen.
         if (this.siteConfig) {
             this.showScanQR = await CoreLoginHelper.displayQRInCredentialsScreen(this.siteConfig.tool_mobile_qrcodetype);
```
dpalou: This won't work as expected for reconnect screen. The reason is that `displayQRInCredentialsScreen` only checks `tool_mobile_qrcodetype` if `displayqroncredentialscreen` is true, or if it's undefined but it's a custom app with a single site. Our official app doesn't fulfill any of those conditions, so the function will always return false, no matter the value of the config.

I think that in reconnect we should check `tool_mobile_qrcodetype` only if QR scan is configured to be displayed either in site screen or credentials screen (that is, checking `displayqronsitescreen` and `displayqroncredentialscreen`). If the QR is not configured to be displayed before login, it doesn't make sense to display it in reconnect.

dpalou (dismissed):

src/core/features/login/services/login-helper.ts
```diff
@@ -1080,33 +1080,71 @@ export class CoreLoginHelperProvider {
...
-     * @param qrCodeType QR Code type from public config, assuming enabled if undefined.
+     * @param qrCodeType QR Code type from public config.
      * @returns Whether the QR reader should be displayed in credentials screen.
      */
-    async displayQRInCredentialsScreen(qrCodeType = CoreSiteQRCodeType.QR_CODE_LOGIN): Promise<boolean> {
-        if (!CoreQRScan.canScanQR()) {
+    async displayQRInCredentialsScreen(qrCodeType?: CoreSiteQRCodeType): Promise<boolean> {
+        if (!CoreQRScan.canScanQR() ||
+            qrCodeType !== CoreSiteQRCodeType.QR_CODE_LOGIN ||
+            CoreConstants.CONFIG.displayqroncredentialscreen === false) {
```
dpalou:
```suggestion
        if (
            !CoreQRScan.canScanQR() ||
            qrCodeType !== CoreSiteQRCodeType.QR_CODE_LOGIN ||
            CoreConstants.CONFIG.displayqroncredentialscreen === false
        ) {
```

src/core/features/login/services/login-helper.ts
```diff
@@ -1080,33 +1080,71 @@ export class CoreLoginHelperProvider {
...
             return false;
         }
 
-        const isSingleFixedSite = await this.isSingleFixedSite();
+        const qrShownInSiteScreen = CoreLoginHelper.displayQRInSiteScreen();
+        // Don't show the button if it has been shown in the site screen.
+        if (qrShownInSiteScreen) {
+            return false;
+        }
```
dpalou: Unfortunately, this check can break custom apps (and our own branded apps). `displayQRInSiteScreen` only checks the value of `displayqronsitescreen`, which is usually undefined (and the default is true). But in custom apps with a single site the value of `displayqronsitescreen` can be undefined or true but the user will still not see the QR login in site screen because the user will never see the site screen.

src/core/features/login/services/login-helper.ts
```diff
@@ -1080,33 +1080,71 @@ export class CoreLoginHelperProvider {
...
+    /**
+     * Check whether the QR reader should be displayed in reconnect screen.
+     *
+     * @param qrCodeType QR Code type from public config.
+     * @returns Whether the QR reader should be displayed in reconnect screen.
+     */
+    async displayQRInReconnectScreen(qrCodeType?: CoreSiteQRCodeType): Promise<boolean> {
+        if (!CoreQRScan.canScanQR() ||
+            qrCodeType !== CoreSiteQRCodeType.QR_CODE_LOGIN ||
+            CoreConstants.CONFIG.displayqroncredentialscreen === false) {
```
dpalou:
```suggestion
        if (
            !CoreQRScan.canScanQR() ||
            qrCodeType !== CoreSiteQRCodeType.QR_CODE_LOGIN ||
            CoreConstants.CONFIG.displayqroncredentialscreen === false
        ) {
```

src/core/features/login/tests/behat/qrlogin.feature
```diff
@@ -0,0 +1,171 @@
...
+    # Reconnect screen NOT SHOWN
+    When I set the following fields to these values in the app:
+      | Username | student1 |
+      | Password | student1 |
+    And I press "Log in" near "Lost password?" in the app
+    And I press the user menu button in the app
+    And I press "Log out" in the app
+    And I wait the app to restart
+    And I press "david student" in the app
+    And I should not be able to press "Scan QR code" in the app
```
dpalou:
```suggestion
    Then I should not be able to press "Scan QR code" in the app
```

src/core/features/login/tests/behat/qrlogin.feature
```diff
@@ -0,0 +1,171 @@
...
```
dpalou:
```suggestion
    Then I should not be able to press "Scan QR code" in the app
```

src/core/features/login/tests/behat/qrlogin.feature
```diff
@@ -0,0 +1,171 @@
...
```
dpalou:
```suggestion
    Then I should not be able to press "Scan QR code" in the app
```

src/core/features/login/tests/behat/qrlogin.feature
```diff
@@ -0,0 +1,171 @@
...
+    # Reconnect screen SHOWN
+    When I set the following fields to these values in the app:
+      | Username | student1 |
+      | Password | student1 |
+    And I press "Log in" near "Lost password?" in the app
+    And I press the user menu button in the app
+    And I press "Log out" in the app
+    And I wait the app to restart
+    And I press "david student" in the app
+    And I should be able to press "Scan QR code" in the app
```
dpalou:
```suggestion
    Then I should be able to press "Scan QR code" in the app
```

src/core/features/login/tests/behat/qrlogin.feature
```diff
@@ -0,0 +1,171 @@
...
```
dpalou:
```suggestion
    Then I should be able to press "Scan QR code" in the app
```

src/core/features/login/tests/behat/qrlogin.feature
```diff
@@ -0,0 +1,171 @@
...
+    # Reconnect screen NOT SHOWN
+    When I set the following fields to these values in the app:
+      | Username | student1 |
+      | Password | student1 |
+    And I press "Log in" near "Lost password?" in the app
+    And I press the user menu button in the app
+    And I press "Log out" in the app
+    And I wait the app to restart
+    And I press "david student" in the app
+    And I should not be able to press "Scan QR code" in the app
```
dpalou:
```suggestion
    Then I should not be able to press "Scan QR code" in the app
```

dpalou (changes requested):

dpalou (approved):

